PR #1, opened 2016-04-15
Refactoring

ash2k:
FYI @aelse.

### DIFF
--- a/backend/backends/null/null.go
+++ b/backend/backends/null/null.go
@@ -15,7 +15,7 @@ func init() {
 	})
 }
 
-// Client is an object that is used to send messages to stdout.
+// Client is an object that is used to discard messages.
 type Client struct{}
 
 // NewClient constructs a StdoutClient object.
@@ -23,7 +23,7 @@ func NewClient() (backend.MetricSender, error) {
 	return &Client{}, nil
 }
 
-// SampleConfig returns the sample config for the stdout backend.
+// SampleConfig returns the sample config for the null backend.
 func (client *Client) SampleConfig() string {
 	return ""
 }
--- a/backend/backends/null/null.go
+++ b/backend/backends/null/null.go
@@ -1,0 +1,39 @@
+package null
+
+import (
+	"github.com/atlassian/gostatsd/backend"
+	"github.com/atlassian/gostatsd/types"
+
+	"github.com/spf13/viper"
+)
+
+const backendName = "null"
+
+func init() {
+	backend.RegisterBackend(backendName, func(v *viper.Viper) (backend.MetricSender, error) {
+		return NewClient()
+	})
+}
+
+// Client is an object that is used to send messages to stdout.
+type Client struct{}
+
+// NewClient constructs a StdoutClient object.
+func NewClient() (backend.MetricSender, error) {
+	return &Client{}, nil
+}
+
+// SampleConfig returns the sample config for the stdout backend.
+func (client *Client) SampleConfig() string {
+	return ""
+}
+
+// SendMetrics discards the metrics in a MetricsMap.
+func (client *Client) SendMetrics(metrics types.MetricMap) error {
+	return nil
+}
+
+// BackendName returns the name of the backend.
+func (client *Client) BackendName() string {
+	return backendName
+}
--- a/backend/backends/stdout/stdout.go
+++ b/backend/backends/stdout/stdout.go
@@ -46,7 +46,7 @@ func (client *Client) SampleConfig() string {
 }
 
 // SendMetrics sends the metrics in a MetricsMap to the Graphite server.
-func (client *Client) SendMetrics(metrics types.MetricMap) error {
+func (client *Client) SendMetrics(metrics types.MetricMap) (retErr error) {
 	buf := new(bytes.Buffer)
 	now := time.Now().Unix()
 	metrics.Counters.Each(func(key, tagsKey string, counter types.Counter) {
@@ -79,14 +79,14 @@ func (client *Client) SendMetrics(metrics types.MetricMap) error {
 		fmt.Fprintf(buf, "stats.set.%s %d %d\n", nk, len(set.Values), now)
 	})
 
-	fmt.Fprintf(buf, "statsd.numStats %d %d\n", metrics.NumStats, now)
-	fmt.Fprintf(buf, "statsd.processingTime %f %d\n", float64(metrics.ProcessingTime)/float64(time.Millisecond), now)
-
-	_, err := buf.WriteTo(log.StandardLogger().Writer())
-	if err != nil {
-		return err
-	}
-	return nil
+	writer := log.StandardLogger().Writer()
+	defer func() {
+		if err := writer.Close(); err != nil && retErr == nil {
+			retErr = err
+		}
+	}()
+	_, err := buf.WriteTo(writer)
+	return err
 }
 
 // BackendName returns the name of the backend.
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,7 +1,7 @@
 FROM alpine:3.3
 
-RUN apk add --no-cache ca-certificates && \
-	rm -rf /var/cache/apk/*
+RUN apk --no-cache add \
+    ca-certificates
 
 ADD bin/linux/gostatsd /bin/gostatsd
 
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,5 +1,8 @@
 FROM alpine:3.3
-RUN apk --update --no-cache add \
-	ca-certificates
+
+RUN apk add --no-cache ca-certificates && \
+	rm -rf /var/cache/apk/*
+
 ADD bin/linux/gostatsd /bin/gostatsd
+
 ENTRYPOINT ["gostatsd"]
--- a/gostatsd.go
+++ b/gostatsd.go
@@ -67,8 +67,8 @@ func main() {
 			return
 		}
 		defer func() {
-			if err := f.Close(); err != nil {
-				augmentErr(&exitErr, fmt.Errorf("Failed to close profile file: %v", err))
+			if closeErr := f.Close(); closeErr != nil {
+				augmentErr(&exitErr, fmt.Errorf("Failed to close profile file: %v", closeErr))
 			}
 		}()
 		err = pprof.StartCPUProfile(f)
--- a/gostatsd.go
+++ b/gostatsd.go
@@ -93,7 +93,6 @@ func main() {
 		FlushInterval:    v.GetDuration(statsd.ParamFlushInterval),
 		MaxReaders:       v.GetInt(statsd.ParamMaxReaders),
 		MaxWorkers:       v.GetInt(statsd.ParamMaxWorkers),
-		MaxMessengers:    v.GetInt(statsd.ParamMaxMessengers),
 		MetricsAddr:      v.GetString(statsd.ParamMetricsAddr),
 		Namespace:        v.GetString(statsd.ParamNamespace),
 		PercentThreshold: toSlice(v.GetString(statsd.ParamPercentThreshold)),
@@ -117,6 +116,9 @@ func augmentErr(exitErr *error, newErr error) {
 
 func toSlice(s string) []string {
 	//TODO Remove workaround when https://github.com/spf13/viper/issues/112 is fixed
+	if s == "" {
+		return nil
+	}
 	return strings.Split(s, ",")
 }
 
--- a/gostatsd.go
+++ b/gostatsd.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
-	"runtime/pprof"
 	"strings"
 	"syscall"
 
@@ -28,8 +27,6 @@ var (
 const (
 	// ParamVerbose enables verbose logging.
 	ParamVerbose = "verbose"
-	// ParamCPUProfile enables use of profiler and write results to this file.
-	ParamCPUProfile = "cpu-profile"
 	// ParamJSON makes logger log in JSON format.
 	ParamJSON = "json"
 	// ParamConfigPath provides file with configuration.
@@ -59,25 +56,6 @@ func main() {
 			log.Fatalf("%v", exitErr)
 		}
 	}()
-	CPUProfile := v.GetString(ParamCPUProfile)
-	if CPUProfile != "" {
-		f, err := os.Create(CPUProfile)
-		if err != nil {
-			augmentErr(&exitErr, fmt.Errorf("Failed to open profile file: %v", err))
-			return
-		}
-		defer func() {
-			if closeErr := f.Close(); closeErr != nil {
-				augmentErr(&exitErr, fmt.Errorf("Failed to close profile file: %v", closeErr))
-			}
-		}()
-		err = pprof.StartCPUProfile(f)
-		if err != nil {
-			augmentErr(&exitErr, fmt.Errorf("Failed to start CPU profiler: %v", err))
-			return
-		}
-		defer pprof.StopCPUProfile()
-	}
 
 	ctx, cancelFunc := context.WithCancel(context.Background())
 	defer cancelFunc()
@@ -149,7 +127,6 @@ func setupConfiguration() (*viper.Viper, bool, error) {
 	cmd.BoolVar(&version, ParamVersion, false, "Print the version and exit")
 	cmd.Bool(ParamVerbose, false, "Verbose")
 	cmd.Bool(ParamJSON, false, "Log in JSON format")
-	cmd.String(ParamCPUProfile, "", "Use profiler and write results to this file")
 	cmd.String(ParamConfigPath, "", "Path to the configuration file")
 
 	statsd.AddFlags(cmd)
--- a/statsd/aggregator.go
+++ b/statsd/aggregator.go
@@ -29,12 +29,12 @@ type aggregator struct {
 	expiryInterval    time.Duration // How often to expire metrics
 	lastFlush         time.Time     // Last time the metrics where aggregated
 	percentThresholds []float64
-	systemTags        string // Tags to add to system metrics
+	defaultTags       string // Tags to add to system metrics
 	types.MetricMap
 }
 
 // NewAggregator creates a new Aggregator object.
-func NewAggregator(percentThresholds []float64, flushInterval, expiryInterval time.Duration, systemTags []string) Aggregator {
+func NewAggregator(percentThresholds []float64, flushInterval, expiryInterval time.Duration, defaultTags []string) Aggregator {
 	a := aggregator{}
 	a.FlushInterval = flushInterval
 	a.lastFlush = time.Now()
@@ -44,7 +44,7 @@ func NewAggregator(percentThresholds []float64, flushInterval, expiryInterval ti
 	a.Timers = types.Timers{}
 	a.Gauges = types.Gauges{}
 	a.Sets = types.Sets{}
-	a.systemTags = types.Tags(systemTags).String()
+	a.defaultTags = types.Tags(defaultTags).String()
 	return &a
 }
 
@@ -60,7 +60,7 @@ func (a *aggregator) Flush(now func() time.Time) *types.MetricMap {
 	flushInterval := startTime.Sub(a.lastFlush)
 
 	statName := internalStatName("num_stats")
-	a.receiveCounter(statName, a.systemTags, int64(a.NumStats), startTime)
+	a.receiveCounter(statName, a.defaultTags, int64(a.NumStats), startTime)
 
 	a.Counters.Each(func(key, tagsKey string, counter types.Counter) {
 		perSecond := float64(counter.Value) / flushInterval.Seconds()
@@ -154,7 +154,7 @@ func (a *aggregator) Flush(now func() time.Time) *types.MetricMap {
 	a.ProcessingTime = flushTime.Sub(startTime)
 
 	statName = internalStatName("processing_time")
-	a.receiveGauge(statName, a.systemTags, float64(a.ProcessingTime)/float64(time.Millisecond), flushTime)
+	a.receiveGauge(statName, a.defaultTags, float64(a.ProcessingTime)/float64(time.Millisecond), flushTime)
 
 	a.lastFlush = flushTime
 
--- a/statsd/aggregator.go
+++ b/statsd/aggregator.go
@@ -15,17 +15,17 @@ import (
 type ProcessFunc func(*types.MetricMap)
 
 // Aggregator is an object that aggregates statsd metrics.
-// The function NewMetricAggregator should be used to create the objects.
+// The function NewAggregator should be used to create the objects.
 //
-// Incoming metrics should be passed via ReceiveMetric function.
+// Incoming metrics should be passed via Receive function.
 type Aggregator interface {
-	ReceiveMetric(*types.Metric, time.Time)
+	Receive(*types.Metric, time.Time)
 	Flush(func() time.Time) *types.MetricMap
 	Process(ProcessFunc)
 	Reset(time.Time)
 }
 
-type metricAggregator struct {
+type aggregator struct {
 	expiryInterval    time.Duration // How often to expire metrics
 	lastFlush         time.Time     // Last time the metrics where aggregated
 	percentThresholds []float64
@@ -33,9 +33,9 @@ type metricAggregator struct {
 	types.MetricMap
 }
 
-// NewMetricAggregator creates a new Aggregator object.
-func NewMetricAggregator(percentThresholds []float64, flushInterval, expiryInterval time.Duration, systemTags []string) Aggregator {
-	a := metricAggregator{}
+// NewAggregator creates a new Aggregator object.
+func NewAggregator(percentThresholds []float64, flushInterval, expiryInterval time.Duration, systemTags []string) Aggregator {
+	a := aggregator{}
 	a.FlushInterval = flushInterval
 	a.lastFlush = time.Now()
 	a.expiryInterval = expiryInterval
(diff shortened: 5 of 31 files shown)


statsd/dispatcher.go
```diff
@@ -0,0 +1,187 @@
...
+	Process(context.Context, ProcessFunc) *sync.WaitGroup
+}
+
+// AggregatorFactory creates Aggregator objects.
+type AggregatorFactory interface {
+	// Create creates Aggregator objects.
+	Create() Aggregator
+}
+
+// The AggregatorFactoryFunc type is an adapter to allow the use of ordinary functions as AggregatorFactory.
```
ojongerius: Rather `// AggregatorFactoryFunc type is ..` ie: cut "The"

statsd/dispatcher.go
```diff
@@ -0,0 +1,187 @@
...
+	go func() {
+		// Wait until all the workers finish to ensure we close the channel only after we know nobody is going to send to it
+		cmd.wg.Wait()
+		close(results) // Signal consumer there are no more flush results
+	}()
+	return results
+}
+
+// Process concurrently executes provided function in goroutines that own Aggregators.
+// Function may be executed zero or up to numWorkers times. It is executed
```
ojongerius: It's not function but a method, right?

ash2k: Hm, it is ambiguous. It should be talking about `ProcessFunc` function, not about `Process` method. Will adjust the wording. Cheers!

ash2k:
I tried to do performance testing. Here is what I found out:
* Current version crashed with out of memory error after a short period of time (I used default packet size, not 10 as it was in the command line history). Without load system memory sits at 360+MB out of 1GB. With load it quickly eats all of that memory. My guess is that it is because of queue sizes. It is just not enough memory to buffer that many metrics.
* CPU is at 100% with both versions. The box has only one core so we are not really testing scalability of the app architecture.
* Loader tool shows roughly the same numbers of metrics/packets per second.

I propose the following improvements to the testing process:
* We should be measuring number of metrics and packets processed at the receiver because loader tool just sends them at maximum speed and receiver drops them at some stack layer - number of received packets is 10 times less than number of sent packets.
* Of course we should use much more beefier receiver box with many cores to actually see how well it can scale up.

ash2k:
Modified the tester to randomize metric names to better simulate load and added `null` backend to avoid wasting time on I/O. Also applied null backend to master branch.

On my laptop (8 core Macbook) results are the following (metrics/second):
this branch:
```
INFO[0053] total_num_stats: 275420
INFO[0054] total_num_stats: 459809
INFO[0055] total_num_stats: 491043
INFO[0056] total_num_stats: 486114
INFO[0057] total_num_stats: 503397
INFO[0058] total_num_stats: 511011
INFO[0060] total_num_stats: 504829
INFO[0061] total_num_stats: 492364
INFO[0062] total_num_stats: 468319
INFO[0063] total_num_stats: 497444
INFO[0064] total_num_stats: 443146
INFO[0065] total_num_stats: 424983
INFO[0066] total_num_stats: 380755
INFO[0067] total_num_stats: 501699
INFO[0069] total_num_stats: 417499
INFO[0070] total_num_stats: 265650
```

master:
```
INFO[0010] num stats: 16566
INFO[0011] num stats: 194639
INFO[0012] num stats: 287221
INFO[0013] num stats: 307048
INFO[0014] num stats: 334974
INFO[0015] num stats: 326196
INFO[0017] num stats: 309394
INFO[0018] num stats: 441583
INFO[0019] num stats: 435501
INFO[0021] num stats: 321533
INFO[0022] num stats: 337904
INFO[0024] num stats: 436203
INFO[0025] num stats: 266969
INFO[0026] num stats: 342134
INFO[0028] num stats: 397097
INFO[0029] num stats: 297965
INFO[0031] num stats: 348746
INFO[0032] num stats: 394785
INFO[0033] num stats: 268378
INFO[0035] num stats: 376261
STOP
INFO[0036] num stats: 570678
INFO[0037] num stats: 618046
INFO[0039] num stats: 528599
INFO[0040] num stats: 523940
INFO[0041] num stats: 603497
INFO[0042] num stats: 710854
INFO[0044] num stats: 132046
INFO[0045] num stats: 0
INFO[0046] num stats: 0
```

Note that master branch shows great performance at the very end after the `STOP`. My guess is that after the stop command is given to the load tester, main process has more cpu at its disposal and also there is no need to read anything from socket so it just spends all time processing messages from the queue(s). Hence performance spike.

Will try this on a server box. Also I think I'll write an in-process benchmark to avoid network stack completely.

build/Dockerfile
```diff
@@ -1,5 +1,8 @@
 FROM alpine:3.3
-RUN apk --update --no-cache add \
-	ca-certificates
+
+RUN apk add --no-cache ca-certificates && \
+	rm -rf /var/cache/apk/*
```
jtblin: I thought `--no-cache` which is new in alpine 3.3 does the same thing than that, no?

ash2k: I don't know. Two files are created in this directory after the `apk add` command and that is what they do in an example here: https://hub.docker.com/_/alpine/

jtblin: The example is for 3.1... So we need to remove the `--update` for the `--no-cache` to work: https://github.com/gliderlabs/docker-alpine/blob/master/docs/usage.md#disabling-cache

```
RUN apk --no-cache add \
	ca-certificates
```

I confirmed locally:

```
/ # ls -l /var/cache/apk/
total 0
```

jtblin: Good catch ;)

ash2k:
Done some benchmarking. This is the `perf` branch with is just `master` branch with the benchmark:
```
time="2016-04-26T13:46:39-05:00" level=info msg="total_num_stats: 367909" 
time="2016-04-26T13:46:40-05:00" level=info msg="total_num_stats: 511962" 
time="2016-04-26T13:46:41-05:00" level=info msg="total_num_stats: 493162" 
time="2016-04-26T13:46:42-05:00" level=info msg="total_num_stats: 464163" 
time="2016-04-26T13:46:43-05:00" level=info msg="total_num_stats: 517408" 
time="2016-04-26T13:46:44-05:00" level=info msg="total_num_stats: 494934" 
time="2016-04-26T13:46:46-05:00" level=info msg="total_num_stats: 495710" 
time="2016-04-26T13:46:47-05:00" level=info msg="total_num_stats: 496206" 
time="2016-04-26T13:46:48-05:00" level=info msg="total_num_stats: 508446" 
time="2016-04-26T13:46:49-05:00" level=info msg="total_num_stats: 515078" 
time="2016-04-26T13:46:50-05:00" level=info msg="total_num_stats: 494761" 
time="2016-04-26T13:46:51-05:00" level=info msg="total_num_stats: 485544" 
time="2016-04-26T13:46:53-05:00" level=info msg="total_num_stats: 508453" 
```
This is `refactoring` branch:
```
time="2016-04-26T13:48:52-05:00" level=info msg="total_num_stats: 1128478" 
time="2016-04-26T13:48:53-05:00" level=info msg="total_num_stats: 1178775" 
time="2016-04-26T13:48:54-05:00" level=info msg="total_num_stats: 1219097" 
time="2016-04-26T13:48:55-05:00" level=info msg="total_num_stats: 1205117" 
time="2016-04-26T13:48:56-05:00" level=info msg="total_num_stats: 1188569" 
time="2016-04-26T13:48:57-05:00" level=info msg="total_num_stats: 1163795" 
time="2016-04-26T13:48:58-05:00" level=info msg="total_num_stats: 1140139" 
time="2016-04-26T13:48:59-05:00" level=info msg="total_num_stats: 1164400" 
time="2016-04-26T13:49:00-05:00" level=info msg="total_num_stats: 1081606" 
time="2016-04-26T13:49:02-05:00" level=info msg="total_num_stats: 1111094" 
time="2016-04-26T13:49:03-05:00" level=info msg="total_num_stats: 1117562" 
time="2016-04-26T13:49:04-05:00" level=info msg="total_num_stats: 1110116" 
time="2016-04-26T13:49:05-05:00" level=info msg="total_num_stats: 1088970" 
```
* `null` output is used to discard flushed metrics instead of sending them to stdout/etc to intensify the load on the part we actually want to test.
* fake network socket is used to provide randomly generated metrics for the same reason ^
* both benchmarks output number of metrics processed per second.
* results are from my laptop.
/cc @jtblin

jtblin:
Wow pretty impressive improvements. Would love to see a test in the platform-dev environment. Nit: I usually look at packets per second.

backend/backends/null/null.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"github.com/atlassian/gostatsd/backend"
+	"github.com/atlassian/gostatsd/types"
+
+	"github.com/spf13/viper"
+)
+
+const backendName = "null"
+
+func init() {
```
jtblin: Don't need to change but FYI, idiomatic go code starts from bottom and moves up. Comes from C/C++ I think.

backend/backends/stdout/stdout.go
```diff
@@ -79,14 +79,14 @@ func (client *Client) SendMetrics(metrics types.MetricMap) error {
...
-	fmt.Fprintf(buf, "statsd.processingTime %f %d\n", float64(metrics.ProcessingTime)/float64(time.Millisecond), now)
-
-	_, err := buf.WriteTo(log.StandardLogger().Writer())
-	if err != nil {
-		return err
-	}
-	return nil
+	writer := log.StandardLogger().Writer()
+	defer func() {
+		if err := writer.Close(); err != nil && retErr == nil {
```
jtblin: Isn't that a bit too much i.e. can just `defer writer.Close()`?

build/Dockerfile
jtblin: - [ ] change to `RUN apk --no-cache add \ ca-certificates`

gostatsd.go
```diff
@@ -67,8 +67,8 @@ func main() {
 			return
 		}
 		defer func() {
-			if err := f.Close(); err != nil {
-				augmentErr(&exitErr, fmt.Errorf("Failed to close profile file: %v", err))
+			if closeErr := f.Close(); closeErr != nil {
```
jtblin: I really think this is too much. I don't think we care about the error here. If it is because of the linter, then we can disable warnings for the `Close` case. This is too verbose imho.

statsd/aggregator.go
```diff
@@ -3,61 +3,48 @@ package statsd
...
-	LastMessage    time.Time
-	LastFlush      time.Time
-	LastFlushError time.Time
-	NumStats       int
-	ProcessingTime time.Duration
-}
+// ProcessFunc is a function that gets executed by Aggregator with its state passed into the function.
+type ProcessFunc func(*types.MetricMap)
 
-// MetricAggregator is an object that aggregates statsd metrics.
```
jtblin: Why changing the name?

statsd/aggregator.go
```diff
@@ -3,61 +3,48 @@ package statsd
...
-	Tags              types.Tags // Tags to add to all metrics
+// Incoming metrics should be passed via ReceiveMetric function.
+type Aggregator interface {
+	ReceiveMetric(*types.Metric, time.Time)
+	Flush(func() time.Time) *types.MetricMap
+	Process(ProcessFunc)
+	Reset(time.Time)
+}
+
+type metricAggregator struct {
```
jtblin: And this one is a `metricAggregator` which implements `Aggregator` and has a `ReceiveMetric` function and is created via a `NewMetricAggregator` function... A bit confusing, I'd rather we are consistent.

jtblin: So now you made `metricAggregator` private which is a "breaking change". Is that so that we are forced to use `NewMetricAggregator`?

statsd/aggregator.go
```diff
@@ -3,61 +3,48 @@ package statsd
...
+	percentThresholds []float64
+	systemTags        string // Tags to add to system metrics
 	types.MetricMap
 }
 
-// NewMetricAggregator creates a new MetricAggregator object.
-func NewMetricAggregator(senders []backend.MetricSender, percentThresholds []float64, flushInterval time.Duration, expiryInterval time.Duration, maxWorkers int, tags []string) *MetricAggregator {
-	a := MetricAggregator{}
+// NewMetricAggregator creates a new Aggregator object.
+func NewMetricAggregator(percentThresholds []float64, flushInterval, expiryInterval time.Duration, systemTags []string) Aggregator {
```
jtblin: Naming is hard, they are not "system" tags imho, there are default tags but idiomatic Go says you should keep the names short so I don't think it brings much to change the name (especially if we disagree on it :P).

statsd/dispatcher.go
```diff
@@ -0,0 +1,185 @@
...
+	aggr        Aggregator
+}
+
+type dispatcher struct {
+	workers map[uint16]*worker
+}
+
+// NewDispatcher creates a new Dispatcher with provided configuration.
+func NewDispatcher(numWorkers int, perWorkerBufferSize int, af AggregatorFactory) Dispatcher {
+	workers := make(map[uint16]*worker)
```
jtblin: `workers := make(map[uint16]*worker, numWorkers)`?

statsd/dispatcher.go
```diff
@@ -0,0 +1,185 @@
...
+
+	n := uint16(numWorkers)
+
+	for i := uint16(0); i < n; i++ {
+		workers[i] = &worker{
+			make(chan *types.Metric, perWorkerBufferSize),
```
jtblin: Should use the variable for assignment, I find it much clearer and depending on the order is a bad pattern imho.

statsd/dispatcher.go
```diff
@@ -0,0 +1,185 @@
...
+	result chan<- *types.MetricMap
+}
+
+type processCommand struct {
+	wg sync.WaitGroup
+	f  ProcessFunc
+}
+
+type worker struct {
+	metrics     chan *types.Metric
```
jtblin: Use alphabetic order when possible.

statsd/dispatcher.go
```diff
@@ -0,0 +1,185 @@
...
+	flushChan   chan *flushCommand
+	processChan chan *processCommand
+	aggr        Aggregator
+}
+
```
jtblin: Could we do something like `type dispatcher map[uint16]*worker` instead?

statsd/dispatcher.go
```diff
@@ -0,0 +1,185 @@
...
+	results := make(chan *types.MetricMap, len(d.workers)) // Enough capacity not to block workers
+	cmd := &flushCommand{ctx, sync.WaitGroup{}, results}
+	cmd.wg.Add(len(d.workers))
+	flushesSent := 0
+loop:
+	for _, worker := range d.workers {
+		select {
+		case <-ctx.Done():
+			cmd.wg.Add(flushesSent - len(d.workers)) // Not all flushes have been sent, should decrement the WG counter.
+			break loop
```
jtblin: Old school, cannot do anything than that?

statsd/dispatcher.go
jtblin: I'd rename `metrics` to something like `metricsQueue` for clarity.

statsd/receiver.go
```diff
@@ -5,133 +5,116 @@ import (
...
+			log.Warnf("Error reading from socket: %v", err)
 			continue
 		}
-		msg := buf[:nbytes]
-		mr.increment("packets_received", 1)
-		mq.enqueue(message{addr, msg}, mr)
-		buf = buf[nbytes:]
-		runtime.Gosched()
+		// TODO consider updating counter for every N-th iteration to reduce contention
+		atomic.AddUint64(&mr.packetsReceived, 1)
```
jtblin: Why moving to an atomic counter and when is that sent now?

backend/backends/null/null.go
ash2k: Could you elaborate more please? I don't understand :) I just copied some existing backend and removed everything except method signatures.

backend/backends/stdout/stdout.go
ash2k: In the past I've been burnt many times by code that was ignoring errors (even logs about "not important" errors could be very useful to troubleshoot something). It is better to find out about any errors happening than see clean logs and think that all is good. I.e. we can always silence down unnecessary noise later but we will never know that something is wrong if we ignore errors from start.

If you think about it, there must be a reason the OS/C-level functions for I/O (including closing the file) return error. If all such errors were not important or non-existent close() would not return error. The actual reason is when you close the file/stream/socket OS flushes the data to disk/network if some was buffered, syncs meta information to disk about last file update time/etc. So if a error occurs while closing a file it might mean e.g. that HDD just died. Don't we want to find out about such errors?

Also you might be interested in reading comments here https://github.com/google/guava/issues/1118

statsd/aggregator.go
ash2k: I don't have strong preference for names, I agree currently this is not ideal. Everything here works with metrics so including "metrics" into each interface name and method name is too much IMHO. I think we should either add "Metric" to interface names and remove it from "ReceiveMetric" OR remove "metric" from everywhere. WDYT?

I introduced a breaking change because I'd argue we want to hide implementation to be able to evolve it. Currently it is public and there is no way we can evolve it once published. E.g. you cannot remove/add fields to the struct if it is API. Using a constructor method (factory method in Java land) decouples implementation from API. We could make the constructor method take a struct with all parameters (just for data transfer purposes/convenience), but we should not make the object, that implements the API public IMHO. WDYT?

statsd/aggregator.go
ash2k: I renamed this parameter because of how these tags are used. They are only applied to statsd's own metrics hence "system". Default tags that we add to all user-generated metrics are passed to `NewMetricReceiver` and are called just `tags`. So I just wanted these two slightly different things have distinct names to avoid confusion. But I can rename them back :)

statsd/dispatcher.go
ash2k: FYI there is Idea plugin called Lines Sorter

statsd/dispatcher.go
ash2k: We could. I'd prefer to keep it as a struct to be able to easily add fields later if needed - making it a type alias does not buy us much in this case.

statsd/dispatcher.go
ash2k: Makes sense. Apparently gofmt is not enough and people still use different styles for no good reason :)

statsd/dispatcher.go
ash2k: `break` is an old good way to exit from block statements. Nothing is wrong with it :) Sometimes I prefer to extract loop into a function and do a `return` instead, but in this case it will make it uglier.

statsd/receiver.go
ash2k: Atomic counter is much faster than creating a metric object and processing it (hopefully just a [lock xadd](http://x86.renejeschke.de/html/file_module_x86_id_327.html) instruction if compiler authors did a good job :) ). Current implementation effectively processes ~2 additional metrics per metric received - 66% overhead! (`packets_received` and `metrics_received` for each packet). This new implementation collects stats from aggregators and receiver and sends them in `Flusher`.

gostatsd.go
ash2k: We could just remove this profiling code. If/when we will want/need to do profiling we can always add it back (without committing) and use. Currently it looks like debugging statements that someone forgot to remove after a big debugging session :)

ash2k:
So I've added `--benchmark` mode to tester (also in `perf` branch). Ran it on m4.10xlarge (40 CPUs) and here are the weird results:
perf branch (i.e. master):
```
INFO[0001] total_num_stats: 209531
INFO[0002] total_num_stats: 210076
INFO[0003] total_num_stats: 215384
INFO[0004] total_num_stats: 249435
INFO[0005] total_num_stats: 213649
INFO[0006] total_num_stats: 240674
INFO[0007] total_num_stats: 227998
INFO[0008] total_num_stats: 246273
INFO[0009] total_num_stats: 234746
INFO[0011] total_num_stats: 273712
INFO[0012] total_num_stats: 228259
INFO[0013] total_num_stats: 252476
INFO[0014] total_num_stats: 249219
```
refactoring branch:
```
INFO[0001] total_num_stats: 714249
INFO[0002] total_num_stats: 823528
INFO[0003] total_num_stats: 810112
INFO[0004] total_num_stats: 796390
INFO[0005] total_num_stats: 772804
INFO[0006] total_num_stats: 798922
INFO[0007] total_num_stats: 820605
INFO[0008] total_num_stats: 807241
INFO[0009] total_num_stats: 793413
INFO[0010] total_num_stats: 807987
INFO[0011] total_num_stats: 853852
INFO[0012] total_num_stats: 820154
INFO[0013] total_num_stats: 820577
INFO[0014] total_num_stats: 827767
```

This is lower than on my laptop (but still this branch is much faster) and lower than previous results using network! So previously I used Go 1.6.1 to test this. Today updated to 1.6.2. Maybe there is a regression? I cannot find anything suspicious in the code. CPUs on the box are barely loaded. I tried to stop/start the box to get another ec2 instance - didn't help.

statsd/dispatcher.go
jtblin: I'm not talking about `break`, I'm talking about the label, `loop:` and that iiuc we break to go to the label ;)

statsd/aggregator.go
jtblin: The code has changed to much that I can't find where this is done anymore but it used to be the default tags that we pass here: `aggregator := NewMetricAggregator(backends, percentThresholds, s.FlushInterval, s.ExpiryInterval, s.MaxWorkers, s.DefaultTags)`.

statsd/dispatcher.go
ash2k: `continue` goes to the label https://golang.org/ref/spec#Continue_statements
`break` exits from the labeled block https://golang.org/ref/spec#Break_statements

statsd/aggregator.go
ash2k: I've extracted flushing logic into a separate object `Flusher` (flusher.go). It is responsible for periodic collection of metrics from
*  `Dispatcher` which in its turn collects them from all `Aggregator`s it manages
* `Receiver` which keeps counters of BadLines, PacketsReceived, MetricsReceived and time LastPacket.

Note that now each "component" is responsible for one thing and is reusable - dependencies are interfaces (can write tests for each one separately).
* `Aggregator` has no locking, no channels and may be used in any way
* `Dispatcher` manages pool of `Aggregator`s and dispatches metrics handled to it
* `Receiver` reads and parses packets from socket using multiple goroutines and calls `Dispatcher` to dispatch metrics to one of the `Aggregator`s.
* `Flusher` does... what I've said above ^

backend/backends/null/null.go
jtblin: The program reads from bottom to top, so you put the `init()` function at the bottom and then the next function in the stack just above, etc. Again don't need to change and I haven't done it consistently, look at the datadog backend for an example.

backend/backends/stdout/stdout.go
jtblin: The problem is not about the verbosity in the logs, it's about the verbosity in the code ;) We will probably never go and look at the errors when statsd is closed but we'll keep the verbosity in the code forever.

statsd/aggregator.go
jtblin: Yeah naming comes from the original author. Agreed with removing everywhere, just want to be consistent :)

SGTM for the private change.

statsd/aggregator.go
jtblin: Cool, but it's still the default tags that we pass, that they are currently only used for the internal metrics is a side effect and system has a different meaning to me so I think we should settle on `defaultTags`.

statsd/dispatcher.go
jtblin: OK, got it so it's that we can break from the `select` and `for loop`?

ash2k: Yep. Context signalled us that we should end whatever we are doing so we want to avoid scheduling more flush requests by breaking out of for/select.

backend/backends/null/null.go
ash2k: Ok, got it. Makes sense. Will follow this style in the future.

statsd/aggregator.go
ash2k: Yes, that is how we used to do. I changed it a bit, hence the rename. Let me walk you through.

* This is how we create an implementation for `AggregatorFactory` interface and pass it to `Dispatcher`:
```go
	// 1. Start the Dispatcher
	factory := agrFactory{
		percentThresholds: percentThresholds,
		flushInterval:     s.FlushInterval,
		expiryInterval:    s.ExpiryInterval,
		systemTags:        s.DefaultTags,
	}
	dispatcher := NewDispatcher(s.MaxWorkers, s.MaxQueueSize, &factory)
```
As you can see we take `DefaultTags` from configuration and pass them as `systemTags`.
* They are called `systemTags` at this stage because `agrFactory` will enrich them and pass to `NewMetricAggregator()`. These tags are applied to system metrics that we send per **each** aggregator (`num_stats` and `processing_time` per aggragator instance):
```go
func (af *agrFactory) Create() Aggregator {
	tags := make([]string, 0, len(af.systemTags)+1)
	tags = append(tags, af.systemTags...)
	tags = append(tags, fmt.Sprintf("aggregator_%d", af.workerNumber))
	af.workerNumber++
	return NewMetricAggregator(af.percentThresholds, af.flushInterval, af.expiryInterval, tags)
}
```
* So we can see metrics per-aggregator and totals.

I'm not insisting at all, I'll rename it. Just wanted to explain why I decided to rename. Default tags are applied to received metrics by receiver, not by aggregator. Aggregator only applies these tags to its own metrics (`num_stats` and `processing_time`).

rsabitov:
Guys, do you know why am I receiving all updates from this repository?

ash2k:
@rsabitov you are watching this repo, that's why :)

rsabitov:
No I'm not.

Darnok99:
@rsabitov it's maybe some armata thing because I'm also receiving updates god knows why

jtblin:
You need to change your notification settings to not watch automatically
all projects in the Atlassian organisation.

PR #5, opened 2016-05-10
Return an error instead of nil when parseline gets nil/empty input

ryancox:
This is a result of my first pass at go-fuzz'ing the lexer

### DIFF
--- a/statsd/receiver.go
+++ b/statsd/receiver.go
@@ -2,6 +2,7 @@ package statsd
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"net"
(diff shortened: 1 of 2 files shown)


statsd/receiver.go
```diff
@@ -188,10 +189,15 @@ func (mr *metricReceiver) getAdditionalTags(addr string) types.Tags {
 
 // ParseLine with lexer impl.
 func (mr *metricReceiver) parseLine(line []byte) (*types.Metric, error) {
+	if line == nil {
+		return nil, errors.New("Error parsing line. Byte array is nil")
+	}
```
ash2k: I don't think this check is necessary, is it?

statsd/receiver.go
```diff
@@ -188,10 +189,15 @@ func (mr *metricReceiver) getAdditionalTags(addr string) types.Tags {
...
+
 	llen := len(line)
 	if llen == 0 {
-		return nil, nil
+		return nil, errors.New("Error parsing line. Byte array is empty")
```
ash2k: I don't think it is possible to hit this branch in the current codebase - there is a check before this function is invoked:
```go
if len(line) > 1 {
    metric, err := mr.parseLine(line)
```
I agree the code is not ideal (a redundant check inside of the function) and I've cleanup it up already in another branch.

ash2k:
Thanks for doing this! Can you share the input data corpus you used? Tooling/code you wrote? I'd like to include it in the build process.

ryancox:
I have a WIP [fuzz branch](https://github.com/atlassian/gostatsd/compare/master...ryancox:fuzz) that includes a few items in the corpus as well as a new 'fuzz' target in the Makefile. Regarding the above questions about defensive checking, you will see if you look at that branch that there is code that does round-tripping using a call to parseFile. Without those checks, go-fuzz reports a crasher.

ash2k:
I've merged my branch where this piece is refactored. Thanks for the effort! I'll take a look at your branch when I have time. Really keen to add fuzz testing to this project.

PR #7, opened 2016-05-20
Remove deprecated -f

ash2k:
I am not sure why it is needed and cannot find documentation on this flag. /cc @jtblin

```sh
docker tag -f atlassianlabs/gostatsd:$(git rev-parse --short HEAD) atlassianlabs/gostatsd:latest
Warning: '-f' is deprecated, it will be removed soon. See usage.
```


jtblin:
It used to be needed to force overwriting an existing tag. Maybe not needed anymore.

PR #8, opened 2016-05-21
Datadog events support

ash2k:


Makefile
```diff
@@ -83,8 +83,10 @@ cross:
 docker: cross
 	cd build && docker build -t $(IMAGE_NAME):$(GIT_HASH) .
 
-release: check test docker
+release-hash: check test docker
 	docker push $(IMAGE_NAME):$(GIT_HASH)
+
+release: check test docker
```
ash2k: I thought that pushing images with hash as tag will create too many tags - created a separate `release-hash` target to do it. Also it is convenient if you want to test some commit without actually doing a release.

backend/backend.go
```diff
@@ -24,23 +23,10 @@ var backends = map[string]backendTypes.Factory{
...
-// SendMetrics calls f(m).
-func (f MetricSenderFunc) SendMetrics(m *types.MetricMap) error {
-	return f(m)
-}
-
-// BackendName returns the name of the backend.
-func (f MetricSenderFunc) BackendName() string {
-	return "MetricSenderFunc"
-}
-
```
ash2k: `backendTypes.Backend` has two methods now so cannot use a function adapter.

backend/types/types.go
```diff
@@ -4,17 +4,20 @@ import (
...
-type MetricSender interface {
+// Backend represents a backend.
+type Backend interface {
 	// BackendName returns the name of the backend.
 	BackendName() string
 	// SampleConfig returns the sample config for the backend.
 	SampleConfig() string
 	// SendMetrics flushes the metrics to the backend.
-	SendMetrics(*types.MetricMap) error
+	SendMetrics(context.Context, *types.MetricMap) error
```
ash2k: context will be useful to have later.

statsd/lexer.go
```diff
@@ -9,29 +10,47 @@ import (
 )
 
 type lexer struct {
-	input     []byte
-	len       int
-	start     int
-	pos       int
```
ash2k: 64 bits are unnecessary.

statsd/receiver.go
```diff
@@ -145,21 +143,34 @@ func (mr *metricReceiver) handleMessage(ctx context.Context, addr net.Addr, msg
...
 
 		if readerr == io.EOF {
 			// if was EOF, finished handling
-			atomic.AddUint64(&mr.metricsReceived, uint64(numMetrics))
-			return nil
+			break
 		}
 	}
+	atomic.AddUint64(&mr.metricsReceived, uint64(numMetrics))
+	atomic.AddUint64(&mr.eventsReceived, uint64(numEvents))
```
ash2k: This change is to make sure we increment counters even in case of early exit (i.e. due to error).

statsd/receiver.go
```diff
@@ -186,14 +197,8 @@ func (mr *metricReceiver) getAdditionalTags(addr string) types.Tags {
...
 }
 
-// ParseLine with lexer impl.
-func (mr *metricReceiver) parseLine(line []byte) (*types.Metric, error) {
-	llen := len(line)
-	if llen == 0 {
-		return nil, nil
-	}
-	metric := &types.Metric{}
-	metric.Tags = append(metric.Tags, mr.tags...)
```
ash2k: We don't know if it will be metric or event so we add tags in the method above ^
Other things are handled by lexer itself.

statsd/receiver_test.go
```diff
@@ -10,122 +9,11 @@ import (
 	"golang.org/x/net/context"
 )
 
-func TestParseLine(t *testing.T) {
```
ash2k: Tests for lexer are in separate file now.

statsd/statsd.go
```diff
@@ -241,6 +245,26 @@ func (s *Server) RunWithCustomSocket(ctx context.Context, sf SocketFactory) erro
...
+	backends   []backendTypes.Backend
+}
+
+func (h *handler) DispatchMetric(ctx context.Context, m *types.Metric) error {
+	return h.dispatcher.DispatchMetric(ctx, m)
+}
+
+func (h *handler) DispatchEvent(ctx context.Context, e *types.Event) error {
+	for _, backend := range h.backends {
+		go func() {
```
ash2k: Ideally we want to have some sort of limit on the number of goroutines we want to spawn concurrently even if they are cheap to avoid OOM/etc in case of event flood. Later :)

Makefile
jtblin: Tags are cheap and are the only immutable entities in docker so I don't think we should remove it from the release (OK to have it as an extra for testing). Moreover `release` below depends on the image with the tag to be created... `docker tag $(IMAGE_NAME):$(GIT_HASH) $(IMAGE_NAME):latest`.

backend/backends/statsdaemon/statsdaemon.go
```diff
@@ -119,19 +120,24 @@ func (client *client) SendMetrics(metrics *types.MetricMap) error {
 	return nil
 }
 
+// SendEvent discards events.
```
jtblin: I don't think it should discard the event, the idea with this backend is that the front rows of statsd server will send to the statsd master which will have the backend logics. So it should send the event to the master.

backend/backends/stdout/stdout.go
```diff
@@ -89,6 +90,11 @@ func (client client) SendMetrics(metrics *types.MetricMap) (retErr error) {
 	return err
 }
 
+// SendEvent discards events.
+func (client client) SendEvent(ctx context.Context, e *types.Event) error {
```
jtblin: It should log the event to stdout like any other type of metrics.

backend/types/types.go
jtblin: OOC why? Isn't the context just to cancel, etc.?

statsd/lexer_test.go
```diff
@@ -0,0 +1,184 @@
+package statsd
```
jtblin: Nice!

statsd/lexer.go
```diff
@@ -41,35 +60,57 @@ func (l *lexer) next() byte {
...
-	return l.m, nil
+	return l.m, l.e, nil
 }
 
 type stateFn func(*lexer) stateFn
 
+// check the first byte for special Datadog type.
+func lexSpecial(l *lexer) stateFn {
+	switch b := l.next(); b {
+	case '_':
```
jtblin: Is that tested only for the 1st characters? I don't remember the full logic of the lexer.

statsd/receiver.go
```diff
@@ -145,21 +143,34 @@ func (mr *metricReceiver) handleMessage(ctx context.Context, addr net.Addr, msg
 				triedToGetTags = true
 				additionalTags = mr.getAdditionalTags(addr.String())
 			}
-			if len(additionalTags) > 0 {
+			if metric != nil {
+				numMetrics++
+				metric.Tags = append(metric.Tags, mr.tags...)
 				metric.Tags = append(metric.Tags, additionalTags...)
+				err = mr.handler.DispatchMetric(ctx, metric)
+			} else {
```
jtblin: Nit: `else if event != nil`?

Makefile
ash2k: Tags are not immutable, but ok, I'll put it back.

backend/backends/statsdaemon/statsdaemon.go
ash2k: You are absolutely right! Will fix this.

backend/types/types.go
ash2k: Yes, so if we need to cancel the send, we could. We will pass context to http library/etc when it is supported (should be in 1.7 I think?).

Makefile
jtblin: No it's not immutable but commit hash are unique for a repository so unless someone is really trying there is no risk that the tag will be overwritten in docker so you can have a high degree of confidence that this is the version that you wanted vs other tags like latest or version numbers that can be easily overwritten by mistakes.

statsd/lexer.go
ash2k: Here we switch between two ways we parse the packet - it is either an event (starts with `_`) or a metric (everything else).

backend/types/types.go
jtblin: Sure, I just don't find this supra "useful" but no worries with passing it ;)

statsd/receiver.go
ash2k: Ok, in that case I'll add an `else` statement as well with a panic to catch bugs. Good idea.

statsd/lexer.go
jtblin: Yeah I understand that, I was just unsure if that is done only for the 1st character. If it is not only for the 1st character then it would be a problem with metrics with an _ in the metric name. But I see we have a test for such metric names so I assume it's fine.

ash2k: Yes, this switch is only executed once per line at the very beginning of the parsing process.

statsd/receiver.go
jtblin: Maybe a panic is a tag aggressive ;)

PR #20, opened 2016-05-28
Send data async

ash2k:
The main idea is to avoid cloning all metrics to generate less garbage.

`master` is the updated `statsd_test`:
```
time="2016-05-28T18:25:31+10:00" level=info msg="numStats: 1182124"
time="2016-05-28T18:25:32+10:00" level=info msg="numStats: 1283931"
time="2016-05-28T18:25:33+10:00" level=info msg="numStats: 1289507"
time="2016-05-28T18:25:34+10:00" level=info msg="numStats: 1255173"
time="2016-05-28T18:25:35+10:00" level=info msg="numStats: 1297849"
time="2016-05-28T18:25:36+10:00" level=info msg="numStats: 1291102"
time="2016-05-28T18:25:37+10:00" level=info msg="numStats: 1260675"
time="2016-05-28T18:25:38+10:00" level=info msg="numStats: 1003565"
time="2016-05-28T18:25:39+10:00" level=info msg="numStats: 1192724"
time="2016-05-28T18:25:41+10:00" level=info msg="numStats: 1195525"
time="2016-05-28T18:25:42+10:00" level=info msg="numStats: 1131726"
time="2016-05-28T18:25:43+10:00" level=info msg="numStats: 1136089"
time="2016-05-28T18:25:44+10:00" level=info msg="numStats: 1142574"

	statsd_test.go:49: Processed metrics: 1061381
		TotalAlloc: 9932219136 (9357 per metric)
		Mallocs: 236540591 (222 per metric)
		NumGC: 385
		GCCPUFraction: 0.163982
```

This branch:
```
time="2016-05-28T18:23:25+10:00" level=info msg="numStats: 1214117"
time="2016-05-28T18:23:26+10:00" level=info msg="numStats: 1307186"
time="2016-05-28T18:23:27+10:00" level=info msg="numStats: 1324007"
time="2016-05-28T18:23:28+10:00" level=info msg="numStats: 1299264"
time="2016-05-28T18:23:29+10:00" level=info msg="numStats: 1296963"
time="2016-05-28T18:23:30+10:00" level=info msg="numStats: 1304763"
time="2016-05-28T18:23:31+10:00" level=info msg="numStats: 1315937"
time="2016-05-28T18:23:33+10:00" level=info msg="numStats: 1294777"
time="2016-05-28T18:23:34+10:00" level=info msg="numStats: 1286428"
time="2016-05-28T18:23:35+10:00" level=info msg="numStats: 1305129"
time="2016-05-28T18:23:36+10:00" level=info msg="numStats: 1283802"
time="2016-05-28T18:23:37+10:00" level=info msg="numStats: 1251706"
time="2016-05-28T18:23:38+10:00" level=info msg="numStats: 1255325"
time="2016-05-28T18:23:39+10:00" level=info msg="numStats: 1155954"

	statsd_test.go:49: Processed metrics: 1155885
		TotalAlloc: 10451872568 (9042 per metric)
		Mallocs: 256015955 (221 per metric)
		NumGC: 410
		GCCPUFraction: 0.156066
```
Note that:
1. More metrics were processed in 15 seconds;
2. `TotalAlloc` is bigger because we managed to churn through more metrics. Per metric alloc is smaller 9357 vs 9042;
3. `Mallocs` is more or less the same... hm;
4. `GCCPUFraction` is less (fraction of CPU time used by GC)

### DIFF
--- a/backend/backends/datadog/datadog.go
+++ b/backend/backends/datadog/datadog.go
@@ -105,6 +105,22 @@ func (d *client) SendMetrics(ctx context.Context, metrics *types.MetricMap) erro
 	if metrics.NumStats == 0 {
 		return nil
 	}
+	return d.postMetrics(d.prepareSeries(metrics))
+}
+
+// SendMetricsAsync flushes the metrics to Datadog, preparing payload synchronously but doing the send asynchronously.
+func (d *client) SendMetricsAsync(ctx context.Context, metrics *types.MetricMap, c backendTypes.SendCallback) {
+	if metrics.NumStats == 0 {
+		c(nil)
+		return
+	}
+	ts := d.prepareSeries(metrics)
+	go func() {
+		c(d.postMetrics(ts))
+	}()
+}
+
+func (d *client) prepareSeries(metrics *types.MetricMap) *timeSeries {
 	ts := timeSeries{Timestamp: time.Now().Unix(), Hostname: d.hostname}
 
 	metrics.Counters.Each(func(key, tagsKey string, counter types.Counter) {
@@ -134,7 +150,10 @@ func (d *client) SendMetrics(ctx context.Context, metrics *types.MetricMap) erro
 	metrics.Sets.Each(func(key, tagsKey string, set types.Set) {
 		ts.addMetric(key, tagsKey, gauge, float64(len(set.Values)), set.Flush)
 	})
+	return &ts
+}
 
+func (d *client) postMetrics(ts *timeSeries) error {
 	return d.post("/api/v1/series", "metrics", ts)
 }
 
@@ -182,7 +201,7 @@ func (d *client) post(path, typeOfPost string, data interface{}) error {
 	return nil
 }
 
-func (d *client) doPost(path string, body []byte) func() error {
+func (d *client) doPost(path string, body []byte) backoff.Operation {
 	authenticatedURL := d.authenticatedURL(path)
 	return func() error {
 		req, err := http.NewRequest("POST", authenticatedURL, bytes.NewBuffer(body))
--- a/backend/backends/datadog/datadog.go
+++ b/backend/backends/datadog/datadog.go
@@ -109,14 +109,14 @@ func (d *client) SendMetrics(ctx context.Context, metrics *types.MetricMap) erro
 }
 
 // SendMetricsAsync flushes the metrics to Datadog, preparing payload synchronously but doing the send asynchronously.
-func (d *client) SendMetricsAsync(ctx context.Context, metrics *types.MetricMap, c backendTypes.SendCallback) {
+func (d *client) SendMetricsAsync(ctx context.Context, metrics *types.MetricMap, cb backendTypes.SendCallback) {
 	if metrics.NumStats == 0 {
-		c(nil)
+		cb(nil)
 		return
 	}
 	ts := d.prepareSeries(metrics)
 	go func() {
-		c(d.postMetrics(ts))
+		cb(d.postMetrics(ts))
 	}()
 }
 
--- a/backend/backends/statsdaemon/statsdaemon.go
+++ b/backend/backends/statsdaemon/statsdaemon.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"strconv"
 	"strings"
+	"sync"
 
 	backendTypes "github.com/atlassian/gostatsd/backend/types"
 	"github.com/atlassian/gostatsd/types"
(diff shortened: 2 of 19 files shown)


ash2k:
Instead of cloning in aggregator to avoid blocking it why not just generate payload (e.g. JSON) in aggregator goroutine and do the rest (e.g. http call) asynchronously? That way we avoid generating a lot of garbage and instead of spending time on cloning we spend that time to actually produce payload that we will be sending. Total amount of work performed is the same minus cloning (producing garbage) and minus GC (to remove that garbage).

backend/backends/datadog/datadog.go
```diff
@@ -105,6 +105,22 @@ func (d *client) SendMetrics(ctx context.Context, metrics *types.MetricMap) erro
...
+}
+
+// SendMetricsAsync flushes the metrics to Datadog, preparing payload synchronously but doing the send asynchronously.
+func (d *client) SendMetricsAsync(ctx context.Context, metrics *types.MetricMap, c backendTypes.SendCallback) {
+	if metrics.NumStats == 0 {
+		c(nil)
+		return
+	}
+	ts := d.prepareSeries(metrics)
+	go func() {
```
jtblin: Why not just `go c(d.postMetrics(ts))` ?

backend/backends/datadog/datadog.go
```diff
@@ -105,6 +105,22 @@ func (d *client) SendMetrics(ctx context.Context, metrics *types.MetricMap) erro
 	if metrics.NumStats == 0 {
 		return nil
 	}
+	return d.postMetrics(d.prepareSeries(metrics))
+}
+
+// SendMetricsAsync flushes the metrics to Datadog, preparing payload synchronously but doing the send asynchronously.
+func (d *client) SendMetricsAsync(ctx context.Context, metrics *types.MetricMap, c backendTypes.SendCallback) {
```
jtblin: Nit: `cb` would be better name than `c` imho.

backend/backends/statsdaemon/statsdaemon.go
```diff
@@ -76,49 +78,125 @@ func (client *client) SendMetrics(ctx context.Context, metrics *types.MetricMap)
...
+		case datagrams <- buf:
+			return bufFree.Get().(*bytes.Buffer), nil
+		}
+	})
+	if err == nil {
+		// All metrics sent to the channel and context wasn't cancelled (yet) - consuming goroutine will exit
+		// with success (unless ctx gets a cancel after the close, which is also ok)
+		close(datagrams)
+	} else if err != context.Canceled && err != context.DeadlineExceeded {
+		log.Panicf("Unexpected error: %v", err)
```
jtblin: I don't think we should `panic` at runtime here.

backend/backends/statsdaemon/statsdaemon.go
```diff
@@ -76,49 +78,125 @@ func (client *client) SendMetrics(ctx context.Context, metrics *types.MetricMap)
...
+func processMetrics(metrics *types.MetricMap, handler overflowHandler) (retErr error) {
+	type failure struct {
+		err error
+	}
+	defer func() {
+		if r := recover(); r != nil {
+			if f, ok := r.(failure); ok {
+				retErr = f.err
+			} else {
+				panic(r)
```
jtblin: Same here and below about `panic`.

ash2k: Initially I wrote it using usual approach (return `error`, check with `if`) but it was ugly and I had to introduce another iteration method `EachWhile()` for counters, timers, gauges and sets. I decided to try to apply the approach with `panic`/`recover` described here https://blog.golang.org/defer-panic-and-recover and it looks MUCH better.

This particular line just re-panics a recovered panic in case it was something else that panicked, not our code (e.g. division by zero somewhere).

backend/backends/statsdaemon/statsdaemon.go
ash2k: Yes, I would prefer to structure code here to avoid having to do this. This `if` just checks the invariant - `processMetrics()` should return `nil` or `context.Canceled` or `context.DeadlineExceeded`. Other error are not possible because... there is no source of other errors. So IF we encounter something else, it is a bug in code because our assumption is wrong.  I haven't found (yet) a better way to structure code here to avoid this ugly condition. I could just remove it but I'd rather not - it is a case that we must catch if it happens for some reason.

backend/backends/datadog/datadog.go
ash2k: As per language spec https://golang.org/ref/spec#Go_statements:
```
The function value and parameters are evaluated as usual in the calling goroutine, but unlike with a regular call, program execution does not wait for the invoked function to complete.
```
I don't want `d.postMetrics(ts)` to be evaluated in the calling goroutine - the point is to execute it in a new goroutine. So it must be wrapped in a func.

PR #33, opened 2016-06-10
Reuse HTTP connections in Datadog backend

ash2k:
Consume HTTP response to free up connection and make it reusable.

### DIFF
--- a/backend/backends/datadog/datadog.go
+++ b/backend/backends/datadog/datadog.go
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io"
+	"io/ioutil"
 	"net/http"
 	"net/url"
 	"os"
@@ -29,6 +31,8 @@ const (
 	defaultClientTimeout         = 9 * time.Second
 	// defaultMetricsPerBatch is the default number of metrics to send in a single batch.
 	defaultMetricsPerBatch uint16 = 1000
+	// maxResponseSize is the maximum response size we are willing to read.
+	maxResponseSize = 10 * 1024
 	// gauge is datadog gauge type.
 	gauge = "gauge"
 	// rate is datadog rate type.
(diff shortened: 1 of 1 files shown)


backend/backends/datadog/datadog.go
```diff
@@ -257,10 +261,13 @@ func (d *client) doPost(path string, body []byte) backoff.Operation {
 			return fmt.Errorf("error POSTing: %s", strings.Replace(err.Error(), d.apiKey, "*****", -1))
 		}
 		defer resp.Body.Close()
-
+		body := io.LimitReader(resp.Body, maxResponseSize)
```
jtblin: OOC, how does that allow to reuse the HTTP connections?

ash2k: From https://golang.org/src/net/http/response.go?s=512:3205#L19 :
```
    46		// Body represents the response body.
    47		//
    48		// The http Client and Transport guarantee that Body is always
    49		// non-nil, even on responses without a body or responses with
    50		// a zero-length body. It is the caller's responsibility to
    51		// close Body. The default HTTP client's Transport does not
    52		// attempt to reuse HTTP/1.0 or HTTP/1.1 TCP connections
    53		// ("keep-alive") unless the Body is read to completion and is
    54		// closed.
    55		//
    56		// The Body is automatically dechunked if the server replied
    57		// with a "chunked" Transfer-Encoding.
    58		Body io.ReadCloser
```
We were closing the Body but Datadog sends some response (something like `{"status": "ok"}`) and we were not reading it. So the transport was not able to reuse the connection. It is very similar to how Java's URLConnection works - you have to consume the response, not just close it, otherwise it will not reuse the connection and just close it instead.

http://www.ceus-now.com/reusing-http-connections-in-golang/
https://docs.oracle.com/javase/8/docs/technotes/guides/net/http-keepalive.html

PR #39, opened 2016-06-13
Add missing statement to Makefile

rtkmhart:
Starting on a fresh VM, I needed to add this line to the Makefile to get gostatsd to build properly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,7 @@ METALINTER_CONCURRENCY ?= 4
 setup: setup-ci
 	go get -v -u github.com/githubnemo/CompileDaemon
 	go get -v -u github.com/jstemmer/go-junit-report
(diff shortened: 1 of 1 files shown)


ash2k:
What commands you are executing and what output you are getting? This should not be required because dependency on `logrus` in declared in `glide.yaml` and `glide install` invoked from `make setup` should pull it into `vendor` directory.

rtkmhart:
It was a "make setup" that failed, on a freshly provisioned Ubuntu 12.04 LTS VM with go version 1.6.2. It's quite possible I messed something up.

ash2k:
Could you provide a full output? It works in Travis builds so... I'm not sure what is wrong.

ash2k:
Please reopen with more information if it is still relevant.

PR #48, opened 2016-06-26
Add maxConcurrency argument to NewDispatchingHandler to rate limit active goroutines

### LINKED ISSUE #24: Rate limiting for event dispatching
Currently a goroutine per backend is created for each incoming event and there is no limit on how many we can start. To avoid running out of memory and/or saturating backend a rate limiting should be implemented in event dispatcher. It should be backend-agnostic.

mark-adams:
Fixes #24.

Previously, the number of goroutines created by calls to DispatchEvent()
was unrestricted. This commit adds a new argument to
NewDispatchingHandler, backed by a new dispatchLimiter struct, that
limits the number of goroutines that DispatchEvent() can have in flight
at one time.

This is based heavily off the design of LimitListener from golang/net

TODO:

- [ ] Write tests
- [ ] Implement configuration mechanism for specifying maxConcurrency

Before completing the remaining tasks, I'd love to get some design feedback.

### DIFF
--- a/statsd/dispatching_handler.go
+++ b/statsd/dispatching_handler.go
@@ -16,14 +16,17 @@ type dispatchingHandler struct {
 	dispatcher Dispatcher
 	backends   []backendTypes.Backend
 	tags       types.Tags // Tags to add to all metrics and events
+
+	limiter *dispatchLimiter // Limit # of goroutines for dispatching events
 }
 
 // NewDispatchingHandler initialises a new dispatching handler.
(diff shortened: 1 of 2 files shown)


ash2k:
There should be a buffered queue in front of the semaphore to accomodate some bursts of traffic. Also I'd prefer direct use of channel to do limiting in a `select` i.e.
```go
select {
case sem <- struct{}{}:
    go dispatchToBackend(...);
default:
    select {
    case <-ctx.Done():
        ...
    case sem <- struct{}{}:
        go dispatchToBackend(...);
    case buferedQueue <- event:
    }
}
```

ash2k:
Fixed in #54.

PR #50, opened 2016-06-30
Push docker image on successful tag build

aelse:
Release to docker on tagged builds.

Should work on lightweight tags and annotated tags (git tag -a). Travis-ci may not build again for lightweight tags since the commit hash does not change, so suggest we use annotated tags only.

Relevant travis-ci docs:
* https://docs.travis-ci.com/user/docker/#Branch-Based-Registry-Pushes
* https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables

See build of my branch:
* https://travis-ci.org/aelse/gostatsd/jobs/141248214
* https://hub.docker.com/r/aelse/gostatsd/tags/

I've added the DOCKER_* env vars to atlassian/gostatsd on travis-ci.
@ash2k

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,9 @@
+sudo: required
+
 language: go
+
+services:
+  - docker
 
 go:
   - 1.6.2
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,5 +1,10 @@
+sudo: required
```
ash2k: Why do we need it?

aelse: I'm not sure. travis-ci docs tell you to add it but do not invoke sudo directly (except in the docker-compose example at the very end). Might be a peculiarity of how they do builds.

https://docs.travis-ci.com/user/docker/#Branch-Based-Registry-Pushes

ash2k: Oh, ok, I guess they do not support docker-in-docker :)

PR #63, opened 2016-09-20
Use Go 1.7 and built-in context package

ash2k:
/cc @aelse


glide.lock
```diff
@@ -67,7 +67,7 @@ imports:
   subpackages:
   - assert
 - name: golang.org/x/net
-  version: c4c3ea71919de159c9e246d7be66deb7f0a39a58
+  version: 71a035914f99bb58fe82eac0f1289f10963d876c
   subpackages:
   - context
```
aelse: Can remove x/net/context from glide.lock

ash2k: I tried removing it but golang.org/x/time/rate dependency needs it.

PR #64, opened 2016-10-12
Update dependencies

ash2k:
The next step would be to remove some workarounds for Viper bugs that were fixed recently upstream - that is why I want to update dependencies in the first place.


ash2k:
@aelse

PR #61, opened 2016-09-06
Implemented ability to disable tags for statsdaemon backend #60

makhmutov:
Please review PR for fix #60.
Let me know if you want any changes.

PR #68, opened 2016-10-18
Add config option for etsy compatibility

natemueller:
Etsy's statsd daemon doesn't understand tags or events


ash2k:
Yeah, this was fixed in #61

PR #69, opened 2016-10-18
Add support for TCP in the statsdaemon backend

natemueller:


ash2k:
Thanks! I have a concerns here.
With UDP we re-dial every time to re-resolve DNS record because it may have changed and there is no way to "notice" this. With TCP it's different - if backend goes down (because it was redeployed and DNS record was updated to point at the new ip) the TCP connection will be reset and the app will be forced to re-establish it, re-resolving the DNS record. Consequently the new ip will be picked up. So for TCP there is no need to re-establish the connection for each send operation. It is pure overhead. WDYT? Some kind of connection "caching" may make sense. Actually, we should probably have just a single connection for TCP case.

p.s. also tests seem to fail.

natemueller:
To me the simplicity of this approach outweighs the cost.  I'm somewhat concerned with how the node process would handle a very large number of long-lived connections.  This approach also makes load balancing simple.  We don't use that today but like having the option.

These tradeoffs make sense for our deployment (we run this daemon on all our servers and push stats every 10s to a central etsy statsd service) but I don't know how typical that is and if this would cause big problems for other people.  What do you think?

I'll take a look at the tests once we agree on desired behavior.

ash2k:
> I'm somewhat concerned with how the node process would handle a very large number of long-lived connections.

I think it is much better performance-wise to have a large number of persistent connections than to have the same number of connections but being re-established every 10s. Also I don't think the implementation will be too complicated.

> This approach also makes load balancing simple. We don't use that today but like having the option.

We can always introduce an option to limit the max number of "packets" submitted over a single TCP connection.

I'll try to come up with a PR myself to try a different approach.

PR #71, opened 2016-10-22
Support TCP transport in statsd backend

ash2k:
- Reuse TCP and UDP connections
- Limit amount of data streamed over single connection
- Reuse memory buffers in more cases
- Support write timeouts in statsd backend


ash2k:
@natemueller WDYT?

/cc @aelse

natemueller:
+1

PR #72, opened 2016-10-24
Use context with AWS SDK

ash2k:
https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/common-examples.html#id10


ash2k:
@aelse

PR #74, opened 2016-10-26
Enable HTTP2

ash2k:
HTTP2 is disabled if TLSClientConfig/Dial/DialTLS is specified on Transport
See https://github.com/golang/go/issues/14275


ash2k:
@aelse

PR #76, opened 2016-10-28
Add missing Run method to Graphite

### LINKED ISSUE #75: version 0.15.0 locks up after one flush window
We're trying gostatsd.  Under load (around 40k incoming metrcs per second) version 0.14.0 recorded it was receiving around 10k metrics per flush window (60s):

```
DEBU[0600] numStats: 10571 packetsReceived: 10537
```

I just tried the new release (0.15) and it locked up after one flush window. That is no more metrics were sent and the app didn't stop when CTRL-C was pressed.  I had to CTRL-Z and `kill` the  pid.

```
./gostatsd_0.15 --flush-interval 60s --verbose --config-path "./config.toml" --metrics-addr "0.0.0.0:8125"                          INFO[0000] Starting server
INFO[0000] No cloud provider specified
INFO[0000] [graphite] address=10.0.0.202:2013 dialTimeout=5s writeTimeout=30s
INFO[0000] Initialised backend "graphite"
DEBU[0060] Sending 2054 metrics to backend graphite
DEBU[0060] Sending 3320 metrics to backend graphite
DEBU[0060] Sending 6079 metrics to backend graphite
DEBU[0060] Sending 1848 metrics to backend graphite
```

config.toml:

```
[graphite]
        address = "10.0.0.202:2013"
        legacy_namespace = false
        global_prefix = "stats"
        prefix_timer = ""
        prefix_counter = ""
```
`10.0.0.202:2013` is carbon relay 

Any idea where I should look to debug this?

We're running on ubuntu 14.04

ash2k:
Fixes #75

### DIFF
--- a/backend/backends/graphite/graphite.go
+++ b/backend/backends/graphite/graphite.go
@@ -85,6 +85,10 @@ type client struct {
 	legacyNamespace  bool
 }
 
+func (client *client) Run(ctx context.Context) error {
+	return client.sender.Run(ctx)
+}
+
 // SendMetricsAsync flushes the metrics to the Graphite server, preparing payload synchronously but doing the send asynchronously.
 func (client *client) SendMetricsAsync(ctx context.Context, metrics *types.MetricMap, cb backendTypes.SendCallback) {
 	if metrics.NumStats == 0 {
@@ -99,7 +103,6 @@ func (client *client) SendMetricsAsync(ctx context.Context, metrics *types.Metri
 	case <-ctx.Done():
 		client.sender.PutBuffer(buf)
 		cb([]error{ctx.Err()})
-		return
 	case client.sender.Sink <- backends.Stream{Cb: cb, Buf: sink}:
 	}
 }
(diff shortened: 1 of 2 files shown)


backend/backends/graphite/graphite.go
```diff
@@ -85,6 +85,10 @@ type client struct {
 	legacyNamespace  bool
 }
 
+func (client *client) Run(ctx context.Context) error {
+	return client.sender.Run(ctx)
+}
```
ash2k: I've added this method to statsd backend but forgot to add it here. Sender wasn't sending :-\

jtblin: Couldn't there be an interface somewhere to prevent that issue from happening?

ash2k: As you know interfaces are used in the "consumer" bit of code, not when methods are defined (unlike e.g. Java). I could have done the following as some projects do but I think this is not idiomatic because this is by design that you do not specify which interfaces you "implement".
```go
var _ backendTypes.RunnableBackend = &client{}
```

jtblin: Maybe write an "integration" test that exercises the interface, that will prevent similar error in the future. Not sure if it's idiomatic but it's more robust.

PR #80, opened 2016-11-24
Restructure

ash2k:
Restructure project into `cmd` and `pkg` directories. Also extract types to the top level and simplify cloud provider and backends package layout.

No functional changes, just a nicer and standard layout. Please review one commit at a time :)


ash2k:
@aelse

PR #87, opened 2017-01-21
Support for TLS in the statsdaemon backend

eriksw:
This is a works-for-me implementation of support for using TLS to secure the transmission of metrics over the statsd TCP protocol.

I'd very much like to get this merged eventually, including support in the receiver as well, but can't really make much progress on that for lack of support for TCP in the receiver. Were that to exist, this would have tests that exercise it end-to-end.

Are there any plans for or interest in TCP receiver support?

(If and when it's the remaining barrier for this being merged, getting the corporate CLA agreement taken care of shouldn't be an issue.)

### DIFF
--- a/pkg/backends/statsdaemon/statsdaemon.go
+++ b/pkg/backends/statsdaemon/statsdaemon.go
@@ -3,6 +3,7 @@ package statsdaemon
 import (
 	"bytes"
 	"context"
+	"crypto/tls"
 	"fmt"
 	"net"
 	"strconv"
@@ -12,8 +13,6 @@ import (
 
 	"github.com/atlassian/gostatsd"
 	"github.com/atlassian/gostatsd/pkg/backends/sender"
-
-	"crypto/tls"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/spf13/viper"
--- a/pkg/backends/statsdaemon/statsdaemon.go
+++ b/pkg/backends/statsdaemon/statsdaemon.go
@@ -13,6 +13,8 @@ import (
 	"github.com/atlassian/gostatsd"
 	"github.com/atlassian/gostatsd/pkg/backends/sender"
 
+	"crypto/tls"
+
 	log "github.com/Sirupsen/logrus"
 	"github.com/spf13/viper"
 )
@@ -199,7 +201,7 @@ func constructEventMessage(e *gostatsd.Event) *bytes.Buffer {
 }
 
 // NewClient constructs a new statsd backend client.
-func NewClient(address string, dialTimeout, writeTimeout time.Duration, disableTags, tcpTransport bool) (*Client, error) {
+func NewClient(address string, dialTimeout, writeTimeout time.Duration, disableTags, tcpTransport bool, tlsConfig *tls.Config) (*Client, error) {
 	if address == "" {
 		return nil, fmt.Errorf("[%s] address is required", BackendName)
 	}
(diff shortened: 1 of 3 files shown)


ash2k:
We are not interested in TLS support ourselves but I don't think there are any reasons why we would not add support if you send PRs.

We can start by merging this in once you sign the CLA.

pkg/backends/statsdaemon/statsdaemon.go
```diff
@@ -13,6 +13,8 @@ import (
 	"github.com/atlassian/gostatsd"
 	"github.com/atlassian/gostatsd/pkg/backends/sender"
 
+	"crypto/tls"
```
ash2k: Please put it together with other stdlib imports

pkg/backends/statsdaemon/tls.go
```diff
@@ -0,0 +1,56 @@
...
+func getTLSConfiguration(caPath, certPath, keyPath string, enable bool) (*tls.Config, error) {
+	if !enable {
+		return nil, nil
+	}
+
+	tlsConfig := &tls.Config{
+		MinVersion: tls.VersionTLS12,
+		// Rationale:
+		// * Retain performance-conscious ordering from the crypto/tls defaults
+		// * Reject deprecated ciphers and modes, including non-DHE
```
ash2k: Hm, this is interesting. I'd like us to either omit cipher configuration completely or specify it for all other HTTPS clients too (AWS and Datadog IIRC).

pkg/backends/statsdaemon/tls.go
```diff
@@ -0,0 +1,56 @@
...
+			tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
+			tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
+			tls.TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,
+		},
+	}
+
+	if caPath != "" {
+		caPEM, err := ioutil.ReadFile(caPath)
+		if err != nil {
+			return nil, fmt.Errorf("[%s] error reading tls ca: %v", BackendName, err)
```
ash2k: Please use upper case for `TLS` and `CA` everywhere in text strings.

pkg/backends/statsdaemon/tls.go
```diff
@@ -0,0 +1,56 @@
...
+		}
+
+		tlsConfig.RootCAs = x509.NewCertPool()
+		if ok := tlsConfig.RootCAs.AppendCertsFromPEM(caPEM); !ok {
+			return nil, fmt.Errorf("[%s] error reading tls ca: no certs found", BackendName)
```
ash2k: `certs` -> `certificate` to be consistent with the error message below.

pkg/backends/statsdaemon/tls.go
ash2k: I suggest we omit it from this PR and add in a separate PR later for other places too. WDYT?

eriksw: Sure. Since MinVersion is configured for AWS and Datadog already, I'm leaving that in place and copying the rationale comment from those places.

eriksw:
Comments addressed. Additionally, a corporate CLA has been submitted by @natemueller.

ash2k:
I'm sorry for the delay, waiting for the confirmation that the CLA has been signed.

Issue #109, opened 2017-06-19
FakePacketConn.Close() is a noop


tiedotguy:
statsd_test uses a fake socket to generate packets, and depends on handlePacket to report an error before checking ctx for expiry.  As part of a larger refactoring I am doing, this behavior is changing and packet handling can no longer error, which removes all indication of "done" from the testing.

PR #110, opened 2017-06-19
#109: Make FakePacketConn closeable

tiedotguy:
Not currently consumed, but I'm aiming to keep my changes isolated and small.

### DIFF
--- a/pkg/fakesocket/fake.go
+++ b/pkg/fakesocket/fake.go
@@ -23,7 +23,7 @@ var ErrAlreadyClosedConnection = errors.New("Connection is already closed")
 
 // FakePacketConn is a fake net.PacketConn providing FakeMetric when read from.
 type FakePacketConn struct {
-	closed chan interface{}
+	closed chan struct{}
 }
 
 func (fpc *FakePacketConn) isClosed() bool {
@@ -112,7 +112,7 @@ func (frpc *FakeRandomPacketConn) ReadFrom(b []byte) (int, net.Addr, error) {
 func Factory() (net.PacketConn, error) {
 	frpc := &FakeRandomPacketConn{
 		FakePacketConn: FakePacketConn{
-			closed: make(chan interface{}),
+			closed: make(chan struct{}),
 		},
 	}
 	return frpc, nil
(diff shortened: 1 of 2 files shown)


pkg/fakesocket/fake.go
```diff
@@ -19,40 +18,73 @@ var FakeAddr = &net.UDPAddr{
 	Port: 8181,
 }
 
+var ErrClosedConnection = errors.New("Connection is closed")
+var ErrAlreadyClosedConnection = errors.New("Connection is already closed")
+
 // FakePacketConn is a fake net.PacketConn providing FakeMetric when read from.
-type FakePacketConn struct{}
+type FakePacketConn struct {
+	closed chan int
```
ash2k: Make it a `chan struct{}` to highlight that nothing valuable can be read from it. It is a convention.

tiedotguy: Nice.  Was not aware, will update.

ash2k: The convention is to use `struct {}` not empty interface - `struct {}` is a zero-sized value (as in its representation is 0 bytes of memory). Any value of interface type consumes memory of two pointers (in general; I think certain cases are/were optimized. Also I remember seeing something about removing this optimization for reasons).

It is not important in this case, but it is confusing to see interface there :)
An example where that matters is a buffered channel of `struct {}` - you can put `N` values in it but it does not consume memory to hold them, it's just a semaphore.

tiedotguy: Doh.  It's late.  My bad!

ash2k:
So, what are you planning to do with this? IIRC this fake impl is used in benchmarks and adding a synchronization primitive there does not help :)

tiedotguy:
I've not been able to get the receiver to scale past about 220kpps, unless I drop the packet as soon as I read it.  I suspect something is not waking up in timely fashion in the kernel for other readers, although I'm not sure.

The plan is to decouple reading / parsing, by reading packets and passing them to a channel for parsing.

Since the current fakesocket ReadFrom can never fail, as MetricReceiver.Receive is now just pushing to a channel, no error is received from MetricReceiver.handlePacket, and ctx is never checked.  The end result is that TestStatsdThroughput doesn't terminate.

I could have made ctx be checked every iteration, but really when a socket is closed, it should not allow reading of packets, so I pushed the check there.  It also means non-test has better perf.

PR #114, opened 2017-08-02
Disable HTTP/2 support in Datadog backend client

aelse:


ash2k:
Start with commenting out `http2.ConfigureTransport()` call which configures HTTP2 explicitly :)

tiedotguy:
This needs context on why, and should be behind a flag (arguably not even enabled by default).  If others are using it and not experiencing issues due to different setups, then this would negate the benefits that HTTP/2 would provide them.

aelse:
I've been chatting to @ash2k out of band about the problem and was going to add context before merging this. The short of it is this is related to current problems with gostatsd talking to a DD endpoint and the recommendation has been to disable HTTP/2 for the time being. I don't know the time frame on this, but it is likely that this affects anyone using gostatsd with Datadog, hence this PR to help everyone.

pkg/backends/datadog/datadog.go
```diff
@@ -279,10 +278,13 @@ func NewClient(apiEndpoint, apiKey string, metricsPerBatch uint, clientTimeout,
...
+		// A non-nil empty map used in TLSNextProto to disable HTTP/2 support in client.
+		// https://golang.org/doc/go1.6#http2
+		TLSNextProto: map[string](func(string, *tls.Conn) http.RoundTripper){},
 	}
-	if err := http2.ConfigureTransport(transport); err != nil {
-		return nil, err
-	}
+	//if err := http2.ConfigureTransport(transport); err != nil {
+	//	return nil, err
+	//}
```
ash2k: Just remove these lines and later we will just revert the commit to restore the functionality.

aelse: Ok. I'll turn this into a single commit.

PR #117, opened 2017-08-16
Refactor internal metrics to use a Statser

tiedotguy:
This touches on a whole heap of stuff, but the gist is that we currently have all these backdoory hooks, and adding more internal metrics means per-metric backdoory hooks.

This removes all of them, and replaces them with a Statser.  We still use use backdoory hooks to get the Statser where it needs to be, but then it's done.  It gives a better foundation to build internal metrics on.

It also gives us the facility to send to an alternate statsd service, which means we should retain metrics under failure-to-submit conditions.  Currently that's not implemented.  Configuring Statsers is also not implemented.

Metrics now send on their own interval, rather than being slaved to the MetricFlusher, because they may not be going to the current flusher.  Intervals are also not yet configurable.

It also splits out an internal metric namespace and internal metric tags.  If we were any other service we would have this control over our internal metrics, so I want them for this service too.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ GIT_HASH := $$(git rev-parse --short HEAD)
 GOBUILD_VERSION_ARGS := -ldflags "-s -X $(VERSION_VAR)=$(REPO_VERSION) -X $(GIT_VAR)=$(GIT_HASH) -X $(BUILD_DATE_VAR)=$(BUILD_DATE)"
 BINARY_NAME := gostatsd
 IMAGE_NAME := atlassianlabs/$(BINARY_NAME)
-ARCH ?= darwin
+ARCH ?= $$(uname -s | tr A-Z a-z)
 METALINTER_CONCURRENCY ?= 4
 GOVERSION := 1.8
 GP := /gopath
@@ -132,7 +132,7 @@ release-race: docker-race
 release: release-normal release-race
 
 run: build
-	./build/bin/$(ARCH)/$(BINARY_NAME) --backends=stdout --verbose --flush-interval=1s
+	./build/bin/$(ARCH)/$(BINARY_NAME) --backends=stdout --verbose --flush-interval=2s
 
 run-docker: docker
 	cd build/ && docker-compose rm -f gostatsd
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ GIT_HASH := $$(git rev-parse --short HEAD)
 GOBUILD_VERSION_ARGS := -ldflags "-s -X $(VERSION_VAR)=$(REPO_VERSION) -X $(GIT_VAR)=$(GIT_HASH) -X $(BUILD_DATE_VAR)=$(BUILD_DATE)"
 BINARY_NAME := gostatsd
 IMAGE_NAME := atlassianlabs/$(BINARY_NAME)
-ARCH ?= $$(uname -s | tr A-Z a-z)
+ARCH ?= darwin
 METALINTER_CONCURRENCY ?= 4
 GOVERSION := 1.8
 GP := /gopath
@@ -132,7 +132,7 @@ release-race: docker-race
 release: release-normal release-race
 
 run: build
-	./build/bin/$(ARCH)/$(BINARY_NAME) --backends=stdout --verbose --flush-interval=2s
+	./build/bin/$(ARCH)/$(BINARY_NAME) --backends=stdout --verbose --flush-interval=1s
 
 run-docker: docker
 	cd build/ && docker-compose rm -f gostatsd
--- a/pkg/statsd/dispatcher.go
+++ b/pkg/statsd/dispatcher.go
@@ -2,10 +2,12 @@ package statsd
 
 import (
 	"context"
+	"fmt"
 	"hash/adler32"
 	"sync"
 
 	"github.com/atlassian/gostatsd"
+	"github.com/atlassian/gostatsd/pkg/statser"
 )
 
 // AggregatorFactory creates Aggregator objects.
@@ -26,6 +28,7 @@ func (f AggregatorFactoryFunc) Create() Aggregator {
 type MetricDispatcher struct {
 	numWorkers int
 	workers    map[uint16]worker
+	runContext context.Context
 }
 
 // NewMetricDispatcher creates a new NewMetricDispatcher with provided configuration.
@@ -50,22 +53,36 @@ func NewMetricDispatcher(numWorkers int, perWorkerBufferSize int, af AggregatorF
 
 // Run runs the MetricDispatcher.
 func (d *MetricDispatcher) Run(ctx context.Context, done gostatsd.Done) {
-	defer done()
-	var wg sync.WaitGroup
-	wg.Add(d.numWorkers)
-	for _, worker := range d.workers {
-		w := worker // Make a copy of the loop variable! https://github.com/golang/go/wiki/CommonMistakes
-		go w.work(wg.Done)
-	}
-	defer func() {
-		for _, worker := range d.workers {
-			close(worker.metricsQueue) // Close channel to terminate worker
-		}
-		wg.Wait() // Wait for all workers to finish
-	}()
+	d.runContext = ctx
 
-	// Work until asked to stop
-	<-ctx.Done()
+	go func() {
+		defer done()
+		var wg sync.WaitGroup
+		wg.Add(d.numWorkers)
+		for _, worker := range d.workers {
+			w := worker // Make a copy of the loop variable! https://github.com/golang/go/wiki/CommonMistakes
+			go w.work(wg.Done)
+		}
+		defer func() {
+			for _, worker := range d.workers {
+				close(worker.metricsQueue) // Close channel to terminate worker
+			}
+			wg.Wait() // Wait for all workers to finish
+		}()
+
+		// Work until asked to stop
+		<-ctx.Done()
+	}()
+}
+
+func (d *MetricDispatcher) runMetrics(ctx context.Context, statser statser.Statser) {
+	for _, worker := range d.workers {
+		worker.runMetrics(ctx, statser)
+	}
+	d.Process(ctx, func(aggrId uint16, aggr Aggregator) {
+		tag := fmt.Sprintf("aggregator_id:%d", aggrId)
+		aggr.TrackMetrics(statser.WithTags(gostatsd.Tags{tag}))
+	})
 }
 
 // DispatchMetric dispatches metric to a corresponding Aggregator.
@@ -73,6 +90,8 @@ func (d *MetricDispatcher) DispatchMetric(ctx context.Context, m *gostatsd.Metri
 	hash := adler32.Checksum([]byte(m.Name))
 	w := d.workers[uint16(hash%uint32(d.numWorkers))]
 	select {
+	case <-d.runContext.Done():
+		return d.runContext.Err()
 	case <-ctx.Done():
 		return ctx.Err()
 	case w.metricsQueue <- m:
@@ -94,6 +113,9 @@ func (d *MetricDispatcher) Process(ctx context.Context, f DispatcherProcessFunc)
 loop:
 	for _, worker := range d.workers {
 		select {
+		case <-d.runContext.Done():
+			wg.Add(cmdSent - d.numWorkers) // Not all commands have been sent, should decrement the WG counter.
+			break loop
 		case <-ctx.Done():
 			wg.Add(cmdSent - d.numWorkers) // Not all commands have been sent, should decrement the WG counter.
 			break loop
--- a/pkg/statsd/dispatcher.go
+++ b/pkg/statsd/dispatcher.go
@@ -51,8 +51,8 @@ func NewMetricDispatcher(numWorkers int, perWorkerBufferSize int, af AggregatorF
 	}
 }
 
-// Run runs the MetricDispatcher.
-func (d *MetricDispatcher) Run(ctx context.Context, done gostatsd.Done) {
+// RunAsync runs the MetricDispatcher in a goroutine.
+func (d *MetricDispatcher) RunAsync(ctx context.Context, done gostatsd.Done) {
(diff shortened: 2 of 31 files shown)


Makefile
```diff
@@ -132,7 +132,7 @@ release-race: docker-race
 release: release-normal release-race
 
 run: build
-	./build/bin/$(ARCH)/$(BINARY_NAME) --backends=stdout --verbose --flush-interval=1s
+	./build/bin/$(ARCH)/$(BINARY_NAME) --backends=stdout --verbose --flush-interval=2s
```
tiedotguy: This is here so people potentially see m vs m.count

pkg/statsd/dispatcher_worker.go
```diff
@@ -38,3 +41,15 @@ func (w *worker) executeProcess(cmd *processCommand) {
 	defer cmd.done() // Done with the process command
 	cmd.f(w.id, w.aggr)
 }
+
+func (w *worker) runMetrics(ctx context.Context, s statser.Statser) {
+	csw := statser.NewChannelStatsWatcher(
+		s,
+		"dispatch_aggregator",
```
tiedotguy: Intent of the naming is "sender_receiver".  We'll see how that pans out if I add more channels.

pkg/statsd/statsd_test.go
```diff
@@ -82,7 +83,20 @@ func (cb *countingBackend) Name() string {
 }
 
 func (cb *countingBackend) SendMetricsAsync(ctx context.Context, m *gostatsd.MetricMap, callback gostatsd.SendCallback) {
-	atomic.AddUint64(&cb.metrics, uint64(m.NumStats))
+	count := 0
```
tiedotguy: This does not count the same value as the original.  This counts the number of tagsets, the original counted the number of datapoints.  As such, the TestStatsdThroughput summary looks worse.

pkg/statser/channel_stats_watcher.go
```diff
@@ -0,0 +1,63 @@
...
+		capacity: capacity,
+		lenFunc:  lenFunc,
+		interval: interval,
+	}
+}
+
+// Run will run a ChannelStatsWatcher in the background until the supplied context is
+// closed, or Stop is called.
+func (csw *ChannelStatsWatcher) Run(ctx context.Context) {
+	go func() {
```
ash2k: Let the caller decide whether they want it in a goroutine - this is much more flexible. Caller cannot undo `go` to make it run synchronously but can always add `go` to make it asynchronous. I think this is idiomatic in Go to make apis synchronous unless there are really good reasons not to.

pkg/statser/channel_stats_watcher.go
```diff
@@ -0,0 +1,63 @@
...
+		ticker := time.NewTicker(csw.interval)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-ctx.Done():
+				ticker.Stop()
```
ash2k: This is not needed because of `defer`.

pkg/statser/statser.go
```diff
@@ -0,0 +1,31 @@
...
+	TimingMS(name string, ms float64, tags gostatsd.Tags)
+	TimingDuration(name string, d time.Duration, tags gostatsd.Tags)
+	NewTimer(name string, tags gostatsd.Tags) *Timer
+	WithTags(tags gostatsd.Tags) Statser
+}
+
+func concatTags(a, b gostatsd.Tags) gostatsd.Tags {
+	t := make(gostatsd.Tags, 0, len(a)+len(b))
+	t = append(t, a...)
+	t = append(t, b...)
```
ash2k: Hm, is it possible/better to use `copy` here since you pre-allocate the slice anyway?

pkg/statser/statser.go
```diff
@@ -0,0 +1,31 @@
...
+	return t
+}
+
+func copyTags(tags gostatsd.Tags) gostatsd.Tags {
+	new := make(gostatsd.Tags, len(tags))
```
ash2k: `new` is a builtin function as you know :) Better not to declare vars with reserved names. Doesn't the linter complain?

pkg/statser/statser_internal.go
```diff
@@ -0,0 +1,100 @@
...
+	"context"
+	"time"
+
+	"github.com/atlassian/gostatsd"
+)
+
+// InternalHandler is an interface to dispatch metrics to.  Exists
+// to break circular dependencies.
+type InternalHandler interface {
+	DispatchMetric(ctx context.Context, m *gostatsd.Metric) error
```
ash2k: I usually omit formal argument names in interface declarations when typed are more than enough to understand what the arguments are.

pkg/statser/statser_internal.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// InternalStatser is a Statser which sends metrics to a handler
+type InternalStatser struct {
+	ctx       context.Context
```
ash2k: The convention is to not store context in fields but pass it as an argument i.e. it is a "per-request thing". Or when invoking a long running function. But sometimes it is inevitable and it feels like this is one of those cases.

pkg/statser/statser_internal.go
```diff
@@ -0,0 +1,100 @@
...
+	return NewInternalStatser(is.ctx, concatTags(is.tags, tags), is.namespace, is.hostname, is.handler)
+}
+
+func (is *InternalStatser) dispatchMetric(metric *gostatsd.Metric) {
+	// the metric is owned by this file, we can change it freely because we know its origins
+	if is.namespace != "" {
+		metric.Name = is.namespace + "." + metric.Name
+	}
+	metric.Tags = append(metric.Tags, is.tags...)
+	is.handler.DispatchMetric(is.ctx, metric)
```
ash2k: It returns an error but we ignore it - linter does not complain? I understand that it is the desired behaviour, but maybe explicitly ignore the error to let the reader know that this is not a mistake but on purpose?

pkg/statsd/dispatcher.go
```diff
@@ -48,30 +52,61 @@ func NewMetricDispatcher(numWorkers int, perWorkerBufferSize int, af AggregatorF
...
 	hash := adler32.Checksum([]byte(m.Name))
 	w := d.workers[uint16(hash%uint32(d.numWorkers))]
+
+	// Protects both the read of d.running, and that the channel remains open
+	d.lock.RLock()
+	defer d.lock.RUnlock()
+	if !d.running {
+		return nil
+	}
+
```
tiedotguy: It does, which I'm definitely not thrilled about.  RWMutex was the best I could come up with.  If there's a better way, I'm open to it.  The problem I had is the CSW tries to submit to a closing channel, which is a data race.

I originally had an explicit stop on the CSWs, but then I need to make sure I have a backdoory path to always notify CSWs to stop in future.  I'd rather they continue to be able to "submit" but have it drop.

pkg/statser/channel_stats_watcher.go
tiedotguy: Derp.  Good spot.

pkg/statser/statser.go
tiedotguy: Possibly.  Can I do ```copy(t[len(a):], b)``` or something?  Still getting across the exact stylings of go.

pkg/statser/statser.go
tiedotguy: I didn't know, shall change.  And nope, it didn't care.

pkg/statser/statser_internal.go
tiedotguy: I still think there's a better way of managing it all (the ctx and the handling of the channels being closed).  I couldn't unravel it all in my head though.

pkg/statser/statser_internal.go
tiedotguy: I like to copy and paste :)

pkg/statser/channel_stats_watcher.go
tiedotguy: Done

pkg/statsd/dispatcher.go
```diff
@@ -48,30 +52,61 @@ func NewMetricDispatcher(numWorkers int, perWorkerBufferSize int, af AggregatorF
 	}
 }
 
-// Run runs the MetricDispatcher.
-func (d *MetricDispatcher) Run(ctx context.Context, done gostatsd.Done) {
```
tiedotguy: If it's launched as a go routine, then we start pumping metrics in to it before it's got running=true.  If we move the flag to NewMetricDispatcher, then we could be accepting metrics on something we haven't started, which seems like accepting metrics on something we've stopped.

tiedotguy:
Let's have a chat about it tomorrow, hopefully you can see a better approach for addressing the close/send race.

pkg/statsd/dispatcher.go
ash2k: > The problem I had is the CSW tries to submit to a closing channel, which is a data race.

Sounds like an issue that can/should be solved by controlling the shutdown order

tiedotguy: Think I got it sorted now, see how this looks.

tiedotguy:
tl;dr: Flusher runs.  Aggregator channels fill up.  Flusher attempts to dispatch internal metrics.  Blocks because channel full.  Deadlock because the aggregators are waiting for the flush to complete.

ash2k:
That is one reason why flusher pulls in meta metrics when it does a flush instead of metrics being pushed via standard codepath. Hm...

tiedotguy:
Easy enough to deal with, I'll put a buffer in InternalStatser for now.  Later I'll add dropping metrics (and counting them) when the channel is full.

I would prefer to read and drop metrics, than block reading and drop packets.  The impact can be more accurately quantified that way.

ash2k:
I think it is better to not read from the socket if we know we cannot handle it anyway. Why spend resources reading data when you can spend them processing what's been read? That was the original behaviour (it also logged them) and I changed it. We already monitor dropped packets on the interface today?

pkg/statser/channel_stats_watcher.go
```diff
@@ -0,0 +1,60 @@
...
+	tags     gostatsd.Tags
+	capacity int
+	lenFunc  func() int
+	interval time.Duration
+}
+
+// NewChannelStatsWatcher creates a new ChannelStatsWatcher
+func NewChannelStatsWatcher(client Statser, channelName string, tags gostatsd.Tags, capacity int, lenFunc func() int, interval time.Duration) *ChannelStatsWatcher {
+	t := gostatsd.Tags{"channel:" + channelName}
+	if tags != nil {
```
ash2k: This check is not necessary

pkg/statser/statser.go
```diff
@@ -0,0 +1,31 @@
...
+}
+
+func concatTags(a, b gostatsd.Tags) gostatsd.Tags {
+	t := make(gostatsd.Tags, 0, len(a)+len(b))
+	t = append(t, a...)
+	t = append(t, b...)
+	return t
+}
+
+func copyTags(tags gostatsd.Tags) gostatsd.Tags {
```
ash2k: Make it a method on `Tags`?

pkg/statser/statser.go
```diff
@@ -0,0 +1,31 @@
...
+	Gauge(name string, value float64, tags gostatsd.Tags)
+	Count(name string, amount float64, tags gostatsd.Tags)
+	Increment(name string, tags gostatsd.Tags)
+	TimingMS(name string, ms float64, tags gostatsd.Tags)
+	TimingDuration(name string, d time.Duration, tags gostatsd.Tags)
+	NewTimer(name string, tags gostatsd.Tags) *Timer
+	WithTags(tags gostatsd.Tags) Statser
+}
+
+func concatTags(a, b gostatsd.Tags) gostatsd.Tags {
```
ash2k: Make it a method on `Tags`?

pkg/statser/statser.go
```diff
@@ -0,0 +1,31 @@
...
+
+	"github.com/atlassian/gostatsd"
+)
+
+// Statser is the interface for sending metrics
+type Statser interface {
+	Gauge(name string, value float64, tags gostatsd.Tags)
+	Count(name string, amount float64, tags gostatsd.Tags)
+	Increment(name string, tags gostatsd.Tags)
+	TimingMS(name string, ms float64, tags gostatsd.Tags)
```
ash2k: Can we only keep `TimingDuration()`?

pkg/statser/statser_internal.go
```diff
@@ -0,0 +1,151 @@
...
+		ctx:       ctx,
+		wg:        wg,
+		buffer:    make(chan *gostatsd.Metric, bufferSize),
+		tags:      tags,
+		namespace: namespace,
+		hostname:  hostname,
+		handler:   handler,
+	}
+	wg.Add(1)
+	go is.run()
```
ash2k: Can we please make it the job of the caller to start (and then join) the goroutine? Then things get simpler - no need to pass `sync.WaitGroup`.

pkg/statser/channel_stats_watcher.go
tiedotguy: Wrapped up in Tags adjustments from below comments.

pkg/statser/statser.go
tiedotguy: Yep.

pkg/statser/statser.go
tiedotguy: Yep.

pkg/statser/statser_internal.go
tiedotguy: Good catch - had that for when I was creating a new InternalStatser, before the TaggingStatser.

tiedotguy: Entire WG can goes away now, we don't need shutdown because the buffer can drop.

ash2k: I'll take care of ordered shutdown in a separate PR later. Not just this place, but other cleanup is necessary (was necessary before your changes).

PR #121, opened 2017-08-19
Use stager for ordered shutdown

ash2k:


pkg/statsd/dispatcher.go
```diff
@@ -51,28 +52,30 @@ func NewMetricDispatcher(numWorkers int, perWorkerBufferSize int, af AggregatorF
...
+	for _, worker := range d.workers {
+		wg.Start(worker.work)
+	}
 
 	// Work until asked to stop
 	<-ctx.Done()
 }
 
-func (d *MetricDispatcher) runMetrics(ctx context.Context, statser statser.Statser) {
+func (d *MetricDispatcher) RunMetrics(ctx context.Context, statser statser.Statser) {
```
tiedotguy: Is there a reason I'm missing for making it public?

pkg/statser/statser_internal.go
```diff
@@ -34,7 +34,7 @@ type InternalStatser struct {
 
 // NewInternalStatser creates a new Statser which sends metrics to the
 // supplied InternalHandler.
-func NewInternalStatser(bufferSize int, tags gostatsd.Tags, namespace, hostname string, handler InternalHandler) Statser {
+func NewInternalStatser(bufferSize int, tags gostatsd.Tags, namespace, hostname string, handler InternalHandler) *InternalStatser {
```
tiedotguy: I'm not liking the change on the return type, specifically as it will be made configurable later on.

pkg/statsd/dispatcher.go
ash2k: If I want to reuse this type in another project, how do I enable metrics collection?

pkg/statser/statser_internal.go
ash2k: I removed the `Run` method from the `Statser` interface (because it is an implementation detail that you need to run this statser) and hence I need to export the type that exports the `Run` method. This is the same as `Dispatcher` interface and `MetricDispatcher` impl.

pkg/statsd/dispatcher.go
tiedotguy: I would say this is an application that exists as a cohesive whole, not a library to pick and choose bits from to use elsewhere.

pkg/statser/statser_internal.go
ash2k: We can always make this return an interface that augments `Statser` and adds a `Run` method.

pkg/statsd/dispatcher.go
ash2k: Why not?

Well, that is how I think software of high quality should be written in 99% of cases - everything, every module is a library. Only exception is `main` package (also "special" are things like OS kernel, drivers, firmware, etc). Every module should provide and API surface and make 0 assumptions about how it is used.

Let me explain why I think this is the right way. Lets take a look at the opposite to see why it is bad :)

1. git is a good example of the opposite approach. Every subcommand is a program (that is fine!) and that program is written without any internal boundaries/abstrations/modules AFAIK from some blopost, cannot find it. E.g. if there is an error it just terminates the whole process. Cannot use such code as library. Found this https://githubengineering.com/move-fast/

> With these limitations in mind, five years ago we began the development of libgit2, a re-implementation of most of Git’s low level APIs as a library that could be linked and used directly within a server-side process.

So because original authors wrote it as a program, someone had to invest enormous effort to rewrite it as a library.

2. Kuberetes has a few places where they use a "registry" pattern. A global structure where e.g. implementations of something register themselves. They are fixing this because initial design makes it hard to re-use the code in new ways that they want to use it. Also third parties are having a hard time to re-use the code!

3. Gostatsd actually used to do this "registry" pattern too. I changed it to not do it. Now every backend/cloud provider can be used as a library somewhere else and it will not pull in the registry itself into the binary. See https://github.com/atlassian/gostatsd/pull/4/commits/24f4ef4abf0bedb3254a2b7151730b65aab6d1ae

Less assumptions/knowledge about global context lead to cleaner designs and smaller dependency graphs. And in the end the software is much more reusable, testable and easier to maintain and understand. And has fewer bugs.

The only downside is that sometimes this requires a bit more thinking and coding. But in the long run I've never seen it to be a bad thing or cause any issues.

PR #127, opened 2017-09-01
Implement deflate compression for metrics uploads in Datadog backend

aelse:
The python dd-agent uses deflate compression when POSTing to /api/v1/series. Doing this results in a substantial reduction in payload size.

### DIFF
--- a/pkg/backends/datadog/datadog.go
+++ b/pkg/backends/datadog/datadog.go
@@ -2,6 +2,7 @@ package datadog
 
 import (
 	"bytes"
+	"compress/zlib"
 	"context"
 	"crypto/tls"
 	"encoding/json"
@@ -16,8 +17,8 @@ import (
 
 	"github.com/atlassian/gostatsd"
 
-	log "github.com/sirupsen/logrus"
 	"github.com/cenkalti/backoff"
+	log "github.com/sirupsen/logrus"
 	"github.com/spf13/viper"
 )
 
@@ -43,6 +44,7 @@ type Client struct {
 	client                http.Client
 	metricsPerBatch       uint
 	now                   func() time.Time // Returns current time. Useful for testing.
+	compressPayload       bool
 }
 
 // event represents an event data structure for Datadog.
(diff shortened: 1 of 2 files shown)


pkg/backends/datadog/datadog.go
```diff
@@ -191,17 +201,44 @@ func (d *Client) post(ctx context.Context, path, typeOfPost string, data interfa
...
+		}
+		var reader io.Reader
+		if compressPayload {
+			// Use deflate (zlib, since DD requires the zlib headers)
+			var buf bytes.Buffer
+			compressor, err := zlib.NewWriterLevel(&buf, zlib.BestCompression)
+			if err != nil {
+				return fmt.Errorf("unable to create zlib writer: %v", err)
+			}
+			compressor.Write(body)
```
ash2k: This does not return an error?

ash2k: Also, this will be compressing the payload on each retry to send - see how the returned function is used. Compression code should probably be mored before the `return func()`.

ash2k:
Do they support gzip? I think it is preferred to deflate because deflate is/was poorly specified and different browsers do it differently IIRC. Also, why not use it for events too? I can hardly imagine that their web server is configured to accept compressed payload only for a specific url path. Or is it?

pkg/backends/datadog/datadog.go
aelse: I considered that failures are uncommon but yes better to do it once. I've refactored.

aelse:
I used deflate because this is in use by the python agent. Given that this is a non-public undocumented API we're using it's best to stick fairly closely to the actual agent's behaviour.

I spent a lot of time debugging use of deflate because I was getting failures I couldn't explain. I believe deflate is supported on the metrics endpoint but not the events endpoint. You get a message that the payload is invalid, so i think it doesn't respect the content-encoding header.

```
WARN[0002] [datadog] failed to send events, sleeping for 1.433515344s: received bad status code 400
INFO[0004] [datadog] failed request status: 400
{"errors": ["Invalid JSON structure"]}
```

PR #129, opened 2017-09-06
Allows for dual stack mode to be enabled in the Datadog backend

tiedotguy:
Datadog recently started publishing AAAA records.  Environments without IPv6 can have a bad time due to network unreachable errors.

Also reworded the compression changelog to be more explicit.


ash2k:
I squash-merged it. Something funny was going on with all those commits.

tiedotguy:
Yeah, my git-fu is pretty lacking.  I should get off my fork.

PR #131, opened 2017-09-12
Add flags for explicit ipv4/ipv6 control

tiedotguy:
Happy eyeballs might be good for browsers, but at this volume it is better to be explicit than to guess.

### DIFF
--- a/pkg/backends/datadog/datadog.go
+++ b/pkg/backends/datadog/datadog.go
@@ -258,22 +258,23 @@ func NewClientFromViper(v *viper.Viper) (gostatsd.Backend, error) {
 	dd.SetDefault("api_endpoint", apiURL)
 	dd.SetDefault("metrics_per_batch", defaultMetricsPerBatch)
 	dd.SetDefault("compress_payload", true)
-	dd.SetDefault("dual_stack", false)
+	dd.SetDefault("network", "tcp")
 	dd.SetDefault("client_timeout", defaultClientTimeout)
 	dd.SetDefault("max_request_elapsed_time", defaultMaxRequestElapsedTime)
+
 	return NewClient(
 		dd.GetString("api_endpoint"),
 		dd.GetString("api_key"),
+		dd.GetString("network"),
 		uint(dd.GetInt("metrics_per_batch")),
 		dd.GetBool("compress_payload"),
-		dd.GetBool("dual_stack"),
 		dd.GetDuration("client_timeout"),
 		dd.GetDuration("max_request_elapsed_time"),
 	)
 }
 
 // NewClient returns a new Datadog API client.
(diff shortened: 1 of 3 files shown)


pkg/backends/datadog/datadog.go
```diff
@@ -258,28 +258,44 @@ func NewClientFromViper(v *viper.Viper) (gostatsd.Backend, error) {
...
+	ipv4 := dd.GetBool("ipv4")
+	ipv6 := dd.GetBool("ipv6")
+	var network string
+	if ipv4 && ipv6 {
+		network = "tcp"
+	} else if ipv4 {
+		network = "tcp4"
+	} else if ipv6 {
+		network = "tcp6"
+	}
```
ash2k: Should probably return an error if both are false?

pkg/backends/datadog/datadog.go
```diff
@@ -258,28 +258,44 @@ func NewClientFromViper(v *viper.Viper) (gostatsd.Backend, error) {
...
-func NewClient(apiEndpoint, apiKey string, metricsPerBatch uint, compressPayload, dualStack bool, clientTimeout, maxRequestElapsedTime time.Duration) (*Client, error) {
+func NewClient(apiEndpoint, apiKey, network string, metricsPerBatch uint, compressPayload bool, clientTimeout, maxRequestElapsedTime time.Duration) (*Client, error) {
 	if apiEndpoint == "" {
 		return nil, fmt.Errorf("[%s] apiEndpoint is required", BackendName)
 	}
 	if apiKey == "" {
 		return nil, fmt.Errorf("[%s] apiKey is required", BackendName)
 	}
+	if network == "" {
+		return nil, fmt.Errorf("[%s] at least one of ipv4 or ipv6 must be enabled", BackendName)
```
ash2k: Huh, this is the error that should be returned from `NewClientFromViper()`.
And this function does not know anything about ipv4/ipv6 flags. It should error out with "network must not be empty". WDYT?

ash2k: Not for this PR: we should remove backend name from error string and instead add it up the call stack.

pkg/backends/datadog/datadog.go
tiedotguy: I had that initially, then moved the validation logic to be with the rest of its friends below.

pkg/backends/datadog/datadog.go
tiedotguy: I think enforcing 4/6 at all is not our problem, and we can pass it off to anything Go supports :)

tiedotguy: Regarding adding it up the callstack: it should probably be on the logger...  The one we add when we do #2

PR #132, opened 2017-09-14
Handle zero-value expiry interval

jsvensson:
The `ParamExpiryInterval` flag says that setting it to zero should disable the expiry interval for metrics, immediately expiring them. However, setting it to zero effectively makes it an infinite interval, sending a zero-valued metric on every flush interval indefinitely.

Current logic says an expiryInterval of zero is never expired. Added quick check to handle zero, which sends the metric once without repeating zero-value sends.

### DIFF
--- a/pkg/statsd/aggregator.go
+++ b/pkg/statsd/aggregator.go
@@ -175,7 +175,7 @@ func (a *MetricAggregator) isExpired(now, ts gostatsd.Nanotime) bool {
 		return true
 	}
 
-	return a.expiryInterval != 0 && time.Duration(now-ts) > a.expiryInterval
+	return time.Duration(now-ts) > a.expiryInterval
 }
 
 func deleteMetric(key, tagsKey string, metrics gostatsd.AggregatedMetrics) {
--- a/pkg/statsd/aggregator.go
+++ b/pkg/statsd/aggregator.go
@@ -171,6 +171,10 @@ func (a *MetricAggregator) Process(f ProcessFunc) {
 }
 
 func (a *MetricAggregator) isExpired(now, ts gostatsd.Nanotime) bool {
+	if a.expiryInterval == 0 {
+		return true
+	}
+
 	return a.expiryInterval != 0 && time.Duration(now-ts) > a.expiryInterval
 }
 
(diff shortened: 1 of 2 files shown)


jsvensson:
One failed test, but I'm not sure of the intent of the test.

``` go
ma := &MetricAggregator{expiryInterval: 0}
assrt.Equal(false, ma.isExpired(now, now))
```

If zero disables the expiry interval, the assert should expect true.

tiedotguy:
My opinion on this is that it should be possible to have both infinite and immediate.  As for what that should look like, I'm not certain.

pkg/statsd/aggregator.go
```diff
@@ -171,6 +171,10 @@ func (a *MetricAggregator) Process(f ProcessFunc) {
 }
 
 func (a *MetricAggregator) isExpired(now, ts gostatsd.Nanotime) bool {
+	if a.expiryInterval == 0 {
```
tiedotguy: This effectively pulls a portion of the below check out.  I believe that's a good thing because it's clearer, but it should also be removed from the below check as well.

jsvensson:
Cleaned up the logic and changed that test to expect true, making it pass again.

As for an infinite expiry interval, it could be signified as a negative value for this config flag, though that's not entirely obvious and might be better as a separate config flag.

ash2k:
@jsvensson hey, sorry for the delay. Have you signed the CLA?

jsvensson:
Signed it just now!

I was looking through related code the other day, and I'm not sure how this change would interact with the expiration of gauges -- it might do weird stuff to gauges, but I'm not familiar enough with the codebase to tell properly.

ash2k:
@tiedotguy I've checked the CLA - signed. Are you ok with merging this?

tiedotguy:
I had a bit of a play around with it, and found the following:

Original:
-1s - sends once
0s - sends forever
1s - sends twice

Change:
-1s - sends once
0s - sends once
1s - sends twice

I think the problem is the documentation is ambiguous on behavior.  When I read ```After how long do we expire metrics (0 to disable)```, I read it as ```0 disables expiry (therefore living forever)```, whereas @jsvensson, you read it as ```0 disables expiry interval checking (therefore always expire)```.

Because with the current logic, it is possible to get both behaviors, but not with the change, I would say the best approach is to clarify the documentation to be explicit.

What do you think,  @jsvensson ?

jsvensson:
Yeah, with that clarification I can do what I want without needing any changes, so I say we scrap this MR and clarify the documentation instead. Didn't actually occur to me to try a negative value.

tiedotguy:
Shall update documentation to clarify things (which is way overdue anyway)

PR #136, opened 2017-09-21
Add node tracking as start of clustering support

tiedotguy:
Start of adding clustering.  The cluster binary is a place where things can live for now.

No metrics emitted yet, because I need to understand what sending metrics from a cluster FE/BE actually looks like.

### DIFF
--- a/cmd/cluster/server.go
+++ b/cmd/cluster/server.go
@@ -1,0 +1,58 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/atlassian/gostatsd/pkg/cluster/nodes"
+	"github.com/spf13/pflag"
+)
+
+// Cluster is everything for running a single node in a cluster
+type Cluster struct {
+	RedisAddr      string
+	Namespace      string
+	Target         string
+	UpdateInterval time.Duration
+	ExpiryInterval time.Duration
+}
+
+// newCluster will create a new Cluster with default values.
+func newCluster() *Cluster {
+	local, _ := nodes.LocalAddress("1.1.1.1:1")
+
+	return &Cluster{
+		RedisAddr:      "127.0.0.1:6379",
+		Namespace:      "namespace",
+		Target:         local.String() + ":8125",
+		UpdateInterval: time.Second,
+		ExpiryInterval: 30 * time.Second,
+	}
+}
+
+// AddFlags adds flags for a specific Server to the specified FlagSet.
+func (c *Cluster) AddFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&c.RedisAddr, "redis-addr", c.RedisAddr, "Redis address")
+	fs.StringVar(&c.Namespace, "namespace", c.Namespace, "Namespace")
+	fs.StringVar(&c.Target, "target", c.Target, "Target port")
+	fs.DurationVar(&c.UpdateInterval, "update-interval", c.UpdateInterval, "Cluster update interval")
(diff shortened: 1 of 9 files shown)


cmd/cluster/server.go
```diff
@@ -0,0 +1,58 @@
...
+// Run runs the specified Cluster.
+func (c *Cluster) Run() error {
+	rnt := nodes.NewRedisNodeTracker(c.RedisAddr, c.Namespace, c.Target, c.UpdateInterval, c.ExpiryInterval)
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	go rnt.Run(ctx)
+
+	t := time.NewTicker(time.Second)
+
+	for range t.C {
```
ash2k: It's an infinite loop, isn't it?

cmd/cluster/server.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// Run runs the specified Cluster.
+func (c *Cluster) Run() error {
+	rnt := nodes.NewRedisNodeTracker(c.RedisAddr, c.Namespace, c.Target, c.UpdateInterval, c.ExpiryInterval)
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	go rnt.Run(ctx)
+
+	t := time.NewTicker(time.Second)
```
ash2k: `defer t.stop()`

cmd/cluster/server.go
```diff
@@ -0,0 +1,58 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/atlassian/gostatsd/pkg/cluster/nodes"
```
ash2k: Missing blank line

pkg/cluster/nodes/nodes_redis_test.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+func (ft *fakeTime) Now() time.Time {
+	t := ft.samples[ft.next]
+	ft.next = (ft.next + 1) % len(ft.samples)
+	fmt.Printf("%v\n", t)
+	return t
+}
+
+func TestLearnNode(t *testing.T) {
```
ash2k: All tests can be `t.Parallel()`

pkg/cluster/nodes/nodes_redis.go
```diff
@@ -0,0 +1,167 @@
...
+	"sort"
+	"sync"
+	"time"
+
+	"github.com/go-redis/redis"
+	"github.com/sirupsen/logrus"
+)
+
+var (
+	errNoNodes = errors.New("No nodes available")
```
ash2k: Error strings should start with lowercase

cmd/cluster/server.go
tiedotguy: It is.  doing a ```for{ select { } }``` with only one case makes the linter complain though.

pkg/cluster/nodes/nodes_redis_test.go
tiedotguy: I knew I forgot something!

cmd/cluster/server.go
ash2k: Why single case? Should also select on `ctx.Done()`.

tiedotguy: ctx is never closed.  The cluster binary is a test harness only for testing the redis behaviour. Things like graceful shutdown and whatnot aren't going here because it's a lower standard, they'll be handled properly in the main binary.

PR #137, opened 2017-09-29
MetricReceiver.Receive batch read optimisation

brentthew:
Update MetricReceiver.Receive to batch read packets. In experimentation we found this to significantly increase performance. 

This implementation does not allow fallback to single read, and forces listening on IPv6. This is because the BatchRead api is protocol specific. This is an issue for backwards compatibility. 

Options I can see: 
- Setup this functionality to be configurable. I'd say a flag to enable or disable batch read. Could also potentially allow a flag to enable IPv4 bind instead... but perhaps overkill?
- Major version bump 

What would be preferable?

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ Running the server
 defaults. You can use `make run` to run the server with just the `stdout` backend
 to display info on screen.
 You can also run through `docker` by running `make run-docker` which will use `docker-compose`
-to run `gostatsd` with a graphite backend and a grafana dashboard. 
+to run `gostatsd` with a graphite backend and a grafana dashboard.
 
 Configuring backends and cloud providers
 ----------------------------------------
@@ -106,6 +106,15 @@ Load balancing and scaling out
 ------------------------------
 It is possible to run multiple versions of `gostatsd` behind a load balancer by having them
 send their metrics to another `gostatsd` backend which will then send to the final backends.
+
+Memory allocation for read buffers
+----------------------------------
+By default `gostatsd` will batch read multiple packets to optimise read performance. The amount of memory allocated
+for these read buffers is determined by the config options:
+
+    max-readers * receive-batch-size * 64KB (max packet size)
+
+The metric `avg_packets_in_batch` can be used to tune this memory usage if necessary.
 
 Using the library
 -----------------
--- a/pkg/statsd/batched_reader.go
+++ b/pkg/statsd/batched_reader.go
@@ -1,0 +1,71 @@
+package statsd
+
+import (
+	"net"
+
+	"golang.org/x/net/ipv6"
+)
+
+type Message struct {
+	Buffers [][]byte
+	Addr    net.Addr
+	N       int
+}
+
+type BatchReader interface {
+	ReadBatch(ms []Message) (int, error)
+}
+
+type V6BatchReader struct {
+	conn *ipv6.PacketConn
+}
+
+type GenericBatchReader struct {
+	conn net.PacketConn
+}
+
+func NewBatchReader(conn net.PacketConn) BatchReader {
+	switch c := conn.(type) {
+	case *net.UDPConn:
+		return &V6BatchReader{
+			conn: ipv6.NewPacketConn(c),
+		}
+	default:
+		return &GenericBatchReader{
+			conn: conn,
+		}
+	}
+}
+
+func (br *V6BatchReader) ReadBatch(ms []Message) (int, error) {
+	ms6 := make([]ipv6.Message, len(ms))
+	for i, m := range ms {
+		ms6[i].Buffers = m.Buffers
+	}
+	count, err := br.conn.ReadBatch(ms6, 0)
+	if err != nil {
+		return 0, err
+	}
+
+	for i := 0; i < count; i++ {
+		ms[i].Addr = ms6[i].Addr
+		ms[i].N = ms6[i].N
+	}
+
+	return count, nil
+}
+
+func (gbr *GenericBatchReader) ReadBatch(ms []Message) (int, error) {
+	if len(ms) == 0 {
+		// This tends to happen in test code when the batch size is not set,
+		// if production code starts at all, then a panic won't be a hazard.
+		panic("attempt to read 0 packets")
+	}
+	nbytes, addr, err := gbr.conn.ReadFrom(ms[0].Buffers[0])
+	if err != nil {
+		return 0, err
+	}
+	ms[0].Addr = addr
+	ms[0].N = nbytes
+	return 1, nil
+}
--- a/pkg/statsd/receiver.go
+++ b/pkg/statsd/receiver.go
@@ -8,8 +8,6 @@ import (
 	"sync/atomic"
 	"time"
 
-	"golang.org/x/net/ipv6"
-
 	"github.com/atlassian/gostatsd"
 	"github.com/atlassian/gostatsd/pkg/statser"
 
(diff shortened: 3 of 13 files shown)


tiedotguy:
Needs a flag for buffer count, and documentation on the memory cost (receivers * buffers * 64k) since it's non-trivial now.  Not sure on the IPv6 thing - I think it's reasonable to assume a modern kernel which can handle it though.

tiedotguy:
I'd also throw in a metric on the peak buffers used in an interval, so memory usage can be tuned.

pkg/statsd/receiver.go
```diff
@@ -65,10 +71,15 @@ func (mr *MetricReceiver) RunMetrics(ctx context.Context) {
 // Receive accepts incoming datagrams on c, parses them and calls Handler.DispatchMetric() for each metric
 // and Handler.DispatchEvent() for each event.
 func (mr *MetricReceiver) Receive(ctx context.Context, c net.PacketConn) {
-	buf := make([]byte, packetSizeUDP)
+	conn := ipv6.NewPacketConn(c)
+	messages := make([]ipv6.Message, receiveBatchSize)
```
ash2k: You don't need to allocate all of that on each `Receive()` invocation. Make it a field of `MetricReceiver`.

ash2k:
Listening on IPv6 does not listen on IPv4 right? This is not going to work for most people.

adampetrovic:
@ash2k Even though it's explicitly using an ipv6 address, the kernel supports dual stack mode. Which means it binds to both ipv6 and ipv4 interfaces.

pkg/statsd/receiver.go
tiedotguy: Receive is invoked once per goroutine, the buffers belong to the local goroutine, not the MR.

ash2k: Ok, then we need to do something about it to avoid reallocation. That is my point :)

tiedotguy: What reallocation?  It's allocated once outside of the forever loop?

brentthew:
@tiedotguy Yeah will add a flag. Will look into buffer size metric too. Where should the note about memory usage go?

brentthew:
@ash2k As @adampetrovic notes it does listen on IPv4 as well. The concern I have here is 
- Potentially some envs may not support IPv6? I think this may be something we can assume is available though... 
- If you supply a IPv4 address as config then it will fail, as it needs to bind on IPv6. This could be true in some cases, and is why I had to change to default config. Depending on how people have configured their environment, this is a breaking change. 

Potential solutions. Either: 
- Add a flag to chose whether you bind on IPv4 or v6
- Attempt a IPv6 bind, and if that fails, fallback to IPv4

Thoughts?

tiedotguy:
Docs should go in README.md.  I'll probably overhaul that file at some point and fully document everything - don't forget METRICS.md for the stat too.

I'm feeling a bit less happy about ipv6 since you pointed out the config change.  It's fine to say it's a breaking change, but now people are being forced to look up how to express an ipv4 address in ipv6 notation, if they already have it explicitly bound ([::ffff:1.2.3.4]:8125] I think?)

If we keep the ipv6 only style, that notation should be noted in the readme and/or changelog to make it a little easier.

brentthew:
@tiedotguy Err actually I was wrong. There isn't a need for an IPv6 address in config here. I've reverted that change 

(I must have been testing this change alongside another PR I'm going to raise for socket reuse which uses a different bind method).

pkg/statsd/receiver.go
brentthew: Yeah this memory shouldn't shared across goroutines. These buffers are (potentially) used for every invocation of ReadBatch. Each Receive invocation needs it's own set of buffers.

ash2k: I misread the code, ignore this comment, sorry.

pkg/statsd/statsd_defaults_and_params.go
```diff
@@ -106,33 +106,10 @@ const (
 	ParamPercentThreshold = "percent-threshold"
 	// ParamHeartbeatInterval is the name of the parameter with the heartbeat interval
 	ParamHeartbeatInterval = "heartbeat-interval"
+	// ParamReceiveBatchSize is the name of the parameter with the number of packets to read in each receive batch
+	ParamReceiveBatchSize = "receive-batch-size"
 )
 
-// NewServer will create a new Server with the default configuration.
-func NewServer() *Server {
```
brentthew: This wasn't used - seemed like cruft so I removed it

pkg/statsd/receiver.go
```diff
@@ -54,21 +59,28 @@ func (mr *MetricReceiver) RunMetrics(ctx context.Context) {
 		case <-ctx.Done():
 			return
 		case <-ticker.C:
-			mr.statser.Count("packets_received", float64(atomic.SwapUint64(&mr.packetsReceived, 0)), nil)
+			packetsReceived := float64(atomic.SwapUint64(&mr.packetsReceived, 0))
+			mr.statser.Count("packets_received", packetsReceived, nil)
 			mr.statser.Count("metrics_received", float64(atomic.SwapUint64(&mr.metricsReceived, 0)), nil)
 			mr.statser.Count("events_received", float64(atomic.SwapUint64(&mr.eventsReceived, 0)), nil)
 			mr.statser.Count("bad_lines_seen", float64(atomic.SwapUint64(&mr.badLines, 0)), nil)
+			mr.statser.Gauge("avg_packets_in_batch", packetsReceived/float64(atomic.SwapUint64(&mr.batchesRead, 0)), nil)
```
brentthew: @tiedotguy I went with avg not max... because it was a lot easier to calculate with atomic operations and I think it'll give what we're after

pkg/statsd/batched_reader.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+type GenericBatchReader struct {
+	conn net.PacketConn
+}
+
+func NewBatchReader(conn net.PacketConn) BatchReader {
+	switch c := conn.(type) {
+	case *net.UDPConn:
+		return &V6BatchReader{
```
tiedotguy: ipv4/ipv6 switching could be put here easily enough

pkg/statsd/receiver.go
tiedotguy: ```INFO[0009] stats.gauge.statsd.avg_packets_in_batch.s.covfefe NaN 1506714688```

Needs a check for zero.

brentthew:
Alright nothing outstanding that I can see - can I get re-review please

pkg/statsd/batched_reader.go
```diff
@@ -0,0 +1,71 @@
...
+	for i := 0; i < count; i++ {
+		ms[i].Addr = ms6[i].Addr
+		ms[i].N = ms6[i].N
+	}
+
+	return count, nil
+}
+
+func (gbr *GenericBatchReader) ReadBatch(ms []Message) (int, error) {
+	if len(ms) == 0 {
```
ash2k: We don't need this check - it will panic on the next line anyway (`ms[0]`). Just leave a comment?

pkg/statsd/receiver.go
```diff
@@ -79,13 +96,19 @@ func (mr *MetricReceiver) Receive(ctx context.Context, c net.PacketConn) {
 			continue
 		}
 		// TODO consider updating counter for every N-th iteration to reduce contention
```
ash2k: If we are reading a lot of packets toady, this may actually make sense. In a separate PR.

README.md
```diff
@@ -107,6 +107,15 @@ Load balancing and scaling out
...
 send their metrics to another `gostatsd` backend which will then send to the final backends.
 
+Memory allocation for read buffers
+----------------------------------
+By default `gostatsd` will batch read multiple packets to optimise read performance. The amount of memory allocated
+for these read buffers is determined by the config options:
+
+    max-readers * receive-batch-size * 64KB (max packet size)
+
+The metric `avg_packets_in_batch` can be used to tune this memory usage if necessary.
```
ash2k: Maybe also mention `receive-batch-size` command line flag.

pkg/statsd/receiver.go
tiedotguy: This was one of the first things I tested, and it didn't seem to make a difference.  With the changes here it's even less important because we basically get the "every N-th" for free.

README.md
tiedotguy: I'll release today, and tweak when I tag (also update changelog)

PR #138, opened 2017-10-02
Refactor handler/dispatch pipeline

tiedotguy:
I found the whole Dispatcher / Handler confusing, especially in Flusher with both a Dispatcher and Handler.  This simplifies it down to a (Metric|Event)Handler which you Dispatch(Metric|Event)s to.

Unfortunately there was some renames, and the files don't display as well as I would like.  dispatcher.go and dispatching_handler.go were largely combined then split in to handler_backend.go and handler_tags.go

I opted for a MetricHandler and an EventHandler rather than a generic Handler to allow the pipeline to diverge. Even though it doesn't.


pkg/statsd/flusher.go
```diff
@@ -20,25 +20,23 @@ type MetricFlusher struct {
...
+	backendHandler *BackendHandler
+	backends       []gostatsd.Backend
+	selfIP         gostatsd.IP
+	hostname       string
+	statser        statser.Statser
 }
 
 // NewMetricFlusher creates a new MetricFlusher with provided configuration.
-func NewMetricFlusher(flushInterval time.Duration, dispatcher Dispatcher, handler Handler, backends []gostatsd.Backend, selfIP gostatsd.IP, hostname string, statser statser.Statser) *MetricFlusher {
+func NewMetricFlusher(flushInterval time.Duration, metricHandler *BackendHandler, backends []gostatsd.Backend, selfIP gostatsd.IP, hostname string, statser statser.Statser) *MetricFlusher {
```
ash2k: Should be `backendHandler`?

pkg/statsd/handler_backend.go
```diff
@@ -0,0 +1,164 @@
...
+type BackendHandler struct {
+	eventWg          sync.WaitGroup
+	backends         []gostatsd.Backend
+	concurrentEvents chan struct{}
+
+	numWorkers int
+	workers    map[uint16]*worker
+}
+
+// NewBackendHandler initialises a new Handler which sends metrics and events to all backends
```
ash2k: All godoc strings must end with a dot. Try running the linter :)

pkg/statser/statser_internal.go
```diff
@@ -8,12 +8,18 @@ import (
 	"github.com/atlassian/gostatsd"
 )
 
-// InternalHandler is an interface to dispatch metrics to.  Exists
+// InternalMetricHandler is an interface to dispatch metrics to.  Exists
 // to break circular dependencies.
-type InternalHandler interface {
+type InternalMetricHandler interface {
 	DispatchMetric(ctx context.Context, m *gostatsd.Metric) error
```
ash2k: May remove variable names - types are descriptive enough.

pkg/statsd/flusher.go
```diff
@@ -20,25 +20,23 @@ type MetricFlusher struct {
...
 
-	flushInterval time.Duration // How often to flush metrics to the sender
-	dispatcher    Dispatcher
-	handler       Handler
-	backends      []gostatsd.Backend
-	selfIP        gostatsd.IP
-	hostname      string
-	statser       statser.Statser
+	flushInterval  time.Duration // How often to flush metrics to the sender
+	backendHandler *BackendHandler
```
ash2k: Make this an interface?

tiedotguy: Flusher is intrinsically tied to aggregators and the backend.  When clustering on the frontend comes in, for example, it won't have a flusher (or if it does for internal metrics, it will still be tied to the BackendHandler only).

If this has a need for being handled differently in future, it can always be put in an interface then.

pkg/statser/statser_internal.go
tiedotguy: I'm a big fan of variable names - it makes it easier to copy and paste the interface definition.  Plus general verbosity is rarely bad.

pkg/statsd/flusher.go
ash2k: Sure. I was more thinking about unit tests.

tiedotguy: Ooh, gotcha.  Shall have a think on that one.

ash2k:
Why not rebase? :) Cluttering git history is not cool IMO.

tiedotguy:
I will, I just wanted to quickly get it up before the receiver/decoder split comes in.

pkg/statsd/statsd.go
```diff
@@ -138,30 +145,30 @@ func (s *Server) RunWithCustomSocket(ctx context.Context, sf SocketFactory) erro
...
-	// 7. Send events on start and on stop
+	// 9. Send events on start and on stop
 	// TODO: Push these in to statser
-	defer sendStopEvent(handler, ip, hostname)
-	sendStartEvent(ctx, handler, ip, hostname)
+	defer sendStopEvent(events, ip, hostname)
+	sendStartEvent(ctx, events, ip, hostname)
 
-	// 8. Listen until done
+	// 10. Listen until done
```
brentthew: leave it here, but I might do away with this numbering in my PR... don't really feel it adds much except diff noise

brentthew:
Sorry only just saw this, had my emails misconfigured. 

Looks good to me! I'm a fan of separating the events and metrics paths. 

(And wow github PRs are :meh:)

pkg/statsd/statsd.go
tiedotguy: The stager is really nice and lets us pull almost everything out to functions (previously there was a heap of defer calls).  That would be a good cleanup to address both the function length and numbering in one.

pkg/statsd/handler_backend.go
tiedotguy: Dag nabbit.  I thought I caught this one.  I've already force pushed enough for today, so I'm ignoring this one :P

tiedotguy:
Think this is ready for a final review.  Quick, before the SO_REUSEPORT discussion is resolved!

brentthew:
LGTM!

aelse:
I don't have bandwidth to properly look at this. Review from @ash2k and @brentthew should be fine.

tiedotguy:
I'll cut a release in a little while.

PR #141, opened 2017-10-06
Separate MetricReceiver into DatagramReceiver and DatagramParser

brentthew:
Current MetricReceiver:
1. Reads off datagrams 
2. Splits them into lines 
3. Lexes lines into metrics/events
4. Passes off to handler 

Steps 2-4 are cpu bound and will benefit from higher concurrency. Step 1 is not, and we've seen higher concurrency lead to lower socket read performance. 

I've split these into 

DatagramReceiver:
1. Reads datagram from socket
2. Pushes them to a chan

and 

DatagramParser: 
1. Read datagrams from chan
2. Splits them into lines 
3. Lexes lines into metrics/events
4. Passes off to handler 

Which can be scaled individually. I've set defaults of DatagramParser (MaxParsers) as num cpu, and DatagramReceiver (MaxReaders) to 8. Optimal value of MaxReaders may need some investigation. We found 8 to be performant in our testing.


pkg/statsd/receiver.go
```diff
@@ -96,99 +83,24 @@ func (mr *MetricReceiver) Receive(ctx context.Context, c net.PacketConn) {
...
 			addr := messages[i].Addr
 			nbytes := messages[i].N
 			buf := messages[i].Buffers[0]
-			if err := mr.handlePacket(ctx, addr, buf[:nbytes]); err != nil {
-				if err == context.Canceled || err == context.DeadlineExceeded {
-					return
-				}
-				log.Warnf("Failed to handle packet: %v", err)
+			msg := make([]byte, nbytes)
+			copy(msg, buf)
```
ash2k: This is a big drawback of this approach 😕

ash2k: To overcome this to some degree we can use [`sync.Pool`](https://golang.org/pkg/sync/#Pool) to pool free message buffers for reuse. I think we do it for something already in this project.

pkg/statsd/statsd.go
```diff
@@ -118,14 +119,32 @@ func (s *Server) RunWithCustomSocket(ctx context.Context, sf SocketFactory) erro
...
 	if s.HeartbeatInterval != 0 {
 		hb := stats.NewHeartBeater(statser, "heartbeat", s.HeartbeatTags, s.HeartbeatInterval)
 		stage = stgr.NextStage()
 		stage.StartWithContext(hb.Run)
 	}
 
-	// 5. Start the Receiver
+	// 5. Start the Parser
+	// Open receiver <-> parser chan
+	datagrams := make(chan *Datagram, s.MaxReaders*s.ReceiveBatchSize)
```
ash2k: Do we need to make this buffered?

pkg/statsd/receiver.go
```diff
@@ -96,99 +83,24 @@ func (mr *MetricReceiver) Receive(ctx context.Context, c net.PacketConn) {
...
+			dg := &Datagram{
+				IP:  getIP(addr),
+				Msg: msg,
 			}
+			dr.out <- dg
```
ash2k: For code below I see that the shutdown order between parser/receiver is not defined (they are started within same Stage). That means this can block forever if nobody is reading (and/or the channel is full). I think it should be a `select` with `<-ctx.Done()`.

pkg/statsd/statsd.go
```diff
@@ -118,14 +119,32 @@ func (s *Server) RunWithCustomSocket(ctx context.Context, sf SocketFactory) erro
...
+	dgStats := stats.NewChannelStatsWatcher(
+		statser,
+		"statsd.channel",
```
tiedotguy: This is the channel name, not the metric name.  It ends up as the channel tag against the <namespace.>channel.* metric.

pkg/statsd/statsd_defaults_and_params.go
```diff
@@ -59,6 +59,8 @@ const (
 	DefaultHeartbeatInterval = time.Duration(0)
 	// DefaultReceiveBatchSize is the number of packets to read in each receive batch
 	DefaultReceiveBatchSize = 50
+	// DefaultMaxReaders is the default number of socket reading goroutines.
+	DefaultMaxReaders = 8
```
tiedotguy: min(8, runtime.NumCPU())

pkg/statsd/receiver.go
tiedotguy: I originally handled this by passing the entire buffer in and having a return channel to pass it back when I was done.  It's a fixed pool of memory and no copies.

tiedotguy: Also if we go that route, I think we should pass the batch, not the datagram.

pkg/statsd/statsd.go
tiedotguy: Also METRICS.md, there's a section just for CSWs

pkg/statsd/statsd_defaults_and_params.go
ash2k: I wanted to suggest literally the same.

pkg/statsd/receiver.go
ash2k: - Using `sync.Pool` is non-blocking vs reading/writing from/to return channel.
- Re. passing the whole batch: I also thought about it. It seems work will be redistributed less fairly that way. However we save on channel interactions so throughput should improve in case of many batches. So +1.

brentthew: Right, because we're worried about the GC overhead? I'll look into buffer reuse.

pkg/statsd/statsd.go
brentthew: We saw some inconsistency in the rate that datagrams were being read - my idea here was to buffer that out. Also gives us insight into which component is backing up. Whats your concern with making it buffered, memory usage?

pkg/statsd/receiver.go
tiedotguy: I think it's more the copying overhead?  I'm vaguely curious about the GC overhead, I think it's up there, but I suspect the channel overhead is the big one - once this is in, I'm (or if you want to, go for it) going to try adding a small local buffer on the decoders, so multiple metrics are sent to the channel at a time.

ash2k: > Right, because we're worried about the GC overhead? I'll look into buffer reuse.

Both GC and copying are bad.

pkg/statsd/statsd.go
ash2k: Datagrams are buffered in network stack already, I don't see why additional buffering inside of the application is needed.
CPU puts this data into its cache (read from network stack), then discards it from the cache (while it sits idle in the buffer), then puts it back again (when data is read from the buffer). Less buffering is better because there is less chance for cache thrashing. Also it is easier to understand unbuffered channels and chance of having a bug hidden by buffering (a deadlock typically) is much smaller.

ash2k: Note that we already have buffering - channels in from of aggregators are buffered. Don't need an extra layer between receiver and parser. If parsers cannot keep up then there is no point in reading more from network.

ash2k: And network stack already provides buffers to handle spikes in traffic.

tiedotguy: The network buffer fills in about 2 seconds, which aggregation exceeds.  The aggregator buffers are unevenly distributed.  This gives us a place to both control and monitor the amount we buffer.

ash2k: Nothing stops us from bumping up the size of the network buffer? And it is probably more memory efficient. And it is evenly distributed. And doesn't have problems I described above. And can and should be monitored. So?

pkg/statsd/statsd.go
brentthew: right yeah sorry, I was looking at Alexander's implementation

pkg/statsd/receiver.go
brentthew: Yeah good catch. I'll make the change to check our context isn't cancelled. 

Would you also recommend splitting off into a new stage(s)? I'm curious about the triggers for creating a new stage... why doesn't the flusher run within it's own stage? Should it?

If I'm not mistaken is it just sensible to run each worker in it's own stage so that shutdown matches startup?

pkg/statsd/statsd.go
brentthew: I can certainly see your argument for letting the network stack handling buffering. I was going to argue that we're having issues exceeding the network buffers... but I think you've hit the nail on the head with 

>  If parsers cannot keep up then there is no point in reading more from network.

I'd prefer to try tweaking network buffers further before we start tweaking here, so I'll remove this buffering for now.

pkg/statsd/receiver.go
ash2k: Stages are necessary to ensure ordered shutdown. If order does not matter then shutdown within a stage will happen concurrently. My rule of thumb is to only create stages when order matters. This also serves as documentation - it is done that way because there is a reason for it.
Feel free to keep it as it is or change it back - I'd like to see if it is possible to get rid of context checking in some places and instead rely on the data channel being closed when producer will no longer send stuff. This is more performant that `select` and reduces boilerplate. But that is for later.

METRICS.md
```diff
@@ -32,6 +32,7 @@ channel tag, and may have additional tags specified below.
 | Channel name        | Additional tags | Description
 | ------------------- | --------------- | -----------
 | dispatch_aggregator | aggregator_id   | Channel to dispatch metrics to a specific aggregator.
+| sender_receiver     |                 | Channel of datagrams read but yet to be parsed.
```
tiedotguy: This is gone now

brentthew: ah yeah good call

brentthew:
Running this build showed a marginal increase in CPU usage. Will investigate this at a later time.

PR #143, opened 2017-10-10
RFC: Add support for creating a seperate connection per reader

brentthew:
Enables reuse of the same address, to allow a seperate connection per reader. 

This change resulted in the single largest performance increase we saw in testing... the issue here is the dependent library (https://github.com/jbenet/go-reuseport) we're pulling in is a) a bit of a hack and b) pulls in a bunch of other dependencies. I haven't vetted the license of all these. 

Alternatively we could also implement the reuse logic ourselves. Thoughts?


aelse:
We have to vet the licences. It might be possible to use sourceclear in our builds to help with this.

glide.lock
```diff
@@ -57,16 +61,29 @@ imports:
   - json/parser
   - json/scanner
   - json/token
+- name: github.com/ipfs/go-log
```
ash2k: Yet another logging library.

glide.lock
```diff
@@ -80,12 +97,14 @@ imports:
 - name: github.com/spf13/viper
   version: 25b30aa063fc18e48662b86996252eabdcf2f0c7
 - name: github.com/stretchr/testify
-  version: 890a5c3458b43e6104ff5da8dfa139d013d77544
+  version: 69483b4bd14f5845b5a1e55bca19e954e827f1d0
   subpackages:
   - assert
   - require
+- name: github.com/whyrusleeping/go-logging
```
ash2k: Yet another logging library number 2.

pkg/statsd/statsd.go
```diff
@@ -153,6 +166,18 @@ func (s *Server) RunWithCustomSocket(ctx context.Context, sf SocketFactory) erro
 	stage = stgr.NextStage()
 	stage.StartWithContext(receiver.RunMetrics)
 	for r := 0; r < s.MaxReaders; r++ {
+		// Open socket
+		c, err := sf()
```
ash2k: What happens to `c` opened on line 141/154? Should probably drop it.

brentthew: Oh, oops - merge issue. Definitely should be dropped

brentthew:
@ash2k 1->2 workers saw the most significant increase. We saw that around 8 workers was most performant, running on c4.8xl with increased network buffers. Increasing the number of connections above this saw performance declines. This is one of the reasons for my PR that separates reading from parsing.

Colby might have the concrete numbers - I'll get him to ping them through.

We did indeed see that network queues were pinned to CPUs. Throughout testing we saw that 2 cores were handling reading, observed by the significant number of soft interrupts for only those 2 cores. 

On ubuntu (14 and 16) we saw this have a more significant impact than on the AWS linux AMI. 

(We saw that the source of the packets also had a very significant impact here - we suspected as a result of partitioning of packets across the queues)

pkg/statsd/statsd_defaults_and_params.go
```diff
@@ -62,6 +62,8 @@ const (
 	DefaultHeartbeatInterval = time.Duration(0)
 	// DefaultReceiveBatchSize is the number of packets to read in each receive batch
 	DefaultReceiveBatchSize = 50
+	// DefaultConnPerReader is the default for whether to create a connection per reader
+	DefaultConnPerReader = false
```
brentthew: note I've left this disabled by default - mainly due to concerns about the portability of socket reuse settings

glide.lock
brentthew: Yeah these also had me doubting this library. Do we think it's worth implementing and maintaining the socket logic ourselves? I'd also saw if we do so we'd drop any kind of portability (while this option is disabled).

tiedotguy: I'm leaning towards just implementing it ourselves, or checking if there's any other library with a reduced footprint.  On the portability issue, abstraction layers are not hard to create.

brentthew: Yeah but I don't want to own the complexities of maintaining socket code for one platform, let alone multiple.

brentthew:
Closing this out in favour of implementing reuse logic myself

PR #145, opened 2017-10-12
Add support for creating a seperate connection per reader

brentthew:
Went back to including this library. There's a lot that goes into creating these sockets, and I don't want to duplicate this logic in gostatsd. 

Reraised from https://github.com/atlassian/gostatsd/pull/143

New dep licenses are all approved for use in Atlassian OS projects  

- ISC https://github.com/jbenet/go-reuseport/blob/master/LICENSE
- BSD https://github.com/jbenet/go-sockaddr/blob/master/LICENSE
- Apache https://github.com/gxed/GoEndian/blob/master/LICENSE
- MIT https://github.com/ipfs/go-log/blob/master/LICENSE
- BSD https://github.com/whyrusleeping/go-logging/blob/master/LICENSE
- This one isn't explicitly labeled, but is BSD https://github.com/gxed/eventfd/blob/master/LICENSE


pkg/statsd/statsd.go
```diff
@@ -46,14 +49,32 @@ type Server struct {
...
+	return s.RunWithCustomSocket(ctx, socketFactory(s.MetricsAddr, s.ConnPerReader))
 }
 
 // SocketFactory is an indirection layer over net.ListenPacket() to allow for different implementations.
 type SocketFactory func() (net.PacketConn, error)
 
+func socketFactory(metricsAddr string, connPerReader bool) SocketFactory {
+	if connPerReader {
+		// go-reuseport requires explicitly representing the unspecified address
+		addr, err := net.ResolveUDPAddr("udp", metricsAddr)
```
brentthew: It was either this, or a breaking change to flags to `metrics-addr` to require specifying the IP

pkg/statsd/statsd.go
```diff
@@ -145,22 +166,22 @@ func (s *Server) RunWithCustomSocket(ctx context.Context, sf SocketFactory) erro
...
 	stage.StartWithContext(receiver.RunMetrics)
 	for r := 0; r < s.MaxReaders; r++ {
+		// Open socket
+		c, err := sf()
+		if err != nil {
+			return err
+		}
+		defer func(c net.PacketConn) {
+			// This makes receivers error out and stop
+			if e := c.Close(); e != nil {
```
tiedotguy: Shouldn't be a problem, but in a non-shared case, this will close the same socket multiple times?

ash2k: The factory function could return a `func() error` that should be used to free up the resources. And the implementation could call `Close()` just one time for non-shared case.

brentthew: I overrode `Close()` and wrapped it in a sync.Once

tiedotguy: :meh: it doesn't panic.

PR #149, opened 2017-11-27
Add metrics in AWS CP and CP handler

tiedotguy:


tiedotguy:
Sigh.  Will fix metalinter.

PR #150, opened 2017-11-27
Add metrics in AWS CP and CP handler

tiedotguy:


cmd/cluster/server.go
```diff
@@ -21,7 +21,11 @@ type Cluster struct {
 
 // newCluster will create a new Cluster with default values.
 func newCluster() *Cluster {
-	local, _ := nodes.LocalAddress("1.1.1.1:1")
+	local, err := nodes.LocalAddress("1.1.1.1:1")
```
tiedotguy: for metalinter

pkg/statsd/handler_cloud.go
```diff
@@ -107,6 +127,45 @@ func (ch *CloudHandler) WaitForEvents() {
...
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			ch.emit(statser)
+		}
+	}
+}
+
+func (ch *CloudHandler) emit(statser stats.Statser) {
```
aelse: As discussed, all these atomic operations worry me a bit. If not hot paths now they could become so in the future with a bit of refactoring. We might be able to replace them with channels between producer and consumer (so that not every operation by the producer is atomic), or else add perf tests so we can see how this changes behaviour vs the current code and regression due to future changes. I'll have a deeper look at it...

pkg/statsd/handler_cloud.go
```diff
@@ -215,11 +297,15 @@ func (ch *CloudHandler) handleLookupResult(ctx context.Context, lr *lookupResult
...
 	if metrics != nil {
 		delete(ch.awaitingMetrics, lr.ip)
+		atomic.AddUint64(&ch.statsMetricItemsQueued, ^uint64(len(metrics)-1))
+		atomic.AddUint64(&ch.statsMetricHostsQueued, ^uint64(0))
 		go ch.updateAndDispatchMetrics(ctx, lr.instance, metrics...)
 	}
 	events := ch.awaitingEvents[lr.ip]
 	if events != nil {
 		delete(ch.awaitingEvents, lr.ip)
+		atomic.AddUint64(&ch.statsEventItemsQueued, ^uint64(len(metrics)-1))
```
tiedotguy: Hawkward.

pkg/statsd/handler_cloud.go
ash2k: It is always possible to test if it matters - comment them all out, rebuild and redeploy :) I would be very surprised if they makes any difference.

pkg/statsd/handler_cloud.go
```diff
@@ -166,15 +225,28 @@ func (ch *CloudHandler) doRefresh(ctx context.Context, t time.Time) {
...
+				positiveDeleted++
+			}
 		} else if t.After(holder.expires) {
 			// Entry needs a refresh.
 			ch.toLookupIPs = append(ch.toLookupIPs, ip)
 		}
 	}
+
+	atomic.AddUint64(&ch.statsCachePositive, ^(positiveDeleted - 1))
+	atomic.AddUint64(&ch.statsCacheNegative, ^(negativeDeleted - 1))
```
ash2k: How does this work?

tiedotguy: Why: https://golang.org/pkg/sync/atomic/#AddUint64
How: https://en.wikipedia.org/wiki/Two's_complement

I considered throwing a function the gostatsd package or something, but it seemed kinda meh.

cmd/cluster/server.go
```diff
@@ -21,7 +21,11 @@ type Cluster struct {
...
+
+	if err == nil {
```
brentthew: `!= nil`?

tiedotguy: Yeeaah.

pkg/statsd/handler_cloud.go
brentthew: I think I'd like this refactored out into another function - I doubt I'll remember this if I came across it in a few months

tiedotguy: But it's idiomatic!

Seriously though, I've removed most of the atomics in a refactor and we don't need to decrement anymore.

pkg/statsd/handler_cloud.go
tiedotguy: 3 of these are with it, 3 are without.  I forget which is which.
```
BenchmarkCloudHandlerDispatchMetric-4            2000000               628 ns/op             704 B/op         16 allocs/op
BenchmarkCloudHandlerDispatchMetric-4            2000000               568 ns/op             704 B/op         16 allocs/op
BenchmarkCloudHandlerDispatchMetric-4            2000000               554 ns/op             704 B/op         16 allocs/op
BenchmarkCloudHandlerDispatchMetric-4            3000000               544 ns/op             704 B/op         16 allocs/op
BenchmarkCloudHandlerDispatchMetric-4            3000000               525 ns/op             704 B/op         16 allocs/op
BenchmarkCloudHandlerDispatchMetric-4            2000000               688 ns/op             704 B/op         16 allocs/op
```

This is on my laptop, it may not be so clear when there's a billion cores hammering it, but I expect it to be fine because normal access takes a read lock on CloudHandler.rw which at its heart is a couple of atomic.AddInt32's.  We can probably reclaim those nanos by fixing line 421.

pkg/statsd/handler_cloud.go
```diff
@@ -314,7 +418,7 @@ func updateInplace(tags *gostatsd.Tags, hostname *string, instance *gostatsd.Ins
 		// Update hostname inplace
 		*hostname = instance.ID
 		// Update tag list inplace
-		*tags = append(*tags, "region:"+instance.Region)
+		*tags = append(*tags, "region:"+instance.Region) // TODO: memory allocation cleanup
```
tiedotguy: This is a fixed value and should be pushed in to instance.Tags.  Didn't want to clutter this PR.  Will try and do some serious memory profiling at some point in the future because I think it's pretty rough.

pkg/statsd/handler_cloud.go
aelse: Yes it's always possible to test but in practice if no one runs the tests (because they don't exist) we won't spot regressions. Thanks for adding the benchmarks @tiedotguy

pkg/statsd/handler_cloud.go
```diff
@@ -145,25 +152,32 @@ func (ch *CloudHandler) RunMetrics(ctx context.Context, statser stats.Statser) {
 		case <-ctx.Done():
 			return
 		case <-ticker.C:
-			ch.emit(statser)
+			select {
```
ash2k: Why not `select` on `<-ctx.Done()` and `ch.emitChan <- statser`? I.e. why blocking here is bad?

tiedotguy: This is (sort of) the model of the metrics-refactor branch, where a channel tells you when to emit and it drops it if you miss your shot.  For now this is just a quick'n'dirty to get things in.

Issue #118, opened 2017-08-16
BenchmarkReceive is not a valid benchmark


tiedotguy:
All it measures is how fast we can allocate and GC 64kB.  The function is normally long running.

ash2k:
I think I was trying to measure number and size of allocations per receive.

PR #152, opened 2017-12-12
Performance work

tiedotguy:
To keep it manageable, the commits are much better isolated and justified.  Recommended is to review them sequentially.

This also addresses #118

### DIFF
--- a/pkg/cloudproviders/aws/aws.go
+++ b/pkg/cloudproviders/aws/aws.go
@@ -107,7 +107,7 @@ func (p *Provider) Instance(ctx context.Context, IP ...gostatsd.IP) (map[gostats
 						gostatsd.NormalizeTagKey(aws.StringValue(tag.Key)),
 						aws.StringValue(tag.Value))
 				}
-				tags[len(instance.Tags)-1] = fmt.Sprintf("region:%s", region)
+				tags[len(instance.Tags)-1] = "region:" + region
 				instances[ip] = &gostatsd.Instance{
 					ID:   aws.StringValue(instance.InstanceId),
 					Tags: tags,
--- a/pkg/cloudproviders/aws/aws.go
+++ b/pkg/cloudproviders/aws/aws.go
@@ -101,16 +101,16 @@ func (p *Provider) Instance(ctx context.Context, IP ...gostatsd.IP) (map[gostats
 				if err != nil {
 					log.Errorf("Error getting instance region: %v", err)
 				}
-				tags := make(gostatsd.Tags, len(instance.Tags))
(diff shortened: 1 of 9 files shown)


pkg/cloudproviders/aws/aws.go
```diff
@@ -101,16 +101,16 @@ func (p *Provider) Instance(ctx context.Context, IP ...gostatsd.IP) (map[gostats
...
 					log.Errorf("Error getting instance region: %v", err)
 				}
-				tags := make(gostatsd.Tags, len(instance.Tags))
+				tags := make(gostatsd.Tags, len(instance.Tags)+1)
 				for idx, tag := range instance.Tags {
 					tags[idx] = fmt.Sprintf("%s:%s",
 						gostatsd.NormalizeTagKey(aws.StringValue(tag.Key)),
 						aws.StringValue(tag.Value))
 				}
+				tags[len(instance.Tags)-1] = fmt.Sprintf("region:%s", region)
```
ash2k: String concatenation should be cheaper than formatting?

pkg/statsd/receiver.go
```diff
@@ -121,3 +123,27 @@ func getIP(addr net.Addr) gostatsd.IP {
...
+type bufferPool struct {
+	p sync.Pool
+}
+
+func newBufferPool() *bufferPool {
+	return &bufferPool{
+		p: sync.Pool{
+			New: func() interface{} {
+				b := [][]byte{make([]byte, packetSizeUDP)}
+				return &b
```
ash2k: Why a pointer?

pkg/cloudproviders/aws/aws.go
tiedotguy: Yep, this isn't the expensive path though.  I'll make a note to fix it later.

pkg/statsd/receiver.go
tiedotguy: gometalinter says to use a pointer to avoid a copy.  I'm not 100% certain the need myself, but it doesn't seem to make much difference either way.  I think if I made the wrapping pool a *[][]byte and pushed some extra work in to the doneFunc I could remove it, but the whole point of the wrapper is to not deal with that.

ash2k: I think it is a bug in metalinter. Pointer is 1 machine word, slice is 3. It is more expensive to put that 3-word slice on the heap than to copy it.

tiedotguy: True story.  Looks like it just makes the copy explicit instead of implicit.  Will update.

PR #154, opened 2018-01-12
Performance work 2

tiedotguy:
Two main changes here, both aimed at addressing memory reuse, based on an 894 cpu second sample.

The first one targets object (metric) allocation in the lexer: 15.5 cpu seconds.  It looks like the lexer itself may also escape to the heap.

The second targets slice growth for tags: 35.7 cpu seconds in cloud handler, 21.9 cpu seconds in tag handler, and 23.5 cpu seconds in lexing.

There is also 112 seconds of GC, some of which should be reduced, but that's harder to quantify.

Changes are broken up by commits.

### DIFF
--- a/pkg/pool/metric_pool.go
+++ b/pkg/pool/metric_pool.go
@@ -1,0 +1,45 @@
+package pool
+
+import (
+	"sync"
+
+	"github.com/atlassian/gostatsd"
+)
+
+// MetricPool is a strongly typed wrapper around a sync.Pool for *gostatsd.Metric, it provides
+// two main benefits: 1) metrics are very short lived and we create a lot of them, 2) reuse
+// of the tags buffer
+type MetricPool struct {
+	p             sync.Pool
+}
+
+// NewMetricPool returns a new metric pool.
+func NewMetricPool() *MetricPool {
+	return &MetricPool{
+		p: sync.Pool{
+			New: func() interface{} {
+				return &gostatsd.Metric{}
+			},
+		},
+	}
+}
+
+// Get returns a *gostatsd.Metric suitable for holding a metric.  The DoneFunc should be called
+// when the metric is no longer required.  It must not be called earlier, and the Tags field may
+// be reused.
+func (mp *MetricPool) Get() *gostatsd.Metric {
+	m := mp.p.Get().(*gostatsd.Metric)
+	if m.DoneFunc != nil { // it was re-used, and the data needs cleaning
+		m.Name = ""
+		m.Value = 0
+		m.Tags = m.Tags[:0]
+		m.StringValue = ""
+		m.Hostname = ""
+		m.SourceIP = ""
+	} else {
+		m.DoneFunc = func() {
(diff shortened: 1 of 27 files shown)


pkg/statsd/handler_cloud_test.go
```diff
@@ -261,6 +261,7 @@ type countingHandler struct {
 }
 
 func (ch *countingHandler) DispatchMetric(ctx context.Context, m *gostatsd.Metric) error {
+	m.DoneFunc = nil // Clear DoneFunc because it contains non-predictable variable data
```
ash2k: It is really not obvious why this is needed here

tiedotguy: It confuses the equality check in the test.  I'll move it to be closer to the use up at line 197 and clarify the comment though.

tiedotguy: Scratch that, I'll clarify the comment.  Moving it is a bit tricky because it's not pointer based.  Even though it's tests, I don't want to break the "return to pool" semantics.

pkg/pool/metric_pool.go
```diff
@@ -0,0 +1,50 @@
...
+		estimatedTags: estimatedTags,
+	}
+}
+
+// Get returns a *gostatsd.Metric suitable for holding a metric.  The DoneFunc should be called
+// when the metric is no longer required.  It must not be called earlier, and the Tags field may
+// be reused.
+func (mp *MetricPool) Get() *gostatsd.Metric {
+	m := mp.p.Get().(*gostatsd.Metric)
+	if m.DoneFunc != nil { // it was re-used, and the data needs cleaning
```
aelse: This all looks a bit messy. What if all the field initialisation went into the Metric.Reset() and this is called from DoneFunc before returning the Metric to the MetricPool?

tiedotguy: Doing it before returning may be wasted work, because the pool is flushed every GC.  I like pushing it to another function though because it puts it near the type declaration.

pkg/pool/metric_pool.go
```diff
@@ -0,0 +1,50 @@
...
+		m.Name = ""
+		m.Value = 0
+		m.Tags = m.Tags[:0]
+		m.StringValue = ""
+		m.Hostname = ""
+		m.SourceIP = ""
+	} else {
+		m.DoneFunc = func() {
+			mp.p.Put(m)
```
aelse: call to m.Reset() before the Put

tiedotguy:
This includes no provision for user supplied tags, practically guaranteeing a grow somewhere in the pipeline, although the pool seems to help a bit.  ~60 / 840 cpu seconds are still taken in reallocation.

I'll raise a fresh PR with a flag targeting this branch so master stays clean, and the approved PR remains unchanged (since flags can be a bit verbose with boilerplate).

tiedotguy:
I've realized that the Tags field isn't reused at all, only the Pool was useful.

PR #155, opened 2018-01-16
Add estimated-tags on the command line, and re-use the Tags in the lexer

tiedotguy:
This provides improvements more in-line with what I expected:

Before (master):
     statsd_test.go:63: Processed metrics: 239965
                        TotalAlloc: 4468756296 (18622 per metric)
                        HeapObjects: 1232201
                        Mallocs: 85311296 (355 per metric)
                        NumGC: 87
                        GCCPUFraction: 0.113637

After:
     statsd_test.go:64: Processed metrics: 279938
                        TotalAlloc: 3635926720 (12988 per metric)
                        HeapObjects: 775769
                        Mallocs: 74816124 (267 per metric)
                        NumGC: 70
                        GCCPUFraction: 0.082639

Significant improvements in GC and number of metrics processed.

### DIFF
--- a/pkg/statsd/lexer.go
+++ b/pkg/statsd/lexer.go
@@ -110,6 +110,8 @@ func lexSpecial(l *lexer) stateFn {
 	default:
 		l.pos--
 		l.m = l.metricPool.Get()
+		// Pull the tags from the metric, because it may have a buffer we can reuse.
+		l.tags = l.m.Tags
 		return lexKeySep
 	}
 }
--- a/pkg/statsd/lexer.go
+++ b/pkg/statsd/lexer.go
@@ -110,7 +110,7 @@ func lexSpecial(l *lexer) stateFn {
 	default:
 		l.pos--
 		l.m = l.metricPool.Get()
-		// Pull the tags from the metric, because it may have a buffer we can reuse.
+		// Pull the tags from the metric, because it may have an empty buffer we can reuse.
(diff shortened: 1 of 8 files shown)


pkg/statsd/lexer.go
```diff
@@ -110,6 +110,8 @@ func lexSpecial(l *lexer) stateFn {
 	default:
 		l.pos--
 		l.m = l.metricPool.Get()
+		// Pull the tags from the metric, because it may have a buffer we can reuse.
+		l.tags = l.m.Tags
```
tiedotguy: Not too sure this is the best way.  Could have another pool, but (despite giving them their own package), I kind of want to avoid the proliferation of pools.

pkg/statsd/statsd_test.go
```diff
@@ -111,7 +112,7 @@ type fakeProvider struct {
 }
 
 func (fp *fakeProvider) EstimatedTags() int {
-	return 0
+	return 3
```
tiedotguy: Should really be len(fp.instance.Tags), will fix before merge.

pkg/statsd/lexer.go
aelse: "may have an empty buffer" (because if a buffer exists it was cleared in m.Reset() before the pool gave it to us, but let's not make the next person go looking for that)

PR #157, opened 2018-02-19
Fix -race build

tiedotguy:
Looks like 16.10 isn't on the supported list, and something got pulled.


ash2k:
Just `FROM ubuntu`? Would be even better to use https://github.com/GoogleCloudPlatform/distroless

tiedotguy:
Probably worth looking in to at some point.  For now I just need to unblock, and wanted minimal change.  I've raised #158

PR #160, opened 2018-03-12
Use default credentials chain

ash2k:
There is no need to provide the credential chain, config defaulting will do the right thing.


tiedotguy:
lgtm

PR #164, opened 2018-03-21
Calculate per-second rates based on actual flush interval rather than configured flush interval.

tiedotguy:


pkg/statsd/flusher.go
```diff
@@ -50,23 +50,24 @@ func (f *MetricFlusher) Run(ctx context.Context) {
 		case <-ctx.Done():
 			return
 		case <-flushTicker.C: // Time to flush to the backends
```
ash2k: `case thisFlush := <-flushTicker.C` IIRC

tiedotguy: TIL!

PR #172, opened 2018-04-06
Reuse JSON buffer

ash2k:
This removes the allocation from deflate function and uses a pool to reuse buffers.


tiedotguy:
Seems solid.  I almost wonder if it's worth using the buffer as the semaphore lock, and avoiding the sync.Pool though.

At present we'll never have more than the semaphore count of buffers, but we may have less due to GC, and after it's done flushing, it's almost guaranteed to release the entire pool before the next flush.

That being said, it should definitely cut down a bunch either way, and large buffers are expensive (although I never did find out if it's allocating or running the GC that's pricy)

ash2k:
@tiedotguy yeah, it ocurred to me after I sent this PR that buffers will be GC'ed each time and then re-allocated. We can do better, as you pointed out. Also the json library uses a buffered channel to pull objects (`BorrowStream()` and `ReturnStream()`). Should probably do the same. Don't merge yet)

PR #175, opened 2018-04-28
Pool timeSeries objects with backpressure

ash2k:
This change fixes the number of goroutines doing the send operation, each of them now owns a buffer so no pooling is needed. Time series objects are pooled to avoid allocating them each time. Their number is fixed to create backpressure so that code does not overflow the RAM with stuff that is ready to be sent when I/O cannot keep up.

p.s. I haven't tested this properly.

### DIFF
--- a/pkg/backends/datadog/datadog.go
+++ b/pkg/backends/datadog/datadog.go
@@ -19,6 +19,7 @@ import (
 	"github.com/atlassian/gostatsd"
 	stats "github.com/atlassian/gostatsd/pkg/statser"
 
+	"github.com/ash2k/stager"
 	"github.com/cenkalti/backoff"
 	jsoniter "github.com/json-iterator/go"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


pkg/backends/datadog/datadog.go
```diff
@@ -86,33 +88,53 @@ type event struct {
 	AlertType      string   `json:"alert_type,omitempty"`
 }
 
+func (d *Client) Run(ctx context.Context) {
```
ash2k: Now implements `RunnableBackend` interface.

pkg/backends/datadog/datadog.go
```diff
@@ -344,6 +364,19 @@ func (d *Client) authenticatedURL(path string) string {
...
+func (d *Client) tsBorrow() *timeSeries {
+	return <-d.tsPool
+}
+
+func (d *Client) tsReturn(ts *timeSeries) {
+	// Zero the slice's contents to remove references to strings so that GC can collect them
+	for i := range ts.Series {
+		ts.Series[i] = metric{}
+	}
+	ts.Series = ts.Series[:0]
```
tiedotguy: 372-375 shouldn't be necessary with this.  Data after the slice length should be GC'd unless there's another slice with a larger view of the data.

ash2k: GC does not understand that this data is not reachable via this slice (and also what if you re-slice it? It will become reachable). GC looks at the backing array and sees pointers and keeps their targets reachable.

https://github.com/golang/go/wiki/SliceTricks#delete-without-preserving-order

tiedotguy: That is many levels of horrifying.

tiedotguy:
@ash2k I'd like to close this one off as "won't do".  Between not being fully tested and the backend rewrite I've (mostly) done, it's probably not worth getting it over the line.  Do you mind?

ash2k:
NP at all - if you are re-writing that backend then there is no point, I agree. Lets keep the branch for a while just in case we need it for some reason.

PR #177, opened 2018-05-11
Build with go 1.10

tiedotguy:
Will be interesting to see if this helps with anything, especially around GC.


.travis.yml
```diff
@@ -6,7 +6,7 @@ services:
   - docker
 
 go:
-  - 1.9
+  - "1.10"
```
aelse: 🤦

ash2k: `1.10.2`?

ash2k:
Afaik 1.10 did not bring any significant GC improvements but we'll see. Maybe they've optimized some other things. And bugfixes of course.

PR #184, opened 2018-06-15
Make datadog user-agent configurable

tiedotguy:
I'm a bit mixed on whether this is a 7.1.0 or an 8.0.0.  Can change it if people have opinions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+7.1.0
+-----
+- Datadog user-agent has changed to `gostatsd` by default.  Can now be configured.
+- New Datadog option: `user-agent`, configures the user agent supplied to Datadog.  Use `python-requests/2.6.0 CPython/2.7.10`
+  for old behavior.
+
 7.0.1
 -----
 - Roll back change to configuration, due to https://github.com/spf13/viper/issues/380.  Documentation is still valid.
--- a/pkg/backends/datadog/datadog.go
+++ b/pkg/backends/datadog/datadog.go
@@ -30,7 +30,7 @@ const (
 	// BackendName is the name of this backend.
 	BackendName                  = "datadog"
 	dogstatsdVersion             = "5.6.3"
-	dogstatsdUserAgent           = "python-requests/2.6.0 CPython/2.7.10"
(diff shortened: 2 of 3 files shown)


pkg/backends/datadog/datadog.go
```diff
@@ -30,7 +30,7 @@ const (
 	// BackendName is the name of this backend.
 	BackendName                  = "datadog"
 	dogstatsdVersion             = "5.6.3"
-	dogstatsdUserAgent           = "python-requests/2.6.0 CPython/2.7.10"
+	defaultDogstatsdUserAgent    = "gostatsd"
```
aelse: nit: defaultUserAgent is a better name

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+7.1.0
```
aelse: Agreed on 7.1 vs 8.0

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
...
+-----
+- Datadog user-agent has changed to `gostatsd` by default.  Can now be configured.
+- New Datadog option: `user-agent`, configures the user agent supplied to Datadog.  Use `python-requests/2.6.0 CPython/2.7.10`
+  for old behavior.
```
aelse: I'm not really sold on the idea that this should be configurable. I can't see a use for this other than to masquerade as the datadog agent and then what we're really trying to do is subvert Datadog's view of which clients access it. I think this is something we could add in down the track if there are good reasons for it (eg. in the unlikely event that someone from Datadog asks us to). It's not likely that Datadog would attempt to block access to its customers by using the User-Agent, and they'd have to have a pretty good reason to do so.

pkg/backends/datadog/datadog.go
tiedotguy: Ahh, yeah.  I initially didn't change the value, only made it configurable.  Will update this.

CHANGELOG.md
tiedotguy: I prefer to have it and not need it. It's an extra bit now vs another cycle later.

PR #185, opened 2018-07-02
Datadog experimental backend (feedback only)

tiedotguy:
Replacement backend for DD I've been working on.  It's pretty close to ready, but wanted to get some feedback on current state of it before I pepper it with metrics.  PR will not be merged, just after feedback.

Currently missing:
- Benchmarks everywhere
- End to end testing (+bench)
- Internal metrics
- Events

The primary goal is to unblock the flusher, and accumulate more than a single datapoint (metricBuffer).  The single big map is periodically sliced up in to batches (metricBatcher), and sent off to a serializer+compressor (batchSerializer) / sender (bodyWriter).  These some stuff in bodyWriter for tracing, I was trying to track the pool idle count, looks like it's not possible though as there's no event for an idle timeout (unless I'm missing something).

Overall, this should enable the flush interval to always be met (although back pressure is still present, there is an upper bound on how much we will accumulate), and decouples CPU bound/network bound goroutines.

It should also reduce network traffic where flush interval isn't met, as we're only sending an additional pair of floats, instead of an entire metric + tags.  Which actually means if we don't write fast enough, that's a good thing - I might make the metricBatcher configurable to do that deliberately.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,9 @@
 /build/bin/
 /vendor/
-/coverage.out
-/profile.out
 /cmd/tester/statsd-tester*
 /test-report.xml
 build/dev/data/
 /.vscode
 *.toml
+/*.out
+/*.test
--- a/backend.go
+++ b/backend.go
@@ -2,6 +2,7 @@ package gostatsd
 
 import (
 	"context"
+	"time"
 
 	"github.com/spf13/viper"
 )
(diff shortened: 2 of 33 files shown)


tiedotguy:
There's a lot to take in, but everything should be very very modular.  I think all(?) the tests use channels only, without exposing the underlying object.

tiedotguy:
This isn't going to be done.  The core purpose of this change (splitting CPU and I/O workload) is being done in the transport refactoring work.

PR #187, opened 2018-08-03
Dedupe tags

juliawong:
We sometimes receive duplicate tags for a metric. This PR will only add unique tags to the metric or event.

The other approach would be to consolidate the dupe tags into one tag. e.g. from `my_tag:my_value,my_tag:my_value,my_tag:my_value` to `my_tag:my_value__my_value__my_value`. Thoughts on this?

### DIFF
--- a/pkg/statsd/handler_tags.go
+++ b/pkg/statsd/handler_tags.go
@@ -56,14 +56,14 @@ func uniqueTags(t1 gostatsd.Tags, t2 gostatsd.Tags) gostatsd.Tags {
 	seen := map[string]bool{}
 
 	for _, v := range t1 {
-		if seen[v] != true {
+		if !seen[v] {
 			tags = append(tags, v)
 			seen[v] = true
 		}
 	}
 
 	for _, v := range t2 {
-		if seen[v] != true {
+		if !seen[v] {
 			tags = append(tags, v)
 			seen[v] = true
 		}
--- a/pkg/statsd/handler_tags.go
+++ b/pkg/statsd/handler_tags.go
@@ -13,7 +13,7 @@ type TagHandler struct {
 	estimatedTags int
 }
 
-// NewTagHandler initialises a new handler which adds tags and sends metrics/events to the next handler
+// NewTagHandler initialises a new handler which adds unqiue tags and sends metrics/events to the next handler
 func NewTagHandler(metrics MetricHandler, events EventHandler, tags gostatsd.Tags) *TagHandler {
 	return &TagHandler{
 		metrics:       metrics,
@@ -28,7 +28,7 @@ func (th *TagHandler) EstimatedTags() int {
 	return th.estimatedTags
 }
 
-// DispatchMetric adds the tags from the TagHandler to the metric and passes it to the next stage in the pipeline
+// DispatchMetric adds the unqiue tags from the TagHandler to the metric and passes it to the next stage in the pipeline
 func (th *TagHandler) DispatchMetric(ctx context.Context, m *gostatsd.Metric) error {
 	if m.Hostname == "" {
 		m.Hostname = string(m.SourceIP)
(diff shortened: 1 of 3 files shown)


juliawong:
# Benchmark results

1. `go test -v -run=XXX -bench=BenchmarkTagMetricHandlerAddsDuplicateTags* -count=1 ./pkg/statsd`

Before:
```
goarch: amd64
pkg: github.com/atlassian/gostatsd/pkg/statsd
BenchmarkTagMetricHandlerAddsDuplicateTagsSmall-8   	10000000	       163 ns/op	     170 B/op	       1 allocs/op
BenchmarkTagMetricHandlerAddsDuplicateTagsLarge-8   	10000000	       121 ns/op	     170 B/op	       1 allocs/op
```

After:
```
goos: darwin
goarch: amd64
pkg: github.com/atlassian/gostatsd/pkg/statsd
BenchmarkTagMetricHandlerAddsDuplicateTagsSmall-8   	 5000000	       407 ns/op	     219 B/op	       3 allocs/op
BenchmarkTagMetricHandlerAddsDuplicateTagsLarge-8   	 2000000	       603 ns/op	     412 B/op	       5 allocs/op
```

2. go test -v -run=Throughput -count=1 ./pkg/statsd

Before:
```
=== RUN   TestStatsdThroughput
--- PASS: TestStatsdThroughput (7.03s)
	statsd_test.go:64: Processed metrics: 280524
			TotalAlloc: 5243524680 (18691 per metric)
			HeapObjects: 1015389
			Mallocs: 93191455 (332 per metric)
			NumGC: 69
			GCCPUFraction: -0.000025
```

After:
```
=== RUN   TestStatsdThroughput
--- PASS: TestStatsdThroughput (7.02s)
	statsd_test.go:64: Processed metrics: 255408
			TotalAlloc: 5871415840 (22988 per metric)
			HeapObjects: 924860
			Mallocs: 112551052 (440 per metric)
			NumGC: 84
			GCCPUFraction: -0.000028
```

pkg/statsd/handler_tags.go
```diff
@@ -28,25 +28,46 @@ func (th *TagHandler) EstimatedTags() int {
...
 }
 
 // WaitForEvents waits for all event-dispatching goroutines to finish.
 func (th *TagHandler) WaitForEvents() {
 	th.events.WaitForEvents()
 }
+
+func uniqueTags(t1 gostatsd.Tags, t2 gostatsd.Tags) gostatsd.Tags {
+	tags := gostatsd.Tags{}
+	seen := map[string]bool{}
```
ash2k: Usually sets are modelled with `map[x]struct{}` - see e.g. https://www.davidkaya.sk/2017/12/10/sets-in-golang/
It is better for these reasons:
- it is idiomatic in Go. Others will immediately understand the point of code. If value is of type `struct{}` that means that value... has no value, only the presence/absence of it is valuable.
- More memory efficient.

pkg/statsd/handler_tags.go
```diff
@@ -28,25 +28,46 @@ func (th *TagHandler) EstimatedTags() int {
...
+
+	for _, v := range t1 {
```
ash2k: You could have two nested for loops to iterate over a slice of `t1` and `t2`. In that case the'd be no need to duplicate the loop body. You could also change the signature to accept a variable number of Tags and then the slice will be created for you. Up to you.

pkg/statsd/handler_tags.go
```diff
@@ -28,25 +28,46 @@ func (th *TagHandler) EstimatedTags() int {
...
+	e.Tags = uniqueTags(e.Tags, th.tags)
 	return th.events.DispatchEvent(ctx, e)
 }
 
 // WaitForEvents waits for all event-dispatching goroutines to finish.
 func (th *TagHandler) WaitForEvents() {
 	th.events.WaitForEvents()
 }
+
+func uniqueTags(t1 gostatsd.Tags, t2 gostatsd.Tags) gostatsd.Tags {
```
ash2k: FYI no need to specify the type for every consecutive argument of the same type. This can be `uniqueTags(t1, t2 gostatsd.Tags)`.

PR #190, opened 2018-08-28
Add support for sampling timers

a-kr:
**Problem**:

`gostatsd` does not support sample rate for timer metrics, only for counters.
Original Etsy's `statsd` implementation supports sample rate for timers since 2013: https://github.com/etsy/statsd/pull/243/commits/cc12534e3b3bf5137a263e4e065876329e1a719c

**Situation**:

Consider a generic API service which does performance tracking by sending a single `timer` metric for each request to its endpoint. Common statsd-carbon-graphite pipeline (or any equivalent one) produces both mean / percentile response time data and requests-per-second data from `timer` metrics, which is sufficient for most monitoring situations.

Now, suppose this service handles quite a lot of requests per second. In this situation is reasonable to use metric sampling and submit only, say, 50% of timer metrics:

    megaservice.supernode-14.get_customers:110|ms|@0.5
    
This works well with Etsy's `statsd`: values of `count` and `count_ps` generated from a timer metric are divided by the sample rate, producing a (reasonable estimate of a) correct requests-per-second data. `mean`, `median`, `upper_95`, `sum`, and other aggregated values which are measured in seconds are not affected by sampling.

With `gostatsd`, however, sampling works only for counters and not for timers. So if I use `gostatsd` and submit timer metrics with 50% sample rate, I get `count_ps` values which are 2 times lower than the actual number of requests per second.

**Note**:

Etsy's `statsd` implementation assumes that, within a single flush interval, a single timer metric is submitted with the same sample rate:

    megaservice.supernode-14.get_customers:54|ms|@0.5
    megaservice.supernode-14.get_customers:22|ms|@0.5
    megaservice.supernode-14.get_customers:5553|ms|@0.5

If this assumption holds, then only the counters (`count`, `count_ps`) need to be adjusted for samping, and mean / percentile / etc time values can be calculated as normal.

If this assumption breaks, and we recieve metrics with different sample rates within a single flush interval...

    megaservice.supernode-14.get_customers:54|ms|@0.1
    megaservice.supernode-14.get_customers:22|ms|@0.5
    megaservice.supernode-14.get_customers:5553|ms|@0.25

...then all aggregations (mean / percentile / etc) should be somehow computed with regard to the sample rate of each individual metric: mean becomes weighted average, etc; or else they would not make sense statistically.

In practice, it is hard to imagine why would one need to change the sample rate continuously, and how would that be implemented on the client side. Most likely the sample rate is chosen by a developer and stays the same for long periods of time. When it does eventually change, only a single flush interval would be affected.

So I've chosen to keep the assumption of a single, rarely-changing sample rate per metric --- same assumption that is used in Etsy's `statsd`.

**Summary**:

This pull request implements support for sample rate for timer counters in `gostatsd`, in original-Etsy's-`statsd`-compatible way.

### DIFF
--- a/pkg/statsd/aggregator.go
+++ b/pkg/statsd/aggregator.go
@@ -85,14 +85,14 @@ func (a *MetricAggregator) Flush(flushInterval time.Duration) {
 			sort.Float64s(timer.Values)
 			timer.Min = timer.Values[0]
 			timer.Max = timer.Values[count-1]
-			timer.Count = len(timer.Values)
-			count := float64(timer.Count)
+			n := len(timer.Values)
+			count := float64(n)
 
-			cumulativeValues := make([]float64, timer.Count)
-			cumulSumSquaresValues := make([]float64, timer.Count)
+			cumulativeValues := make([]float64, n)
+			cumulSumSquaresValues := make([]float64, n)
 			cumulativeValues[0] = timer.Min
 			cumulSumSquaresValues[0] = timer.Min * timer.Min
-			for i := 1; i < timer.Count; i++ {
+			for i := 1; i < n; i++ {
 				cumulativeValues[i] = timer.Values[i] + cumulativeValues[i-1]
 				cumulSumSquaresValues[i] = timer.Values[i]*timer.Values[i] + cumulSumSquaresValues[i-1]
 			}
@@ -103,8 +103,8 @@ func (a *MetricAggregator) Flush(flushInterval time.Duration) {
 			var thresholdBoundary = timer.Max
 
 			for pct, pctStruct := range a.percentThresholds {
-				numInThreshold := timer.Count
-				if timer.Count > 1 {
+				numInThreshold := n
+				if n > 1 {
 					numInThreshold = int(round(math.Abs(pct) / 100 * count))
 					if numInThreshold == 0 {
 						continue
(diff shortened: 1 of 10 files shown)


pkg/statsd/aggregator.go
```diff
@@ -284,20 +287,23 @@ func (a *MetricAggregator) receiveGauge(m *gostatsd.Metric, tagsKey string, now
...
 func (a *MetricAggregator) receiveTimer(m *gostatsd.Metric, tagsKey string, now gostatsd.Nanotime) {
 	v, ok := a.Timers[m.Name]
-	if ok {
-		t, ok := v[tagsKey]
-		if ok {
-			t.Values = append(t.Values, m.Value)
-			t.Timestamp = now
-		} else {
-			t = gostatsd.NewTimer(now, []float64{m.Value}, m.Hostname, m.Tags)
+	if !ok {
```
tiedotguy: This restructuring changes it away from the other 3 ```receive*``` calls.  I generally agree that it's a cleaner pattern, however historically whenever I've done this, there's a performance impact (not a lot, but present) that I can't justify just for the style change.

tiedotguy:
Can you please also ensure the CLA is signed as appropriate.

pkg/statsd/aggregator.go
a-kr: Fixed, reverted `receiveTimer` to original structure

a-kr:
Thanks for the review.
* Added a note to the changelog;
* reverted refactoring in `receiveTimer`;
* pushed the changes as `fixup!` commits (will squash them if everything is ok).

Signed _Atlassian Individual CLA_.

tiedotguy:
I'll keep an eye out for the CLA on our side (unfortunately it's still a bit manual), but it looks good to me. 
 I'll get it released once that shows up.

Thanks!

a-kr:
Any news on CLA? If it helps, I've signed it on August 28, as Alexey Kryuchkov.

Not sure what can be done about the failing CI build --- `make test-race` consistently passes on my machine. Any way to restart it?

tiedotguy:
Sorry for the delays - finally came in overnight.  Not sure about the test, looks like a flake.  Trying to repro it locally.  If I can I'll raise an issue, otherwise call it cosmic rays from space.

PR #191, opened 2018-08-30
Add tag level filtering + a couple of minor adjustments

tiedotguy:


pkg/statsd/filtering.go
```diff
@@ -0,0 +1,43 @@
...
+func toStringMatch(tests []string) []gostatsd.StringMatch {
+	var matches []gostatsd.StringMatch
+	for _, test := range tests {
+		matches = append(matches, gostatsd.NewStringMatch(test))
+	}
+	return matches
+}
+
+// NewFilterFromViper creates a new Filter given a *viper.Viper
+func NewFilterFromViper(name string, v *viper.Viper) Filter {
```
hligit: `name` is not used?

tiedotguy: Good spot.  It was at one point, I was going to emit metrics on the name.  I'll ditch it.

pkg/statsd/filtering.go
```diff
@@ -0,0 +1,43 @@
...
+	ExcludeMetrics gostatsd.StringMatchList // Name must not match
+	MatchTags      gostatsd.StringMatchList // Any tag must match
+	DropTags       gostatsd.StringMatchList // Any tag matching anything will be dropped
+	DropMetric     bool                     // Drop the entire metric
+	DropHost       bool                     // Clears Hostname if present
+}
+
+// toStringMatch turns a []string in to a []gostatsd.StringMatch
+func toStringMatch(tests []string) []gostatsd.StringMatch {
+	var matches []gostatsd.StringMatch
```
ash2k: `matches` can be pre-allocated

pkg/statsd/handler_tags.go
```diff
@@ -34,7 +54,52 @@ func (th *TagHandler) DispatchMetric(ctx context.Context, m *gostatsd.Metric) er
...
+			// no tags match
+			continue
+		}
+
+		if filter.DropMetric {
+			return false
+		}
+
+		// TODO: This is not the most efficient.  It might be better to replace with an empty string, and ensure empty
+		// tags are handled downstream.
```
aelse: Current implementation does _n_ copies for _m_ filters, or O(n*m) copies. If you iterate over the tags and then test each filter in turn you can make this O(n) copies. If we don't care about preserving order then we can reduce that further to just the number of dropped tags, which may be zero a lot of the time.

pkg/statsd/handler_tags.go
```diff
@@ -51,21 +116,24 @@ func (th *TagHandler) WaitForEvents() {
...
+			seen[v] = present
 		}
 	}
 
 	for _, v := range t2 {
-		if !seen[v] {
+		if _, ok := seen[v]; !ok {
 			tags = append(tags, v)
-			seen[v] = true
+			seen[v] = present
```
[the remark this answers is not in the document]
tiedotguy: :mindblown:

pkg/statsd/handler_tags.go
tiedotguy: Yeah, it's pretty bad.  I'll collect the tags to drop and pass it to ```uniqueTags```.  It needs to build a list regardless.

pkg/statsd/handler_tags.go
```diff
@@ -51,21 +116,24 @@ func (th *TagHandler) WaitForEvents() {
 	th.events.WaitForEvents()
 }
 
+var present = struct{}{}
+
+// uniqueTags returns the set of t1 | t2.
 func uniqueTags(t1 gostatsd.Tags, t2 gostatsd.Tags) gostatsd.Tags {
 	tags := gostatsd.Tags{}
```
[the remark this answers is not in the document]
tiedotguy: I tried a few variations here, and struggled to get any bump that wasn't down in the noise. I think I tried that on the map, but not the tags, I'll give it a shot there.

pkg/statsd/handler_tags.go
aelse: It doesn't need to do _n_ copies though. We're destructively appending to a slice backed by the same array. You can delete in place by replacing m.Tags[i] with m.Tags[len(m.Tags)-1] and resize the slice.

    a[i] = a[len(a)-1] 
    a = a[:len(a)-1]

https://github.com/golang/go/wiki/SliceTricks

tiedotguy: Ahh, got it.  I can incorporate that.

pkg/statsd/handler_tags.go
```diff
@@ -51,23 +132,31 @@ func (th *TagHandler) WaitForEvents() {
...
-	seen := map[string]bool{}
+	return uniqueTagsWithSeen(map[string]struct{}{}, t1, t2)
+}
 
-	for _, v := range t1 {
-		if !seen[v] {
-			tags = append(tags, v)
-			seen[v] = true
+// uniqueTags returns the set of (t1 | t2) - seen.  It may modify the contents of t1, t2, and seen.
+func uniqueTagsWithSeen(seen map[string]struct{}, t1 gostatsd.Tags, t2 gostatsd.Tags) gostatsd.Tags {
```
aelse: Looks ok but i'd like to see unit tests

PR #197, opened 2018-10-31
Correct NR config option to hyphen

Kav91:


tiedotguy:
Good catch!

I'll cut a new version later tonight.

PR #199, opened 2018-11-07
fix the legacy counter namespace

BLIII:
The prefix for counters in statsd was "counters" which is not reflected in the legacy namespaces for graphite in gostatsd. The reference to this in statsd can be found [here](https://github.com/etsy/statsd/blob/8d5363cb109cc6363661a1d5813e0b96787c4411/docs/namespacing.md)

### DIFF
--- a/pkg/backends/graphite/graphite.go
+++ b/pkg/backends/graphite/graphite.go
@@ -218,7 +218,7 @@ func NewClient(config *Config, disabled gostatsd.TimerSubtypes) (*Client, error)
 		legacyNamespace = DefaultLegacyNamespace
 	}
 	if legacyNamespace {
-		counterNamespace = DefaultGlobalPrefix + `.`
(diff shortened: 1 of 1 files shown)


tiedotguy:
From the reference:
```
In the legacy setting rates were recorded under stats.counter_name directly, whereas the absolute count could be found under stats_counts.counter_name.
```

This (with an optional suffix) is how it's emitted currently when legacy is enabled:
- rate: stats.name
- absolute count: stats_counters.name

It would change to:
- rate: stats.counters.name
- absolute count: stats_counters.name

Personally I don't think either is great, and if I was running a graphite backend, I'd likely turn off legacy mode entirely (with suitable migration, etc)

BLIII:
Thank you for the note @tiedotguy! After looking through the code again, I believe setting the legacy_namespace variable to false in the config is exactly what I needed. Thank you for pointing me in the right direction with this. I am closing this PR as it is not needed.

PR #204, opened 2018-12-02
[cloudwatch] Batch requests to Cloudwatch to reduce cost

JorgenEvens:
Cloudwatch is billed per request and by batching requests to Cloudwatch we can significantly reduce the cost of running statsd with Cloudwatch as configured backend.

screenshot 2018-12-02 21 53 49

### DIFF
--- a/pkg/backends/cloudwatch/cloudwatch.go
+++ b/pkg/backends/cloudwatch/cloudwatch.go
@@ -3,6 +3,7 @@ package cloudwatch
 import (
 	"context"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/atlassian/gostatsd"
@@ -21,10 +22,23 @@ const MAX_DIMENSIONS = 10
 // BackendName is the name of this backend.
 const BackendName = "cloudwatch"
 
+type QueueItem struct {
+	metrics  []*cloudwatch.MetricDatum
+	callback *gostatsd.SendCallback
+}
+
 // Client is an object that is used to send messages to AWS CloudWatch.
 type Client struct {
 	cloudwatch cloudwatchiface.CloudWatchAPI
 	namespace  string
+
+	batchDuration time.Duration
+
+	queue      []*QueueItem
+	queueMutex sync.RWMutex
+
+	timer      *time.Timer
+	timerMutex sync.Mutex
 
 	disabledSubtypes gostatsd.TimerSubtypes
 }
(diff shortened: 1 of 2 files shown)


pkg/backends/cloudwatch/cloudwatch.go
```diff
@@ -157,59 +176,115 @@ func (client Client) buildMetricData(metrics *gostatsd.MetricMap) (metricData []
...
-	length := len(metricData)
-	errors := []error{}
+func (client *Client) flush() {
+	client.timerMutex.Lock()
 
-	if length < 1 {
-		cb(errors)
+	// If there already is a timer, we already have a function waiting to execute
+	if client.timer != nil {
+		client.timer.Reset(client.batchDuration)
```
tiedotguy: The logic on this works because of the mutex, but I did have to work through it, which should be unnecessary cognitive load.

Can you add a comment to the effect that even if `client.timer` has already fired, everything will still process due to the mutex?

pkg/backends/cloudwatch/cloudwatch.go
```diff
@@ -21,34 +22,52 @@ const MAX_DIMENSIONS = 10
 // BackendName is the name of this backend.
 const BackendName = "cloudwatch"
 
+type QueueItem struct {
+	metrics  []*cloudwatch.MetricDatum
+	callback *gostatsd.SendCallback
```
tiedotguy: Is there any reason for this to be a pointer?  It makes the actual invocation rather ugly, and I don't see any actual need for it? (although I might be missing something)

tiedotguy:
Nice savings! Do you mind if I ask what your data volume is?

wooly:
Is this able to be merged if the conflicts are resolved?

tiedotguy:
Yep - it also needs the comments addressed though, because there's a critical issue in there (the race)

 Summary:
- [ ] `QueueItem.callback` doesn't need to be a callback (I believe)
- [ ] `client.flush` needs a comment clarifying what's going on
- [ ] `log.Infof` should be changed to send metrics, or removed (probably removed)
- [ ] `client.SendMetricAsync` buffer has a race, the mutex can be changed to `Lock` instead of `RLock`
- [ ] `client.queueMutex` can be changed to a regular `sync.Mutex` because of the above
- [ ] `batchDuration` flag should be changed to `batch-duration`, and needs documentation (BACKENDS.md)

pkg/backends/cloudwatch/cloudwatch.go
```diff
@@ -21,34 +22,52 @@ const MAX_DIMENSIONS = 10
...
+	timerMutex sync.Mutex
+
 	disabledSubtypes gostatsd.TimerSubtypes
 }
 
 // NewClientFromViper constructs a Cloudwatch backend.
 func NewClientFromViper(v *viper.Viper) (gostatsd.Backend, error) {
 	g := getSubViper(v, "cloudwatch")
 	g.SetDefault("namespace", "StatsD")
+	g.SetDefault("batchDuration", "50ms")
```
tiedotguy: `batchDuration` should be `batch-duration` for consistency, also needs documentation.

tiedotguy:
@wooly , were you planning to work on the PR, or were you giving a nudge?  I don't have access to any CW, so I can't properly test any changes if I clean it up.

JorgenEvens:
@wooly @tiedotguy If I ever find the time I'd love to help with this. But I've not had the time to work on this and we are no longer using CW at my daytime job.

wooly:
@tiedotguy unfortunately my go chops aren't good enough to be able to fix the issues, so it was more of a nudge!

tiedotguy:
No problem.  I'll take a stab at it soon, possibly this weekend.  The actual changes should all be fairly simple.  I do need to caveat that they'll be tested only with the tests provided by the package, so I'll probably cut a release candidate and ask you to validate it in a non-production environment, if you can arrange that.

JorgenEvens:
@tiedotguy That can be arranged.

wooly:
@tiedotguy we can also sort that out for you, yep

pkg/backends/cloudwatch/cloudwatch.go
tiedotguy: Turns out there was no documentation for CW anyway /facepalm

tiedotguy:
I've raised a fresh PR as #279, just to avoid any permissions issues pushing to your fork, @JorgenEvens (although I'm not sure if there would actually be any issues).

@wooly - I can't tag you on that PR, as I guess you're not involved with the project, but FYI.

It's released to docker as 19.2.0-rc1, if you can both give it a "real" test.

I'll close this PR off, and we can continue in #279.

PR #206, opened 2018-12-11
Add the HTTP endpoint with stub status URL

dmitryilyin:
Add primitive HTTP status endpoint with a always passing health check instead of unused Profile.
Rename Profile into Endpoint.

This url can be used by load balancers with tcp checks to determine is gostatsd is running.
More sensible checks may be added later.


tiedotguy:
Hi - bit awkward, I have a branch which adds this (and more), which I'm aiming to get in soon.  Most of my recent PRs have been laying the foundations (because it was ultimately a big change, I wanted to keep the review manageable)

I expect to have a PR in for it after #205 is merged.  Hopefully I can get it merged today, then the http change raised tomorrow.

tiedotguy:
It took way longer than I wanted, but 11.0.0 just released has facility for a healthcheck endpoint.  I recommend against using the http transport for moving metrics at present, but the healthcheck is safe (although it just returns 200)

PR #207, opened 2018-12-13
Newrelic events

TylerLubeck:
I thought I'd take a stab at implementing NewRelic style events. I largely copied from how it's done for datadog.

### DIFF
--- a/pkg/backends/newrelic/newrelic.go
+++ b/pkg/backends/newrelic/newrelic.go
@@ -202,12 +202,18 @@ func (n *Client) SendEvent(ctx context.Context, e *gostatsd.Event) error {
 
 		m := make(map[string][]map[string]string)
 		new_event := make(map[string]string)
-		new_event["summary"] = e.Title
-		m["events"] = append(m["events"], new_event)
 
+		/*
+		* NewRelic Events support a summary and a category rather than a
+		* summary and a title. There isn't a clear mapping between these fields
+		* either. Rathern than losing all of the information about an event,
+		* send just the title as the summary of the event.
+		 */
+		new_event["summary"] = e.Title
+
+		m["events"] = append(m["events"], new_event)
 		return n.post(ctx, buffer, m)
 	}
-	return nil
 }
 
 // Name returns the name of the backend.
--- a/pkg/backends/newrelic/newrelic.go
+++ b/pkg/backends/newrelic/newrelic.go
@@ -34,7 +34,8 @@ const (
 	// defaultMetricsPerBatch is the default number of metrics to send in a single batch.
 	defaultMetricsPerBatch = 1000
 	// maxResponseSize is the maximum response size we are willing to read.
-	maxResponseSize = 10 * 1024
+	maxResponseSize     = 10 * 1024
+	maxConcurrentEvents = 20
 
 	defaultEnableHttp2 = false
 )
@@ -75,6 +76,7 @@ type Client struct {
 	client                http.Client
 	metricsPerBatch       uint
 	metricsBufferSem      chan *bytes.Buffer // Two in one - a semaphore and a buffer pool
+	eventsBufferSem       chan *bytes.Buffer // Two in one - a semaphore and a buffer pool
 	now                   func() time.Time   // Returns current time. Useful for testing.
 
 	disabledSubtypes gostatsd.TimerSubtypes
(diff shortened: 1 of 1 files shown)


TylerLubeck:
I slept on it and had a few more ideas, but would be interested in feedback before I chase any particular one. 

- Make the category configurable rather than relying on NR's default "notification". I think this would be one category for all events sent to NR, set in the `[newrelic]` config block.
- If the Title has no whitespace, treat it as the category and treat the Text as the summary. Drop all other events. Add a `eventsDropped` (akin to `batchesDropped`) that counts these events, and log the events that have been dropped. The downside here is that if someone wants to use NR then it forces them to make all their events conform to this style.

I also think there could be some benefit in using a `flush` for events as well - perhaps utilizing a single `flush` for each `Client` rather than a new one per batch of metrics. I haven't thought this one fully through yet, but if there's interest then I'd be happy to tackle it in a different PR.

pkg/backends/newrelic/newrelic.go
```diff
@@ -189,7 +191,29 @@ func (n *Client) processMetrics(metrics *gostatsd.MetricMap, cb func(*timeSeries
...
+	case <-ctx.Done():
+		return ctx.Err()
+	case buffer := <-n.eventsBufferSem:
+		defer func() {
+			buffer.Reset()
+			n.eventsBufferSem <- buffer
+		}()
+
+		m := make(map[string][]map[string]string)
+		new_event := make(map[string]string)
```
tiedotguy: In golang, it's camelCase rather than snake_case

pkg/backends/newrelic/newrelic.go
```diff
@@ -189,7 +191,29 @@ func (n *Client) processMetrics(metrics *gostatsd.MetricMap, cb func(*timeSeries
...
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case buffer := <-n.eventsBufferSem:
+		defer func() {
+			buffer.Reset()
+			n.eventsBufferSem <- buffer
+		}()
+
+		m := make(map[string][]map[string]string)
```
tiedotguy: This is a pretty rough type to reason about, and it's leaning more towards the pythonic style.  I would much prefer to see proper structure definitions.

pkg/backends/newrelic/newrelic.go
```diff
@@ -189,7 +191,29 @@ func (n *Client) processMetrics(metrics *gostatsd.MetricMap, cb func(*timeSeries
...
+		new_event := make(map[string]string)
+
+		/*
+		* NewRelic Events support a summary and a category rather than a
+		* summary and a title. There isn't a clear mapping between these fields
+		* either. Rathern than losing all of the information about an event,
+		* send just the title as the summary of the event.
+		 */
```
tiedotguy: Do NR events support other fields beyond summary and category?  There's quite a lot of data from the event being lost here.

tiedotguy:
> Make the category configurable rather than relying on NR's default "notification". I think this would be one category for all events sent to NR, set in the [newrelic] config block.

Is the category something that might make sense to pull out of a user supplied tag, with a fallback to something in the configuration block?

>  Add a eventsDropped (akin to batchesDropped) that counts these events, and log the events that have been dropped.

`batchesDropped` is a signal that something is horribly wrong, as it's reporting "I tried and failed to talk to the monitoring platform", so I would recommend naming the reporting metric something more along the theme of `backend.bad_events` (see also: `parser.bad_lines_seen` - that's something we track, and occasionally we notify people that they're doing something wrong, but we would never hook it up to the pager).

Also for logging the bad lines, we put it through a rate limiter, to protect against floods in our logs.  I would highly recommend that.

> I also think there could be some benefit in using a flush for events as well

A lot of the changes (or lack of) is driven by our environment, which is a) not bottlenecking on events, and b) sending to Datadog, which supports one event per request.  Hence why we've never bothered to work on that code path :)

You may want to have a look at my `datadog-experimental-backend` branch, which refactors the Datadog metrics to do async batching.  If you want to pick up batching in events, there may be some value in there, I'm not too sure though.

pkg/backends/newrelic/newrelic.go
Kav91: NR Events, can store a max of 255 attributes per event which is essentially a flat JSON object, so there is no reason to drop all the other data. 

Also the comment doesn't make much sense to me with how the NR platform works, and since the gostatsd.Event is pretty flat, all attributes could be retained quite easily. Probably just do some looping for tags to flatten like I've done for the newrelic backend.

TylerLubeck:
I appreciate the review, and I learned a bunch, but I think if #208 is being merged then this is redundant so I'm gonna close it for now. Thanks!

PR #208, opened 2018-12-14
Add support to flush directly to New Relic Insights Event API 

Kav91:
Additional details provided & updated within readme.

### DIFF
--- a/pkg/backends/newrelic/newrelic.go
+++ b/pkg/backends/newrelic/newrelic.go
@@ -388,7 +388,7 @@ func NewClient(address, eventType, flushType, apiKey, tagPrefix,
 	if maxRequestElapsedTime <= 0 {
 		return nil, fmt.Errorf("[%s] maxRequestElapsedTime must be positive", BackendName)
 	}
-	if !contains(flushTypes, flushType) {
+	if !contains(flushTypes, flushType) && flushType != "" {
 		return nil, fmt.Errorf("[%s] flushType (%s) is not supported", BackendName, flushType)
 	}
 	if flushType == flushTypeInsights && apiKey == "" {
--- a/pkg/backends/newrelic/newrelic.go
+++ b/pkg/backends/newrelic/newrelic.go
@@ -2,6 +2,7 @@ package newrelic
 
 import (
 	"bytes"
+	"compress/zlib"
 	"context"
 	"crypto/tls"
 	"encoding/json"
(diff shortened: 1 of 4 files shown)


pkg/backends/newrelic/newrelic.go
```diff
@@ -303,10 +328,23 @@ func NewClientFromViper(v *viper.Viper) (gostatsd.Backend, error) {
 	nr.SetDefault("enable-http2", defaultEnableHttp2)
 	nr.SetDefault("user-agent", defaultUserAgent)
 
+	//New Relic Config Defaults & Recommendations
+	v.SetDefault("statser-type", "null")
+	v.SetDefault("flush-interval", "10s")
+	if v.GetDuration("flush-interval").Seconds() < 10 {
+		log.Warnf("[%s] flush-interval below 10s, recommended to be >= 10s", BackendName)
+	} else if v.GetDuration("flush-interval").Seconds() == 10 {
+		log.Infof("[%s] flush-interval default 10s", BackendName)
```
tiedotguy: It doesn't make a lot of sense to only log this when it's 10s, I'd just make it an "else" not an "else if".  That being said, I  suggest pushing this down to line 393.

pkg/backends/newrelic/newrelic.go
```diff
@@ -303,10 +328,23 @@ func NewClientFromViper(v *viper.Viper) (gostatsd.Backend, error) {
...
+	}
+	if v.GetString("statser-type") == "null" {
+		log.Infof("[%s] internal metrics OFF, to enable set 'statser-type' to StatserInternal", BackendName)
```
tiedotguy: The internal statser type is `internal` (actually it's the default switch case, but "internal" is the intended type :)), maybe something more like "to enable set 'statser-type' to 'logging' or 'internal'"?

Kav91: oh yes, of course thats much better 👍

pkg/backends/newrelic/newrelic.go
Kav91: Shuffled and switched to else.

TylerLubeck:
Hi @tiedotguy! This code is going to unblock a bunch of work for me. Any chance there's an update on the release?

tiedotguy:
Yep, should have done this sooner, sorry!

PR #215, opened 2019-02-05
Add http transport

tiedotguy:
Sadly, a pretty big change.

The coverage shows handler_http_forwarder_v1 as having very little coverage, however that's because it's mostly exercised through another package, so it's not covered.  Not sure a good way to address that.

Generally, the package layout is a bit of a mess, which results in some less than optimal things.

@ash2k - note that this is not meant to be efficient yet (ie, it sends all the raw metrics over instead of doing any sort of local pre-processing).  This is just a step to get http in.


metrics.go
```diff
@@ -99,6 +99,15 @@ type MetricMap struct {
 	Sets     Sets
 }
 
+func NewMetricMap() *MetricMap {
```
tiedotguy: surprisetool

pkg/statsd/handler_http_forwarder_v1.go
```diff
@@ -0,0 +1,494 @@
...
+	messagesRetried uint64 // atomic - retries (first send is not a retry, final failure is not a retry)
+	messagesDropped uint64 // atomic - final failure
+
+	logger                logrus.FieldLogger
+	apiEndpoint           string
+	maxRequestElapsedTime time.Duration
+	metricsSem            chan struct{}
+	client                http.Client
+	compress              bool
+	consolidator          gostatsd.RawMetricHandler
```
tiedotguy: Naming on these two fields is off, should be batcher, not consolidator.

pkg/web/http_healthchecker.go
```diff
@@ -0,0 +1,24 @@
...
+}
+
+// healthCheck reports if the server is ready to process traffic.  It does not validate downstream dependencies.
+func (hc *healthChecker) healthCheck(w http.ResponseWriter, req *http.Request) {
+	hc.logger.Info("healthCheck")
+	_, _ = w.Write([]byte("OK"))
+}
+
+// deepCheck reports on the status of downstream dependencies.  It should be non-blocking (ie, report on the status
+// of watchdogs, rather than make roundtrips).  For now, it just reports that it's always happy.
```
juliawong: What are `watchdogs`?

tiedotguy: https://en.wikipedia.org/wiki/Watchdog_timer

The basic idea is if you don't pat the watchdog, the watchdog kills you.

So as an example: when we get to clustering, rather than pinging Redis every time `/deepcheck` is queried, we would "pat the watchdog" (save a timestamp) whenever we get an update from Redis about a neighboring host.  `/deepcheck` would then check the watchdog (timestamp), and if it's too old, it would report a failure.

The example isn't exactly the same, but it's close enough.

pkg/web/http_receiver_v1.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+	req.Body.Close()
+
+	encoding := req.Header.Get("Content-Encoding")
+	switch encoding {
+	case "deflate":
+		b, err = decompress(b)
+		if err != nil {
+			atomic.AddUint64(&rhh.requestFailureDecompress, 1)
+			rhh.logger.WithError(err).Info("failed decompressing body")
```
aelse: i think this is a debug level log. We're also capturing a metric for it.

pkg/web/http_receiver_v1.go
```diff
@@ -0,0 +1,228 @@
...
+			return nil, http.StatusBadRequest
+		}
+	case "identity", "":
+		// no action
+	default:
+		atomic.AddUint64(&rhh.requestFailureEncoding, 1)
+		if len(encoding) > 64 {
+			encoding = encoding[0:64]
+		}
+		rhh.logger.WithField("encoding", encoding).Info("invalid encoding")
```
aelse: same again

pkg/web/http_receiver_v1.go
```diff
@@ -0,0 +1,228 @@
...
+	statser.Count("http.incoming", float64(requestFailureUnmarshal), []string{"result:failure", "failure:unmarshal"})
+	statser.Count("http.incoming.metrics", float64(metricsProcessed), nil)
+	statser.Count("http.incoming.events", float64(eventsProcessed), nil)
+}
+
+func (rhh *rawHttpHandlerV1) readBody(req *http.Request) ([]byte, int) {
+	b, err := ioutil.ReadAll(req.Body)
+	if err != nil {
+		atomic.AddUint64(&rhh.requestFailureRead, 1)
+		rhh.logger.WithError(err).Info("failed reading body")
```
aelse: same again

PR #216, opened 2019-02-08
Extract MetricMap

tiedotguy:
This is a "small" isolated commit to extract the MetricMap Receive logic out from the MetricAggregator.

It's mostly noise and moving things around unfortunately, but better the noise be in its own PR than in what follows.

This will be re-pointed to master once http-transport is merged.

### DIFF
--- a/metric_map.go
+++ b/metric_map.go
@@ -1,0 +1,229 @@
+package gostatsd
+
+import (
+	"bytes"
+	"fmt"
+	"time"
+
+	"github.com/sirupsen/logrus"
+)
+
+// MetricMap is used for storing aggregated or consolidated Metric values.
+// The keys of each map are metric names.
+type MetricMap struct {
+	Counters Counters
+	Timers   Timers
+	Gauges   Gauges
+	Sets     Sets
+}
+
+func NewMetricMap() *MetricMap {
+	return &MetricMap{
+		Counters: Counters{},
+		Timers:   Timers{},
+		Gauges:   Gauges{},
+		Sets:     Sets{},
+	}
+}
+
+// Receive adds a single Metric to the MetricMap, and releases the Metric.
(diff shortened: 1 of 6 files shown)


metric_map.go
```diff
@@ -0,0 +1,229 @@
...
+	case COUNTER:
+		mm.receiveCounter(m, tagsKey, nowNano)
+	case GAUGE:
+		mm.receiveGauge(m, tagsKey, nowNano)
+	case TIMER:
+		mm.receiveTimer(m, tagsKey, nowNano)
+	case SET:
+		mm.receiveSet(m, tagsKey, nowNano)
+	default:
+		logrus.StandardLogger().Errorf("Unknown metric type %s for %s", m.Type, m.Name)
```
aelse: Is there sanitisation on m.Type before we reach this point? I don't like the idea of emitting error logs on bad user input (bad statsd clients / unsupported types) but if we should've caught it before here that's another matter.

metric_map.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+	m.Done()
+}
+
+func (mm *MetricMap) Merge(mmFrom *MetricMap) {
+	mmFrom.Counters.Each(func(metricName string, tagsKey string, counterFrom Counter) {
+		v, ok := mm.Counters[metricName]
+		if ok {
```
aelse: consider handling the !ok case first and returning early to avoid all these nested ifs

aelse: and for the functions following this one also

tiedotguy:
I agree there's issues (I've wanted to redo the logic on those `receive*` calls for a long time), but this is a "lift and shift" refactor only, none of this is new code, just new locations for it.

For the sanitization generally, it only happens in tests.  It's one of those "this should panic to make the test fail, but we won't *just in case* it's not a test" situations.

I'll raise an issue for refactoring the `receive*` calls later.

PR #222, opened 2019-02-28
Shipit 45 percentiles

llepper-atlassian:
This isn't quite ready yet, but here's what we have so far!

@tiedotguy FYI


pkg/statsd/aggregator.go
```diff
@@ -161,15 +172,67 @@ func (a *MetricAggregator) Flush(flushInterval time.Duration) {
...
+		return []int{10, 20, 50, 100, 200, 500, 1000, 2000, 5000, 10000}, nil
+	} else {
+		return nil, errors.New("no buckets set algorithm specified")
+	}
+}
+
+func calculatePercentileBuckets(timer gostatsd.Timer, bucketsSet []int) map[int]int {
+	var buckets = append(bucketsSet, InfinityBucketSize)
+
+	var bucketToCountMap = *treemap.NewWithIntComparator()
```
tiedotguy: I'm not a big fan of dragging in the external data structure to hold this.  You know that the timer values will be sorted (or you can make it contractual, either way), which means you can just straight up count the number of values before you reach the next threshold.  This is much more efficient, and removes the ugly `interface{}` based map, and the need to cast everything.

tiedotguy:
Oh - you also need to support all backends except statsdaemon.

pkg/backends/datadog/datadog.go
```diff
@@ -203,6 +206,32 @@ func (d *Client) processMetrics(metrics *gostatsd.MetricMap, cb func(*timeSeries
...
+				val := timer.Buckets[bucket]
+				var bucketMax string
+				if bucket == statsd.InfinityBucketSize {
+					bucketMax = "+Inf"
+				} else {
+					bucketMax = strconv.Itoa(bucket)
+				}
+				bucketTag := "between:" + strconv.Itoa(previousBucket) + "_" + bucketMax
+				newTags := timer.Tags.Concat([]string{bucketTag})
+				fl.addMetricf(counter, float64(val), timer.Hostname, newTags, "%s.buckets", key)
```
llepper-atlassian: @tiedotguy said this should be a gauge. Are we sure that counter is ok?

pkg/backends/datadog/datadog_test.go
```diff
@@ -195,3 +196,79 @@ func metricsOneOfEach() *gostatsd.MetricMap {
...
+			`{"host":"h2","interval":1.1,"metric":"t1.lower","points":[[100,0]],"tags":["tag2"],"type":"gauge"},` +
+			`{"host":"h2","interval":1.1,"metric":"t1.upper","points":[[100,0]],"tags":["tag2"],"type":"gauge"},` +
+			`{"host":"h2","interval":1.1,"metric":"t1.count","points":[[100,0]],"tags":["tag2"],"type":"gauge"},` +
+			`{"host":"h2","interval":1.1,"metric":"t1.count_ps","points":[[100,0]],"tags":["tag2"],"type":"rate"},` +
+			`{"host":"h2","interval":1.1,"metric":"t1.mean","points":[[100,0]],"tags":["tag2"],"type":"gauge"},` +
+			`{"host":"h2","interval":1.1,"metric":"t1.median","points":[[100,0]],"tags":["tag2"],"type":"gauge"},` +
+			`{"host":"h2","interval":1.1,"metric":"t1.std","points":[[100,0]],"tags":["tag2"],"type":"gauge"},` +
+			`{"host":"h2","interval":1.1,"metric":"t1.sum","points":[[100,0]],"tags":["tag2"],"type":"gauge"},` +
+			`{"host":"h2","interval":1.1,"metric":"t1.sum_squares","points":[[100,0]],"tags":["tag2"],"type":"gauge"},` +
+			`{"host":"h2","interval":1.1,"metric":"t1.buckets","points":[[100,5]],"tags":["tag2","between:0_20"],"type":"count"},` +
```
llepper-atlassian: Does anyone know why all the point arrays start with 100?

pkg/backends/datadog/datadog.go
tiedotguy: We can iterate and see what works

pkg/backends/datadog/datadog_test.go
tiedotguy: It's the timestamp:
```
	client.now = func() time.Time {
		return time.Unix(100, 0)
	}
```

pkg/backends/datadog/datadog.go
```diff
@@ -170,6 +172,26 @@ func (d *Client) processMetrics(metrics *gostatsd.MetricMap, cb func(*timeSeries
 	})
 
 	metrics.Timers.Each(func(key, tagsKey string, timer gostatsd.Timer) {
+
+		if timer.Buckets != nil {
```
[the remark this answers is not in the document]
brentthew: How's it look now? Wasn't 100% sure what you meant by this

PR #223, opened 2019-02-28
Propagate TagsKey properly through MetricMap

tiedotguy:


tiedotguy:
This was resulting in metrics with differing tags all landing in the same space, and stomping on each other.

PR #224, opened 2019-03-04
Propagate timestamp and remove v1 ingestion

tiedotguy:
This is some prep work to propagate the timestamp through the system, starting from "receive".  It doesn't cross the http boundary because that's not in the v2 protobuf.

Also nukes v1 ingestion.


metric_map.go
```diff
@@ -191,39 +195,41 @@ func (mm *MetricMap) receiveTimer(m *Metric, tagsKey string, now Nanotime) {
...
+			tagsKey: NewSet(m.Timestamp, map[string]struct{}{m.StringValue: {}}, m.Hostname, m.Tags),
 		}
 	}
 }
 
 func (mm *MetricMap) String() string {
 	buf := new(bytes.Buffer)
 	mm.Counters.Each(func(k, tags string, counter Counter) {
-		fmt.Fprintf(buf, "stats.counter.%s: %d tags=%s\n", k, counter.Value, tags)
+		_, _ = fmt.Fprintf(buf, "stats.counter.%s: %d tags=%s\n", k, counter.Value, tags)
```
aelse: Is this to silence linter warnings?

tiedotguy: Yeeaah.  I updated my goland recently, and now it has a sad about it.

Issue #211, opened 2019-01-29
Send metrics through the pipeline in batches


tiedotguy:
Essentially:
```
DispatchMetric(ctx context.Context, m *Metric)
```
to:
```
DispatchMetric(ctx context.Context, m []*Metric)
```

This should be done before #210 for efficiency.

PR #225, opened 2019-03-06
Pass metrics as a slice (#211)

tiedotguy:


types.go
```diff
@@ -51,7 +51,7 @@ type Runner interface {
 // RawMetricHandler is an interface that accepts a Metric for processing.  Raw refers to pre-aggregation, not
 // pre-consolidation.
 type RawMetricHandler interface {
-	DispatchMetric(ctx context.Context, m *Metric)
+	DispatchMetrics(ctx context.Context, m []*Metric)
```
ash2k: Might be better for GC if this was a slice of metrics, not of pointers. Maybe not, needs testing.

tiedotguy: It possibly would, but with the whole distributed thing, this should hopefully be less critical.  Once everything is stabilized, I may make another optimization sweep.

pkg/statsd/parser.go
```diff
@@ -76,15 +76,20 @@ func (dp *DatagramParser) Run(ctx context.Context) {
 		case <-ctx.Done():
 			return
 		case dgs := <-dp.in:
-			accumM, accumE, accumB := uint64(0), uint64(0), uint64(0)
+			var metrics []*gostatsd.Metric
+
+			accumB, accumE := uint64(0), uint64(0)
 			for _, dg := range dgs {
-				m, e, b := dp.handleDatagram(ctx, dg.Timestamp, dg.IP, dg.Msg)
+				parsedMetrics, e, b := dp.handleDatagram(ctx, dg.Timestamp, dg.IP, dg.Msg)
```
tiedotguy: I'm not loving this, because we're dispatching metrics externally, and events internally.  I'll probably clean that up later.

hligit: Mixed variable styles.
But I agree verbose name is better here. I have to look signature of handleDatagram to know what m, e, b means. And long names are used even in that function.

tiedotguy: Well I do like my verbose names, I'll update that.

PR #226, opened 2019-03-07
Enabling hostname overrides

MovieStoreGuy:


cmd/tester/server.go
```diff
@@ -106,9 +108,8 @@ func (s *Server) Run() error {
 	http.HandleFunc("/start", s.startHandler)
 	http.HandleFunc("/stop", s.stopHandler)
 	http.HandleFunc("/exit", s.exitHandler)
-	http.ListenAndServe(s.WebAddr, nil)
 
-	return nil
+	return http.ListenAndServe(s.WebAddr, nil)
```
MovieStoreGuy: We were ignoring the error returned from the `ListenAndServe`, is this a worthwhile change?

tiedotguy: Try to keep unrelated changes to a minimum. The tester is also a tester, and doesn't need the same level of care.

MovieStoreGuy: Cool, I have reverted it back.

pkg/statsd/statsd_defaults_and_params.go
```diff
@@ -32,6 +32,9 @@ var DefaultTags = gostatsd.Tags{}
 // DefaultInternalTags is the default list of additional tags on internal metrics
 var DefaultInternalTags = gostatsd.Tags{}
 
+// DefaultHostname will be set to the default hostname of the system
+var DefaultHostname = getHost()
```
tiedotguy: This becomes a global thing, which we don't like.  It should be called from `AddFlags`

PR #232, opened 2019-03-28
Fix gometalinter installation

2easy:
go get installation method is no longer supported and fails for
newer golang versions, refer to: https://github.com/alecthomas/gometalinter/issues/552#issuecomment-434266465

New installation method taken from gometalinter README:
https://github.com/alecthomas/gometalinter/blob/master/README.md#binary-releases


aelse:
Thanks for the PR. As gometalinter is deprecated (per https://github.com/alecthomas/gometalinter) we've instead moved to golangci-lint.

PR #240, opened 2019-05-21
Add timeout context to dispatchEvent

juliawong:
Events were not always being sent due to the context being cancelled as it was returned back up to the request. This meant that `dispatchEvent` was not always executed due to the context being cancelled. This PR adds a new timeout context for `internalDispatchEvent`.

### DIFF
--- a/pkg/statsd/handler_backend.go
+++ b/pkg/statsd/handler_backend.go
@@ -194,6 +194,7 @@ func (bh *BackendHandler) DispatchEvent(ctx context.Context, e *gostatsd.Event) 
 		case bh.concurrentEvents <- struct{}{}:
 			// Creates a new context for dispatching the event.
 			// We create a new one otherwise it uses the request context which is cancelled as soon as this function returns.
+			//nolint:govet
 			timeoutCtx, _ := context.WithTimeout(context.Background(), 20*time.Second)
 			go bh.internalDispatchEvent(timeoutCtx, backend, e)
 			eventsDispatched++
--- a/pkg/statsd/handler_backend.go
+++ b/pkg/statsd/handler_backend.go
@@ -192,7 +192,10 @@ func (bh *BackendHandler) DispatchEvent(ctx context.Context, e *gostatsd.Event) 
 			bh.eventWg.Add(eventsDispatched - len(bh.backends))
 			return
 		case bh.concurrentEvents <- struct{}{}:
-			go bh.dispatchEvent(ctx, backend, e)
+			// Creates a new context for dispatching the event.
+			// We create a new one otherwise it uses the request context which is cancelled as soon as this function returns.
+			timeoutCtx, _ := context.WithTimeout(context.Background(), 20*time.Second)
+			go bh.internalDispatchEvent(timeoutCtx, backend, e)
 			eventsDispatched++
 		}
 	}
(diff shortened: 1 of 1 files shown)


pkg/statsd/handler_backend.go
```diff
@@ -192,7 +192,10 @@ func (bh *BackendHandler) DispatchEvent(ctx context.Context, e *gostatsd.Event)
 			bh.eventWg.Add(eventsDispatched - len(bh.backends))
 			return
 		case bh.concurrentEvents <- struct{}{}:
-			go bh.dispatchEvent(ctx, backend, e)
+			// Creates a new context for dispatching the event.
+			// We create a new one otherwise it uses the request context which is cancelled as soon as this function returns.
+			timeoutCtx, _ := context.WithTimeout(context.Background(), 20*time.Second)
```
juliawong: We ditch the CancelFunc because if we defer it here, we still have the same issue that the event may not be sent. We want this func to return so that we don't block the caller.
From https://golang.org/pkg/context/#pkg-overview
>  Failing to call the CancelFunc leaks the child and its children until the parent is canceled or the timer fires.

tiedotguy: I think we should be explicit - we can use a pattern like:
```
go func() {
    bh.internalDispatchEvent(timeoutCtx, backend, e)
    cancel()
}()
```

... or, pass the cancel to `internalDispatchEvent` ...

```
func (bh *BackendHandler) internalDispatchEvent(ctx context.Context, backend gostatsd.Backend, e *gostatsd.Event, cancel func()) {
  defer cancel()
...
}
```

The former is a bit cleaner from "who is responsible", but the latter is neater, because the `go func() {}()` syntax is a bit wordy.  Either will do.

ash2k:
```go
go func() {
    timeoutCtx, cancelTimeout := context.WithTimeout(context.Background(), 20*time.Second)
    defer cancelTimeout()
    bh.internalDispatchEvent(timeoutCtx, backend, e)
}()
```

pkg/statsd/handler_backend.go
```diff
@@ -194,10 +194,12 @@ func (bh *BackendHandler) DispatchEvent(ctx context.Context, e *gostatsd.Event)
...
 			// We create a new one otherwise it uses the request context which is cancelled as soon as this function returns.
-			//nolint:govet
-			timeoutCtx, _ := context.WithTimeout(context.Background(), 20*time.Second)
-			go bh.internalDispatchEvent(timeoutCtx, backend, e)
-			eventsDispatched++
+			go func() {
+				timeoutCtx, cancelTimeout := context.WithTimeout(context.Background(), 20*time.Second)
+				defer cancelTimeout()
+				bh.internalDispatchEvent(timeoutCtx, backend, e)
+				eventsDispatched++
```
tiedotguy: This is now a race condition, `eventsDispatched` needs to be outside the go func.  `eventsDispatched` is a variable from the outer scope, being captured by the func (ie, it's storing a reference to it).  This means if there's multiple backends setup, then two goroutines could both try to update it, without any kind of synchronization.

I also just realized there's also a similar hazard with `backend`, except from the other direction.  The loop modifies `backend`.  The function will always be referring to the outer variable, not its own copy.  You can see the behavior at https://play.golang.org/p/g6jPXwxeXw0

Issue #233, opened 2019-04-04
Graphite tags support


TomMizrachi:
I'm trying to understand if you guys support the new graphite tagging system introduced in version 1.1.x

After going through the repo's intro doc I found out you guys don't support the graphite tagging format, as can be viewed here:
https://grafana.com/blog/2018/01/11/graphite-1.1-teaching-an-old-dog-new-tricks/

I would like to know if there are any plans for gostatsd to support graphite tags

tiedotguy:
Hi

There was some discussion on #213 - my rough plan was to drop tags entirely in "graphite mode" (with a warning on startup), and there was some concern that there may be some explosion of file handles if we include tags.  The big problem is I don't know the graphite system well enough to assess the potential impact.

TomMizrachi:
Can you please elaborate what about this concern? I don't see how adding tags might cause issues, since its just as adding tags for other endpoints.

tiedotguy:
As I understand it, every unique metrics (ie, metric name + tagset) requires an open file handle, so cardinality can starve the system.

I don't know if this is a historical thing or not though, so it may well be a non-issue now.

TomMizrachi:
I dont see how it has anything to do with gostatsd.
It is true that every unique metric is resembled as a file (whisper) in graphite, but the tag system does not create a whisper per tag - the tags are being stored in a different database, which is being queried when needed.

Here is a quote from the link I send in my first message:
```
At its core, the tagging system is implemented as a tag database (TagDB) alongside the metrics that allows them to be efficiently queried by individual tag values rather than having to traverse the metrics tree looking for series that match the specified query. Internally the tag index is stored in one of a number of pluggable tag databases, currently supported options are the internal graphite-web database, redis, or an external system that implements the Graphite tagging HTTP API. Carbon automatically keeps the index up to date with any tagged series seen.
```

tiedotguy:
Ahh, sorry - I misunderstood this.  In the prior discussion, there was concern that baking tags in to the metric name could cause problems with lots of file handles (I don't know if it's a issue or not with modern Graphite).

I'll have a shot at adding it over the weekend, it would be nice to cleanup some of the outstanding Graphite issues :)

tiedotguy:
@TomMizrachi - done! Please do raise a new issue if there's any problems - I don't have a good setup to really test it at scale.

Issue #213, opened 2019-01-29
Graphite backend problem when using same metric-name from multiple client-processes


RasmusWL:
We're encountering an error in the graphite backend. 

Whenever we redeploy a service, the values of our statsd timers go to 0 in graphite for a few minutes. The services changes hosts, and therefore also IP. We run >1 instances of our service, but they all report stats to the same metric name `my_service.my_metric`.

By using tcpdump on the statsd server, we could see the data that gostatsd sends to graphite. We would see something like the following just after a redeploy:

```
stats.timers.my_service.my_metric.time.count 76 1548758589
...
stats.timers.my_service.my_metric.time.count 0 1548758589
...
stats.timers.my_service.my_metric.time.count 0 1548758590
```

So because graphite [only stores the latest value in a given timeslot](https://www.michael-noll.com/blog/2013/11/06/sending-metrics-from-storm-to-graphite/#understanding-how-graphite-handles-incoming-data) graphite will end up only storing 0 :(

So it seems like some thread/goroutine is still expecting to handle `my_service.my_metric` from the old instances that are now dead and therefore keeps on sending zeroes, while a new thread/goroutine is handling the new instaces giving us the real values.

Looking at the metrics in "stable" times with 2 instances of the service running, `stats.timers.my_service.my_metric.time.count` is roughly 50% of what we would expect (also compared to our logs). So it seems like this is in fact a more general problem when we send stats to the same metric name from multiple instances.

---

We're currently using version [`9.0.0`](https://github.com/atlassian/gostatsd/releases/tag/9.0.0). I looked over [the changes since then](https://github.com/atlassian/gostatsd/compare/9.0.0...master) and it didn't look like any of these would have an influence on our problem.

tiedotguy:
This is because tags (and hostname) aren't propagated through to Graphite - see #78.  In Gostatsd it's tracking (roughly):
```
[
{name=my_service.my_metric, hostname=server1, tags=[], value=x},
{name=my_service.my_metric, hostname=server2, tags=[], value=y}
]
```

When this is rendered to Graphite format, Gostatsd doesn't include hostname or tags, resulting in the same metric name written twice at the same timestamp.  When you deploy, you're seeing persistence from the old server (`expiry-interval` option, defaults to `5m`, unfortunately not well documented).  With no new metrics coming in from the old server, Gostatsd will continue to emit 0s until the metric expires.  Due to no tagging, that will conflict with the new server.

[...]

Issue #78, opened 2016-11-01
Graphite backend ignores tags


glennslaven:
The functions inside `preparePayload` in `/backend/backends/graphite/graphite.go` accept the `tagsKey` parameter and each metric is prepared based on its tag. However the tag is never used in constructing the metric name to send to graphite.

So what happens is that if you have metric key `a.b.c` that is broken up into 10 tags, the metric `a.b.c` is sent to graphite 10 times with the same name & timestamp, but with the various values for each of the tags.  Then graphite simply takes the last one to arrive and put that in as the value for that timestamp, the other values are lost.

Ideally if there is a tag, the `preparePayload` function would insert it as (probably) a dot suffix after the key name.

ash2k:
@glennslaven thanks for reporting this, we should probably add a configuration option to tweak this. However note that this is kind of related to the other issue you created (#77) - how timers are going to work? It is not clear.

glennslaven:
@ash2k to be honest I'm not sure , I just assumed the timers were supposed to be fragmented by the tags :)

I suppose the other alternative is to keep the timers in one bucket and duplicate the data across the tags, although at that point the tags seem kind of useless.

ivan-section-io:
Tags are part of the aggregation mechanism and should be output if provided.
The key to #77 (for graphite) is to not include a tag in the first place.
I think the only question is where tags belong:, prepended, appended, or even just before the derived stat name.
Including the tag (in some fashion) is the only safe option.
I'm sure a PR will be coming from @glennslaven shortly.

tiedotguy:
@glennslaven, I've added 1.1 tagging style support.  I realize that your original request was more about node style naming rather than actual tags (because the issue was raised a year before tagging was added), but hopefully this is a suitable alternative.  Please raise another issue if you specifically want node style.

PR #242, opened 2019-06-07
Adds support for graphite tags

### LINKED ISSUE #161: Flaky test
```
--- FAIL: TestSendMetricsAsync (0.01s)
	Error Trace:	graphite_test.go:125
			asm_amd64.s:2337
	Error:      	Received unexpected error:
	            	accept tcp 127.0.0.1:32810: use of closed network connection
```
https://travis-ci.org/atlassian/gostatsd/builds/354706831#L874

tiedotguy:
This does a few things:
- Adds support for graphite tags (including a bunch of refactor work), however only using actual tags (#233, #213), and not node style (which is probably what was intended with #78).  I'm calling all 3 resolved by this, however.  If someone actually wants node style tags, a fresh issue should be opened.
- Pulls backend configuration to its own documentation file, because it's getting noisy.
- Fixed #161

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ and again if the dependencies change.  A [protobuf](https://github.com/protocolb
 directory.  Managing this in a platform agnostic way is difficult, but PRs are welcome. Hopefully it will be sufficient to use the generated protobuf
 files in the majority of cases.
 
-If you are unable to build `gostatsd` please try running `make setup` again before reporting a bug.
+If you are unable to build `gostatsd` please check your go version, and try running `make setup` again before reporting a bug.
 
 Running the server
 ------------------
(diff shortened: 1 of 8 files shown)


pkg/backends/graphite/graphite_test.go
```diff
@@ -135,7 +145,6 @@ func TestSendMetricsAsync(t *testing.T) {
 		}
 	}()
 	defer acceptWg.Wait()
-	defer l.Close()
```
tiedotguy: This is #161, this is:
- duplicate a line from above
- executing before `acceptWg.Wait()`, leading to the socket being closed before the goroutine has accepted from it

ash2k: I think the idea here was to close the socket and unblock the accept goroutine if something goes wrong in the main one. It should be ok to close multiple times, I think. Anyway, I'm not saying this is correct code and doesn't have bugs. Just trying to recollect why it is like that and I may be misremembering. Not opposed to making this change at all.

tiedotguy: Makes sense.  I can re-factor it if we start seeing failures again for sure, but this was close enough.

README.md
```diff
@@ -119,81 +119,18 @@ There is no capability to run an https server at this point in time, and no auth
...
-- Directly to the Insights Collector - [Insights Event API](https://docs.newrelic.com/docs/insights/insights-data-sources/custom-data/send-custom-events-event-api)
-- Via the Infrastructure Agent's inbuilt HTTP Server
-
-### [New Relic Insights Event API](https://docs.newrelic.com/docs/insights/insights-data-sources/custom-data/send-custom-events-event-api)
-Sending directly to the Event API alleviates the requirement of needing to have the New Relic Infrastructure Agent. Therefore you can run this from nearly anywhere for maximum flexibility. This also becomes a shorter data path with less resource requirements becoming a simpler setup.
-
-To use this method, create an Insert API Key from here: https://insights.newrelic.com/accounts/YOUR_ACCOUNT_ID/manage/api_keys
+Configuring backends
+--------------------
+Refer to [BACKENDS.md] for configuration options for the backends
```
juliawong: Is this link meant to be clickable? Because when previewing it I see the text `[BACKENDS.md]`

tiedotguy: I'm not too sure, actually.  It's the style used throughout the file when referencing local files, so I just followed it.

PR #245, opened 2019-06-19
Latency Histograms

tmszdmsk:
Implementing latency histograms, based on previous work in #222.

Latency histograms can be enabled with "configuration tag" `gsd_histogram` where its value says what thresholds should be used to output counter metrics.

E.g. timer `t1` with tag `gsd_histogram:2.5_20_100_2000` will output additional 5 cumulative counters
* `t1.histogram` with tag `le:2.5`
* `t1.histogram` with tag `le:20`
* `t1.histogram` with tag `le:100`
* `t1.histogram` with tag `le:2000`
* `t1.histogram` with tag `le:+Inf`

where `le` means `less or equal` and follows [Prometheus convention](https://prometheus.io/docs/practices/histograms/)

counter values will be equal to the number of data points with values `<=` configured histogram thresholds.
`le:+Inf` will naturally contain total number of data points as all values are `<= +Inf`


pkg/statsd/latency_histogram_test.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"math"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/atlassian/gostatsd"
+)
+
+func Timer(histogramThresholds string, values ...float64) gostatsd.Timer {
```
tiedotguy: Shouldn't be exported.

timers.go
```diff
@@ -19,6 +19,14 @@ type Timer struct {
 	Timestamp    Nanotime    // Last time value was updated
 	Hostname     string      // Hostname of the source of the metric
 	Tags         Tags        // The tags for the timer
+
+	// Map bounds to count of measures seen in that bucket.
+	// This map only non-empty if the metric specifies percentile aggregation in its tags.
+	Histogram map[HistogramThreshold]int
+}
+
+type HistogramThreshold struct {
```
tiedotguy: What's the value in having a struct instead of just a float?

pkg/statsd/aggregator.go
```diff
@@ -170,6 +170,14 @@ func (a *MetricAggregator) Flush(flushInterval time.Duration) {
 			timer.PerSecond = 0
 		}
 	})
+
+	a.metricMap.Timers.Each(func(key, tagsKey string, timer gostatsd.Timer) {
+		histogram := LatencyHistogram(timer)
```
tiedotguy: This should calculate histograms, or aggregated functions (lines 78-169), not both.

glide.yaml
```diff
@@ -30,3 +30,6 @@ import:
   version: ^1.2.0
 - package: github.com/tilinna/clock
   version: 1.0.0
+- package: github.com/emirpasic/gods/maps/treemap
```
tiedotguy: I believe all the imports can be reset, they're from earlier in the branch.

pkg/backends/datadog/datadog_test.go
```diff
@@ -196,3 +196,74 @@ func metricsOneOfEach() *gostatsd.MetricMap {
...
+
+func metricsWithHistogram() *gostatsd.MetricMap {
+	return &gostatsd.MetricMap{
+		Timers: gostatsd.Timers{
+			"t1": map[string]gostatsd.Timer{
+				"tag2": {
+					Values:    []float64{0, 1},
+					Timestamp: gostatsd.Nanotime(200),
+					Hostname:  "h2",
+					Tags:      gostatsd.Tags{"tag2"},
```
tiedotguy: These aren't the tags that produce this histogram.

pkg/backends/cloudwatch/cloudwatch.go
```diff
@@ -144,6 +146,19 @@ func (client Client) buildMetricData(metrics *gostatsd.MetricMap) (metricData []
...
 	})
 
+	metrics.Timers.Each(func(key, tagsKey string, timer gostatsd.Timer) {
+		if timer.Histogram != nil {
+			for histogramThreshold, count := range timer.Histogram {
+				bucketTag := "le:+Inf"
+				if !math.IsInf(histogramThreshold.Le, 1) {
+					bucketTag = "le:" + strconv.FormatFloat(histogramThreshold.Le, 'f', -1, 64)
+				}
+				newTags := timer.Tags.Concat([]string{bucketTag})
```
tiedotguy: `[]string` should be `gostatsd.Tags` (all throughout other backends too)

pkg/statsd/latency_histogram.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"github.com/atlassian/gostatsd"
+)
+
+const (
+	HistogramThresholdsTagPrefix = "gsd_histogram:"
+	HistogramThresholdsSeparator = "_"
+)
+
+func LatencyHistogram(timer gostatsd.Timer) map[gostatsd.HistogramThreshold]int {
```
tiedotguy: Doesn't need to be exported.

timers.go
tmszdmsk: probably none, will get rid of that

tmszdmsk: wanted to use
```golang
type HistogramThreshold float64
```
because I think it helps with readability, but then I can't compare `float64`s with it. Any tip on that?

pkg/statsd/latency_histogram.go
tmszdmsk: I lack golang knowledge, how can I use it outside in `aggregator` then?

pkg/backends/datadog/datadog.go
```diff
@@ -203,6 +204,20 @@ func (d *Client) processMetrics(metrics *gostatsd.MetricMap, cb func(*timeSeries
...
 
+	metrics.Timers.Each(func(key, tagsKey string, timer gostatsd.Timer) {
+		if timer.Histogram != nil {
+			for histogramThreshold, count := range timer.Histogram {
+				bucketTag := "le:+Inf"
+				if !math.IsInf(histogramThreshold.Le, 1) {
+					bucketTag = "le:" + strconv.FormatFloat(histogramThreshold.Le, 'f', -1, 64)
+				}
+				newTags := timer.Tags.Concat([]string{bucketTag})
+				fl.addMetricf(counter, float64(count), timer.Hostname, newTags, "%s.histogram", key)
```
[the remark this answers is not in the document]
tmszdmsk: I think I don't understand all the implications of introducing `counter` to `datadog` backend. It was implemented like that in #222. What's the benefit of having that? I don't really want to make a risky change if it's not really needed.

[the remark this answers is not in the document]
tiedotguy: The hazard is there's a time component to both a rate and a count, but not a gauge.

It's the number of things that occurred in (for example) 10 seconds.  If you miss a flush interval, then you collect 20 seconds of data, but you're telling upstream it's 10 seconds worth.  If you know the window you can go from counter to rate (or vice versa), but once the data leaves gostatsd, you don't know the window anymore.

It's not directly a risk (because it's a new thing), but rather things may not behave as expected when querying.  You can't test it in gostatsd, because it's how the backend handles the query.  Then if it turns out you then need to change it, **that's** a breaking change.

This is why I'm advising to test it end to end, with the real pipeline, with both happy and sad conditions :)

timers.go
tiedotguy: `x < HistogramThreshold(y)` or `float64(x) < y` should work.

tmszdmsk:
@tiedotguy @adampetrovic @MovieStoreGuy 

 I am thinking about configuration options & restrictions around histograms and I am not sure what do we need, here are my thoughts:
1. configuration to enable/disable support for `gsd_histogram:`, enabled by default
1. being able to specify the maximum number of histogram buckets (what should be the default? 10?)
1. I am not really convinced that we should do aggregations XOR histograms. It can slow down the adoption because you will have to sacrifice something upfront. Maybe it should be 2 buckets "for free" + every next bucket eliminates one aggregation, in some specific order. But that's probably too intricate. What do you think? 
1. What other configuration options?

tiedotguy:
For the configuration, rather than two options for enable/disable and count, I would have it as a single configuration setting, where 0 is disabled (with special logic to check for 0), and a default of `math.MaxInt32` - effectively unlimited, but practically bound by the UDP datagram size.

They should absolutely be distinct.  You don't sacrifice anything - you make a choice to use it or not.  If you want both, you send twice.  If they're not distinct then there's no way to opt out of the wasteful additional metrics.  If I was running the deployment, I would look carefully at the cardinality impact, then set the maximum buckets to 2, one of which is +Inf.

To put it bluntly, there's a lot of bad tagging that goes on, and this is a chance to clean house.  I hope for two things to come out of this work:
1. because it will give better data, people will want to adopt it (aka the carrot)
2. because it's new, strict controls can be put in place from day 0 to ensure only good data is accepted (aka the stick)

On a more technical level, this is something I may re-factor at some point to change from a timer to a new type (histogram) when the tag is present, because there's some efficiency gains to be made by pushing the work to a local process.  This would be internal only - the statsd interface would be unchanged.

I don't think there's any other configuration options needed.  From memory, in the original branch it was using "named sets", ie `gsd_histogram:pow2`, and the administrator had to define the sets ahead of time, but this approach bypasses it.

tmszdmsk:
I thought more about operational backward-compatibility, not binary, but if it's being used as a library then it makes sense. I'll bump that.

Should I add support for `nil`able values in `Timer` or should the `if` be replicated in all the backends? (The  one seems cleaner ;)

Also, I am gonna pause work on this for now as obs needs to decide whether to proceed with this change or not. Thanks, Harold!

tiedotguy:
No problem, thanks for working on it - it's definitely very close to ready if they decide to proceed.  There was something I remembered today that needs changing, but now I can't recall it.  I'll give the whole thing a fresh review if it proceeds.

Not certain what you mean by `nil`lable values in `Timer` though?

pkg/statsd/latency_histogram_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+func timer(histogramThresholds string, values ...float64) gostatsd.Timer {
+	return gostatsd.Timer{
+		Values:    values,
+		Tags:      []string{HistogramThresholdsTagPrefix + histogramThresholds},
+		Histogram: map[gostatsd.HistogramThreshold]int{},
+	}
+}
+
+func TestLatencyHistogram(t *testing.T) {
```
brentthew: What about negative thresholds, and negative timer values?

tmszdmsk: those are "latency histograms" and in nature, they should have only non-negative values. Apart from that:
* it should work
* It makes sense to add those "unnatural" test cases. :thumbsup:

tiedotguy: Really they're just numbers - latency is a number, but a number isn't latency.  The problem stems from the name `timer`, it pre-supposes the meaning of the number.  It was raised years ago in https://github.com/statsd/statsd/issues/98

brentthew: They're documented (in the readme, and any code except this file) as just histograms. 

Which I guess raises the question - what's special about them that makes them latency histograms?

tmszdmsk: I am not that opinionated on that one. At the moment the histograms are only generated for "timers". I have never used timers for "non-timing" purposes and this nomenclature is quite widely spread, but I see how timers can be repurposed for different scenarios.

I will make a change to get rid of "latency" bit from the code and add more tests to cover that case

tiedotguy: They're also generated for the histogram type - histogram and timer are a literal alias (https://github.com/atlassian/gostatsd/blob/master/pkg/statsd/lexer.go#L368-L379)

Near as I can tell, this alias is the defacto standard across all implementations, except the original etsy who IIRC don't support the `h` type at all.

I don't think it's documented in the `README.md` though, only in `CHANGELOG.md` when I added it.

tmszdmsk:
@tiedotguy 
> Not certain what you mean by nillable values in Timer though?

There is no way to set default aggregations (`timer.Count`, `timer.Sum`, ...) to "no value" (`nil`) so that they are not outputted in the backends. I don't want to add knowledge about `gsd_histogram` meta tag to the backends.
My idea is to change the definition of `Timer` to
```go
type Timer struct {
	Count        *int         // The number of timers in the series
	SampledCount *float64     // Number of timings received, divided by sampling rate
        ...
```
so that we can assign `nil`s when histogram is enabled and check that in the backends before outputting.

tiedotguy:
Ahh, understood.  I believe each backend should just check for `Timer.Histogram == nil`.  It might be worth pulling each path (regular timer + histogram) out to a function, just to clean things up though, but I won't hold you to it.

I'm also considering adding Datadog distributions through the timer statsd type (with a tag), so at some point in the future I may add a `Timer.Type` (regular/histogram/distribution), or possibly make the tags route to more concrete types (ie, timer type + `gsd_histogram` becomes a `gostatsd.Histogram` instead of a `gostatsd.Timer`).

For now, I think it's ok not to do any big refactor to support it.

README.md
```diff
@@ -186,7 +186,18 @@ upper-pct=false
 
 By default (for compatibility), they are all false and the metrics will be emitted.
 
+Timer histograms
+----------------
 
+Timer histograms inspired by [Promehteus implementaion](https://prometheus.io/docs/concepts/metric_types/#histogram) can be
```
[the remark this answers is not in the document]
aelse: also 'implementation'

tiedotguy:
Also if you like, next time I'm in the office I can whiteboard why I'm paranoid about the counter vs gauge vs rate thing.

MovieStoreGuy:
This branch is lagging behind master a lot, is this something we are going to keep going with?

tmszdmsk:
> This branch is lagging behind master a lot, is this something we are going to keep going with?

Yes, I have some changes on my localhost, but wanted to address all the comments first

tmszdmsk:
master branch is merged & I think all comments are resolved, please @aelse, @tiedotguy review once more.

tmszdmsk:
Also, how can I test it with one of my projects in e.g. staging?

tiedotguy:
Thanks! I'll try and give it a thorough review over the weekend. For testing, you'll need to discuss with the team, because it's the aggregation tier that's changing, not the on-host tier.

pkg/statsd/aggregator.go
```diff
@@ -76,99 +78,102 @@ func (a *MetricAggregator) Flush(flushInterval time.Duration) {
 	})
 
 	a.metricMap.Timers.Each(func(key, tagsKey string, timer gostatsd.Timer) {
-		if count := len(timer.Values); count > 0 {
-			sort.Float64s(timer.Values)
-			timer.Min = timer.Values[0]
-			timer.Max = timer.Values[count-1]
-			n := len(timer.Values)
-			count := float64(n)
+		if hasHistogramTag(timer) {
```
tiedotguy: As pointed out in another comment, this can be restructured a bit to make the diff a lot cleaner (I think the suggested diff might be broken, sorry if it's not clear):
```suggestion
		if hasHistogramTag(timer) {
			timer.Histogram = latencyHistogram(timer, a.histogramLimit)
			a.metricMap.Timers[key][tagsKey] = timer // line needs to be done on both sides of the condition, because it's not a pointer
			return
		}
		... original code with original indenting ...
		a.metricMap.Timers[key][tagsKey] = timer // line needs to be done on both sides of the condition, because it's not a pointer
```

There's a few other places where similar could be done, but those diffs are fairly clean compared to this one (I think the line breaks confuse githubs diff algo), so only this one really needs it.

pkg/statsd/latency_histogram.go
```diff
@@ -0,0 +1,105 @@
...
+import (
+	"math"
+	"strconv"
+	"strings"
+
+	"github.com/atlassian/gostatsd"
+)
+
+const (
+	HistogramThresholdsTagPrefix = "gsd_histogram:"
```
tiedotguy:
```suggestion
	histogramThresholdsTagPrefix = "gsd_histogram:"
```

pkg/statsd/latency_histogram.go
```diff
@@ -0,0 +1,105 @@
...
+	HistogramThresholdsSeparator = "_"
```
tiedotguy:
```suggestion
	histogramThresholdsSeparator = "_"
```

pkg/statsd/latency_histogram.go
```diff
@@ -0,0 +1,105 @@
...
+	result[infiniteThreshold] = 0
+
+	return result
+}
+
+func retrieveThresholds(timer gostatsd.Timer, bucketlimit uint32) []gostatsd.HistogramThreshold {
+	tag, found := findTag(timer.Tags, HistogramThresholdsTagPrefix)
+	if found {
+		bucketsTagValue := tag[len(HistogramThresholdsTagPrefix):]
+		stringThresholds := strings.Split(bucketsTagValue, HistogramThresholdsSeparator)
```
tiedotguy:
```suggestion
		stringThresholds := strings.Split(bucketsTagValue, histogramThresholdsSeparator)
```

pkg/statsd/latency_histogram.go
```diff
@@ -0,0 +1,105 @@
...
+
+	result[infiniteThreshold] = 0
+
+	return result
+}
+
+func retrieveThresholds(timer gostatsd.Timer, bucketlimit uint32) []gostatsd.HistogramThreshold {
+	tag, found := findTag(timer.Tags, HistogramThresholdsTagPrefix)
+	if found {
+		bucketsTagValue := tag[len(HistogramThresholdsTagPrefix):]
```
tiedotguy:
```suggestion
		bucketsTagValue := tag[len(histogramThresholdsTagPrefix):]
```

pkg/statsd/latency_histogram.go
```diff
@@ -0,0 +1,105 @@
...
+		result[histogramThreshold] = 0
+	}
+
+	result[infiniteThreshold] = 0
+
+	return result
+}
+
+func retrieveThresholds(timer gostatsd.Timer, bucketlimit uint32) []gostatsd.HistogramThreshold {
+	tag, found := findTag(timer.Tags, HistogramThresholdsTagPrefix)
```
tiedotguy:
```suggestion
	tag, found := findTag(timer.Tags, histogramThresholdsTagPrefix)
```

pkg/statsd/latency_histogram.go
```diff
@@ -0,0 +1,105 @@
...
+		floatBucket, err := strconv.ParseFloat(v, 64)
+		if err == nil {
+			lb = append(lb, gostatsd.HistogramThreshold(floatBucket))
+		}
+	}
+	return lb
+}
+
+func hasHistogramTag(timer gostatsd.Timer) bool {
+	_, found := findTag(timer.Tags, HistogramThresholdsTagPrefix)
```
tiedotguy:
```suggestion
	_, found := findTag(timer.Tags, histogramThresholdsTagPrefix)
```

pkg/statsd/latency_histogram_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/atlassian/gostatsd"
+)
+
+func timer(histogramThresholds string, values ...float64) gostatsd.Timer {
+	return gostatsd.Timer{
+		Values:    values,
+		Tags:      []string{HistogramThresholdsTagPrefix + histogramThresholds},
```
tiedotguy:
```suggestion
		Tags:      []string{histogramThresholdsTagPrefix + histogramThresholds},
```

README.md
```diff
@@ -184,7 +184,40 @@ upper-pct=false
 
 By default (for compatibility), they are all false and the metrics will be emitted.
 
+Timer histograms (experimental feature)
+----------------
 
+Timer histograms inspired by [Prometheus implementation](https://prometheus.io/docs/concepts/metric_types/#histogram) can be
+enabled on a per time series basis using `gsd_histogram` meta tag with value containing histogram bucketing definition (joined with `_`)
```
tiedotguy:
```suggestion
enabled on a per time series basis using `gsd_histogram` meta tag with value containing histogram bucketing definition (joined with `_`)
```

README.md
```diff
@@ -184,7 +184,40 @@ upper-pct=false
...
+* `<base>.histogram` with tag `le:25`
+* `<base>.histogram` with tag `le:50`
+* `<base>.histogram` with tag `le:+Inf`
+
+Each time series will contain a total number of timer data points that had a value less or equal `le` value, e.g. counter `<base>.histogram` with the tag `le:5` will contain the number of all observations that had a value not bigger than `5`.
+Counter `<base>.histogram` with tag `le:+Inf` is equivalent to `<base>.count` and contains the total number.
+
+All original timer tags are preserved and added to all the time series.
+
+To limit cardinality, `timer-histogram-limit` option can be specified to limit the number of buckets that will be created (default is `math.MaxUint32`).
```
tiedotguy:
```suggestion
To limit cardinality, `timer-histogram-limit` option can be specified to limit the number of buckets that will be created (default is `math.MaxUint32`).
```

build/.dockerignore
```diff
@@ -1,2 +0,0 @@
-*
```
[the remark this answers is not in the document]
tmszdmsk: that was an accident, I expect that `./build/` dir can be safely deleted.

adampetrovic:
FYI i'll be testing this branch out in an innovation week this week. Will report back any issues I find.

pkg/backends/cloudwatch/cloudwatch.go
```diff
@@ -142,6 +144,16 @@ func (client Client) buildMetricData(metrics *gostatsd.MetricMap) (metricData []
 		for _, pct := range timer.Percentiles {
 			addMetricData(key+"."+pct.Str, "Milliseconds", pct.Float, timer.Tags)
 		}
+		if timer.Histogram != nil {
```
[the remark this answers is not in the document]
adampetrovic: @tmszdmsk if we wanted to enable the ability to send both the histogram and the original timer metrics, would it be as simple as changing this logic slightly to drop the histogram buckets tag for those timer metrics?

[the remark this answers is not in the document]
tiedotguy: Sending both through would be bad, because it makes it impossible to not send both. If both are desired (for example, doing side by side testing) then the client should be sending both - ideally with different names so as not to pollute the namespace in the event that the wrong tags make it out to a test gostatsd deployment.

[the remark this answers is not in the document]
adampetrovic: Could we control that behaviour with a configuration flag? The UX isn't really that great if we're asking users to duplicate code for sending metrics anywhere they want a histogram.

> ideally with different names so as not to pollute the namespace in the event that the wrong tags make it out to a test gostatsd deployment

wouldn't this be a non-issue given we append `.histogram` to those metrics already?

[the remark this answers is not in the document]
tiedotguy: I haven't looked at the code for a while, yeah - if we're adding a suffix, pollution isn't an issue.

The primary issue really is the inability to not send both. Lack of control in what is emitted is a large cause of high cardinality (ie, timer aggregations).

I'd personally make the argument that work is required to move to the new style anyway, both in what is emitted (adding buckets) and querying the data, so there's going to be code change regardless.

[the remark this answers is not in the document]
adampetrovic: Alright, that makes more sense. We'll leave it as is.

tiedotguy:
Sounds good to me. Sync up the README.md and cut a release. It's clearly listed as experimental, so if breaking changes are needed, that's also groovy. Looking at the requested changes I had historically they appear to be done.

tiedotguy:
Also, let's make sure this has a squash merge, it's a little noisy :)

tiedotguy:
:itshappening:

tmszdmsk:
The future is now! :D Thanks @adampetrovic for pushing that through. 😃

adampetrovic:
@tiedotguy in terms of performance, I just wanted to double check you are happy with the implementation. How would you recommend we best test / benchmark the performance?

tiedotguy:
I didn't see anything obviously bad - even if there's some garbage created, it's important to remember that work occurs on flush, which is (ideally) ~10s, rather than per datagram/metric.

As a hypothetical, if everyone switched immediately, I would actually expect this to be slightly better performance, because IIRC the sorting was enough that it showed up on the flamegraphs, and that's not there in the histogram path.

Practically, I am unsure of a good approach. We never got a reliable way to tee data pre-aggregation. In theory you you could put a proxy in to tee the data (one day that'll be built in... when I finish the transport rewrite... one day...), then you could manipulate timers, add an arbitrary number of buckets, and forward it on to a regular aggregation tier that's configured with a stdout backend.  That's a non-trivial amount of work though.

I'd suggest keeping an eye on the network usage from forwarder->aggregation tier too. The raw length of tags is relatively high now, although they should compress well.

Not sure if that's super useful or not.

PR #250, opened 2019-07-19
Add support for events via New Relic backend

Kav91:
When the event is sent through, it is sent as a full New Relic event, so all context is maintained and queryable.

eg.
image

### DIFF
--- a/pkg/backends/newrelic/newrelic.go
+++ b/pkg/backends/newrelic/newrelic.go
@@ -2,7 +2,7 @@ package newrelic
 
 import (
 	"bytes"
-	"compress/zlib"
+	"compress/gzip"
 	"context"
 	"crypto/tls"
 	"encoding/json"
@@ -12,6 +12,7 @@ import (
 	"net"
 	"net/http"
 	"runtime"
+	"strconv"
 	"strings"
 	"sync/atomic"
 	"time"
@@ -28,7 +29,7 @@ const (
 	// BackendName is the name of this backend.
 	BackendName                  = "newrelic"
 	integrationName              = "com.newrelic.gostatsd"
-	integrationVersion           = "2.1.0"
+	integrationVersion           = "2.2.0"
 	protocolVersion              = "2"
 	defaultUserAgent             = "gostatsd"
 	defaultMaxRequestElapsedTime = 15 * time.Second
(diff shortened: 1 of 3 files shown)


pkg/backends/newrelic/newrelic.go
```diff
@@ -2,7 +2,7 @@ package newrelic
 
 import (
 	"bytes"
-	"compress/zlib"
+	"compress/gzip"
```
MovieStoreGuy: What was the reason for changing the compression library here?

pkg/backends/newrelic/newrelic.go
```diff
@@ -486,6 +539,22 @@ func newInfraPayload(data interface{}) NewRelicInfraPayload {
 	}
 }
 
+func (n *Client) setTags(tags gostatsd.Tags, data *map[string]interface{}) {
```
MovieStoreGuy: You don't need to pass a pointer to a map since a map object itself is already a reference to a memory block.

pkg/backends/newrelic/newrelic_test.go
```diff
@@ -186,3 +187,43 @@ func metricsOneOfEach() *gostatsd.MetricMap {
...
+	gostatsdEvent := gostatsd.Event{Title: "EventTitle", Text: "hi", Hostname: "blah", Priority: 1}
+
+	formattedEvent := client.EventFormatter(&gostatsdEvent)
+	fevent, err := json.Marshal(formattedEvent)
+	require.NoError(t, err)
+
+	expected := `{"name":"com.newrelic.gostatsd","protocol_version":"2","integration_version":"2.2.0","data":` +
+		`[{"metrics":[{"AggregationKey":"","AlertType":"","DateHappened":0,"Hostname":"blah","Priority":"low","SourceTypeName":"","Text":"hi","Title":"EventTitle","event_type":"GoStatsD","name":"event"}]}]}`
+
+	if string(fevent) != expected {
```
MovieStoreGuy: This should also be a `require` statement

pkg/backends/newrelic/newrelic.go
```diff
@@ -257,6 +296,12 @@ func (n *Client) constructPost(ctx context.Context, buffer *bytes.Buffer, data i
 		return nil, fmt.Errorf("[%s] unable to marshal: %v", BackendName, mErr)
 	}
 
+	return n.postWrapper(ctx, mJSON)
+}
+
+// postWrapper compresses JSON for Insights
+func (n *Client) postWrapper(ctx context.Context, mJSON []byte) (func() error, error) {
+	json := mJSON
```
MovieStoreGuy: Not sure if this line is needed since both will point to same block of memory.

Kav91:
Thanks @MovieStoreGuy updated all per your comments.

Kav91:
@MovieStoreGuy @tiedotguy hey guys just checking in, anything else needed? 
thx

tiedotguy:
Seems reasonable to me.  I'll make a release.

PR #252, opened 2019-08-21
Allow disabling of retries

MovieStoreGuy:
Making sure we have the configurability to disable retires.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,8 @@ following configuration options:
 - `enable-http2`: boolean to enable the usage of http2 on the request.  There seems to be some incompatibility with the
   golang http2 implementation and AWS ELB/ALBs.  If you experience strange timeouts and hangs, this should be the first
   thing to disable.  Defaults to `false`
+- `enable-retires`: This will tell the forwarder to attempt to send payloads again in case it failed for the first time, 
+   with exponential back off. Defaults to `true`
 - `api-endpoint`: configures the endpoint to submit raw metrics to.  This setting should be just a base URL, for example
   `https://statsd-aggregator.private`, with no path.  Required, no default
 - `max-requests`: maximum number of requests in flight.  Defaults to `1000` (which is probably too high)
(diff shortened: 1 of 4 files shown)


tiedotguy:
The retry pattern used throughout the code is:
```
b := backoff.NewExponentialBackOff()
b.MaxElapsedTime = maxRequestElapsedTime
for {
  if err := do(); err == nil { // always try at least once
    break
  }
  if next := b.NextBackoff(); next == backoff.Stop {
    // stop
  } else {
    // sleep
  }
}
```

It's not clear in the documentation, but a duration of 1 (nanosecond) will result in only one attempt (a duration of 0 will retry forever).

The fact that there's now a section of code with different behavior results in inconsistencies.

Please remove this and update the documentation instead.

README.md
```diff
@@ -72,6 +72,8 @@ following configuration options:
 - `enable-http2`: boolean to enable the usage of http2 on the request.  There seems to be some incompatibility with the
   golang http2 implementation and AWS ELB/ALBs.  If you experience strange timeouts and hangs, this should be the first
   thing to disable.  Defaults to `false`
+- `enable-retires`: This will tell the forwarder to attempt to send payloads again in case it failed for the first time,
```
aelse: "retires"

pkg/statsd/handler_http_forwarder_v2.go
```diff
@@ -110,6 +126,7 @@ func NewHttpForwarderHandlerV2(logger logrus.FieldLogger, apiEndpoint, network s
 		"client-timeout":           clientTimeout,
 		"compress":                 compress,
 		"enable-http2":             enableHttp2,
+		"enable-retries":           enableRetires,
```
aelse: "retires"

pkg/statsd/handler_http_forwarder_v2.go
```diff
@@ -162,6 +179,7 @@ func NewHttpForwarderHandlerV2(logger logrus.FieldLogger, apiEndpoint, network s
 			Transport: transport,
 			Timeout:   clientTimeout,
 		},
+		retries: enableRetires,
```
aelse: "retires"

PR #257, opened 2019-08-22
Revert retries logic (#252-#254) and updates documentation

adampetrovic:


adampetrovic:
As you pointed out, setting max-request-elapsed-time to -1 won't work as there's an explicit check in `NewClient`. Should I rip that check out?

tiedotguy:
You can if you like, but there's a bunch of clients all over the code base which would all need to be updated to keep it consistent.  It's probably not worth it at this point, as I'll get it under my branch anyway.

PR #259, opened 2019-08-22
support debug-mode to print custom metrics to stdout.

XianbinLiu:
- Provide a debug-mode configurable item which defaults to false.
- Once the debug-mode is on, the server should print the metrics received from network to stdout.
  - Should ONLY print the custom metrics.
  - If the terminal/stdout speed is less than metrics incoming speed, it’s reasonable to ignore some of them. So you could not see all the metrics sent to the server from the stdout under such circumstances.

Sample Usage
- dummy client
`while [[ 1 ]];do echo 'abc.def.g:10|c' | nc -w1 -u localhost 8125;echo -n '.';sleep 1;done`
- server with --log-raw-metric=true
`{"level":"info","msg":"[{counter, abc.def.g, 10.000000, , }]","time":"2019-08-26T16:04:33+10:00"}`

### DIFF
--- a/pkg/statsd/parser.go
+++ b/pkg/statsd/parser.go
@@ -3,7 +3,9 @@ package statsd
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"time"
 
@@ -14,6 +16,9 @@ import (
 	log "github.com/sirupsen/logrus"
 	"golang.org/x/time/rate"
 )
+
+// Default buffer size for debug channel
+const debugChannelBufferSize = 1000
 
 // DatagramParser receives datagrams and parses them into Metrics/Events
 // For each Metric/Event it calls Handler.HandleMetric/Event()
@@ -34,10 +39,14 @@ type DatagramParser struct {
 	badLineLimiter *rate.Limiter
 
 	in <-chan []*Datagram // Input chan of datagram batches to parse
+
+	debugMode     bool
+	debugInitOnce sync.Once
+	debugChan     chan []*gostatsd.Metric
 }
 
 // NewDatagramParser initialises a new DatagramParser.
-func NewDatagramParser(in <-chan []*Datagram, ns string, ignoreHost bool, estimatedTags int, handler gostatsd.PipelineHandler, badLineRateLimitPerSecond rate.Limit) *DatagramParser {
+func NewDatagramParser(in <-chan []*Datagram, ns string, ignoreHost bool, estimatedTags int, handler gostatsd.PipelineHandler, badLineRateLimitPerSecond rate.Limit, debugMode bool) *DatagramParser {
 	limiter := &rate.Limiter{}
 	if badLineRateLimitPerSecond > 0 {
 		limiter = rate.NewLimiter(badLineRateLimitPerSecond, 1)
@@ -50,6 +59,7 @@ func NewDatagramParser(in <-chan []*Datagram, ns string, ignoreHost bool, estima
 		namespace:      ns,
 		metricPool:     pool.NewMetricPool(estimatedTags + handler.EstimatedTags()),
 		badLineLimiter: limiter,
+		debugMode:      debugMode,
 	}
 }
 
@@ -71,6 +81,8 @@ func (dp *DatagramParser) RunMetrics(ctx context.Context) {
 }
 
 func (dp *DatagramParser) Run(ctx context.Context) {
+	dp.initDebugMode(ctx)
+
 	for {
 		select {
 		case <-ctx.Done():
@@ -89,6 +101,14 @@ func (dp *DatagramParser) Run(ctx context.Context) {
 			}
 			if len(metrics) > 0 {
 				dp.handler.DispatchMetrics(ctx, metrics)
+
+				if dp.debugMode {
+					// Deliver only once, may lose packet on stdout if the channel buffer is full (e.g. slow terminal).
+					select {
+					case dp.debugChan <- metrics:
+					default:
+					}
+				}
 			}
 			atomic.AddUint64(&dp.metricsReceived, uint64(len(metrics)))
 			atomic.AddUint64(&dp.eventsReceived, accumE)
@@ -169,4 +189,23 @@ func (dp *DatagramParser) parseLine(line []byte) (*gostatsd.Metric, *gostatsd.Ev
 		metricPool: dp.metricPool,
 	}
 	return l.run(line, dp.namespace)
+}
+
+func (dp *DatagramParser) initDebugMode(ctx context.Context) {
+	if dp.debugMode {
+		dp.debugInitOnce.Do(func() {
+			dp.debugChan = make(chan []*gostatsd.Metric, debugChannelBufferSize)
+
+			go func() {
+				for {
+					select {
+					case <-ctx.Done():
+						return
+					case metrics := <-dp.debugChan:
+						fmt.Printf("[%s]: %s\n", time.Now().Format("2019-08-22 00:00:00"), metrics)
+					}
+				}
+			}()
+		})
+	}
 }
--- a/pkg/statsd/parser.go
+++ b/pkg/statsd/parser.go
@@ -3,7 +3,6 @@ package statsd
 import (
 	"bytes"
 	"context"
-	"fmt"
 	"strings"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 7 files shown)


aelse:
There's already a backend to write to stdout, which you can select with `--backends=stdout`
Does that satisfy this use case?

aelse:
An example invocation is:
`gostatsd --backends=stdout --flush-interval=2s --expiry-interval=1s --internal-namespace=statsd.internal.metrics`

XianbinLiu:
> There's already a backend to write to stdout, which you can select with `--backends=stdout`
> Does that satisfy this use case?

The stdout backend would print lots of internal metrics collected by server itself, this would flush the terminal screen while users just need to inspect what metrics are being sent by their applications.

tiedotguy:
The internal metrics can be disabled through `--statser-type=null`.  I also don't like the term "debug mode", as it's quite generic.   My assumption is that debug mode refers to the log level.

To be expicit, do you want to see the raw data (packets, which may include unparseable data), the raw metrics, or the aggregated metrics?

pkg/statsd/parser.go
```diff
@@ -170,3 +190,22 @@ func (dp *DatagramParser) parseLine(line []byte) (*gostatsd.Metric, *gostatsd.Ev
...
+		dp.debugInitOnce.Do(func() {
+			dp.debugChan = make(chan []*gostatsd.Metric, debugChannelBufferSize)
+
+			go func() {
+				for {
+					select {
+					case <-ctx.Done():
+						return
+					case metrics := <-dp.debugChan:
+						fmt.Printf("[%s]: %s\n", time.Now().Format("2019-08-22 00:00:00"), metrics)
```
tiedotguy: This bypasses the normal logger, which breaks json output.

tiedotguy:
It might be worth putting this through a rate limiter (see the `badLineRateLimitPerSecond` in `NewDatagramParser`).

Also, if you're looking at a terminal you should be able to just run a `tcpdump` - that depends what you're looking to get out of it of course.

aelse:
> The stdout backend would print lots of internal metrics collected by server itself, this would flush the terminal screen while users just need to inspect what metrics are being sent by their applications.

Depending on what you want you could disable those internal metrics per @tiedotguy , or grep out the ones you don't want. For viewing raw metrics sent by an application you can simply use netcat.

XianbinLiu:
> The internal metrics can be disabled through `--statser-type=null`. I also don't like the term "debug mode", as it's quite generic. My assumption is that debug mode refers to the log level.
> 
> To be expicit, do you want to see the raw data (packets, which may include unparseable data), the raw metrics, or the aggregated metrics?

Just the raw metrics from the user's perspective, when user deploy the sidecar locally to debug their applications.

And you're right, the term may be misleading, any suggestion on that?

pkg/statsd/parser.go
tiedotguy: The time format is wrong.

tiedotguy:
I'd call it something like "log raw metrics".  Something descriptive.  I'm all about being descriptive and intent.

Can you please also include sample output for both json logs and regular.

pkg/statsd/parser.go
tiedotguy: ... it doesn't actually matter once it's replaced with a proper logger, but worth investigating for the exposure to golangs time formatting.

pkg/statsd/statsd_defaults_and_params.go
```diff
@@ -180,6 +184,7 @@ func AddFlags(fs *pflag.FlagSet) {
 	fs.Bool(ParamConnPerReader, DefaultConnPerReader, "Create a separate connection per reader (requires system support for reusing addresses)")
 	fs.String(ParamServerMode, DefaultServerMode, "The server mode to run in")
 	fs.String(ParamHostname, getHost(), "overrides the hostname of the server")
+	fs.Bool(ParamLogRawMetric, DefineLogRawMetric, "Print metrics received from network to stdout in JSON format")
```
tiedotguy: It just logs it - whether it's to stdout or is JSON formatted is dependent on other configuration.

pkg/statsd/statsd_defaults_and_params.go
```diff
@@ -84,6 +84,8 @@ const (
 	DefaultBadLinesPerMinute = 0
 	// DefaultServerMode is the default mode to run as, standalone|forwarder
 	DefaultServerMode = "standalone"
+	// DefineLogRawMetric is the default value for debug-mode flag
+	DefineLogRawMetric = false
```
tiedotguy: `Default`, not `Define`.  Also leftover `debug-mode`  language.

tiedotguy:
Code looks ok, I'll defer to @adampetrovic or @MovieStoreGuy to determine if it satisfies the internal requirements, as they're not listed on an issue in the project.

pkg/statsd/parser.go
```diff
@@ -34,10 +38,14 @@ type DatagramParser struct {
 	badLineLimiter *rate.Limiter
 
 	in <-chan []*Datagram // Input chan of datagram batches to parse
+
+	logRawMetric         bool
+	logRawMetricInitOnce sync.Once
+	logRawMetricChann    chan []*gostatsd.Metric
```
adampetrovic: Nitpick: can we name this `logRawMetricChan` to be consistent with other areas of the codebase and with the `chan` keyword itself?

pkg/statsd/statsd_defaults_and_params.go
```diff
@@ -180,6 +184,7 @@ func AddFlags(fs *pflag.FlagSet) {
 	fs.Bool(ParamConnPerReader, DefaultConnPerReader, "Create a separate connection per reader (requires system support for reusing addresses)")
 	fs.String(ParamServerMode, DefaultServerMode, "The server mode to run in")
 	fs.String(ParamHostname, getHost(), "overrides the hostname of the server")
+	fs.Bool(ParamLogRawMetric, DefaultLogRawMetric, "Print metrics received from network to stdout in JSON format")
```
MovieStoreGuy: This should also be added to the readme as well please :)

MovieStoreGuy:
Not entirely sure that this is required, @tiedotguy, @adampetrovic or @aelse can jump in here but it might be worth while that read the channel open status (not sure of its technical name) so that we can ensure that clean up go routines once the program is stopped.

Doing something like this https://play.golang.org/p/c7_7lNDm8QN

This being said, I am not sure if its required or considered good practice to ensure that the channel is still open and use that information to exit out of functions so I'll leave that to the other reviewers on this PR.

Otherwise, it would just adding the changes within the readme as well and I think are golden 🎉

tiedotguy:
@MovieStoreGuy the goroutine termination is handled ok - the context is cancelled, resulting in the receive on line 199 causing things to return.  It could be done as more like...
```
go func() {
  for metrics := range dp.logRawMetricChann {
    log metrics
  }
}
```

... but that means **something** needs to close `dp.logRawMetricChann`, and the pattern we use throughout the code base for that is already contexts.

There is a hazard that once the context is cancelled, the `DatagramParser` can't be cleanly restarted (because `logRawMetricInitOnce` has already been fired), but that's not something we support generally anyway.

pkg/statsd/parser.go
```diff
@@ -170,3 +183,37 @@ func (dp *DatagramParser) parseLine(line []byte) (*gostatsd.Metric, *gostatsd.Ev
...
+	}
+}
+
+func (dp *DatagramParser) doLogRawMetric(metrics []*gostatsd.Metric) {
+	if dp.logRawMetric {
+		// Deliver only once, we don't want the whole pipeline being blocked due to a slow terminal.
+		// So may lose packet if the channel buffer is full.
+		select {
+		case dp.logRawMetricChann <- metrics:
+		default:
```
juliawong: nit: should be able to drop this line

pkg/statsd/statsd_defaults_and_params.go
```diff
@@ -84,6 +84,8 @@ const (
 	DefaultBadLinesPerMinute = 0
 	// DefaultServerMode is the default mode to run as, standalone|forwarder
 	DefaultServerMode = "standalone"
+	// DefaultLogRawMetric is the default value for log-raw-metric flag
```
juliawong:
```suggestion
	// DefaultLogRawMetric is the default value for whether to log the metrics received from network
```

Instead of referring to the _log-raw-metric flag_, I think it would improve readability if we included something explaining what this flag is doing

tiedotguy:
Someone plz make a release.

Issue #251, opened 2019-08-06
TestDispatchMetricsShouldDistributeMetrics is flaky


tiedotguy:
```
--- FAIL: TestDispatchMetricsShouldDistributeMetrics (0.01s)
    handler_backend_test.go:148: 
        	Error Trace:	handler_backend_test.go:148
        	Error:      	Not equal: 
        	            	expected: 304
        	            	actual  : 303
        	Test:       	TestDispatchMetricsShouldDistributeMetrics
    handler_backend_test.go:153: aggregator 0 was invoked 63 time(s)
    handler_backend_test.go:153: aggregator 1 was invoked 62 time(s)
    handler_backend_test.go:153: aggregator 2 was invoked 59 time(s)
    handler_backend_test.go:153: aggregator 3 was invoked 56 time(s)
    handler_backend_test.go:153: aggregator 4 was invoked 63 time(s)
```

Seems to fail 1-2% of the time.  Race related:
```
	wg.Wait()       // Wait for all metrics to be dispatched
	cancelFunc()    // After all metrics have been dispatched, we signal dispatcher to shut down
	wgFinish.Wait() // Wait for dispatcher to shutdown
```

We dispatch all the metrics then cancel the backend as soon as they're in the pipeline.  The backend handler may not have drained the pipeline before we cancel it.

Oddly `TestDispatchMetricMapShouldDistributeMetrics` doesn't seem to have the same failure.  It does a `DispatchMetricMap` on the same goroutine but there's still channels to process it.  It's not immediately obvious what the difference is (it might just be much rarer)

PR #260, opened 2019-08-25
Configurable transport (#249)

### LINKED ISSUE #249: Consolidate http.Client creation
There are several `http.Client`s throughout the code base:
- AWS
- HTTP forwarder
- Newrelic
- Datadog

In addition there's plans for clustering and requests for InfluxDB, SignalFX, and various stream backends.

By centralizing the creation, it ensures things like TLS are setup right without copying and pasting all over the place.

This will also enable the ability to inject alternate `http.RoundTripper`s.  Specifically this is the first step in adding support for streams (#10, #84, #159), allowing both aggregated (by injecting an alternate transport for Datadog, etc) and raw (by injecting an alternate transport to the HTTP forwarder).

tiedotguy:
This centralizes the creation of HTTP clients throughout the codebase, for everything except the AWS Cloud Provider (it has some http2 stuff in there that I don't have a good way to test in a real environment, so I'm hesitant to touch it).

It's a pretty big change, but broken in to isolated commits, all of which should be standalone.

This closes #249, and will lead to stream support.


TRANSPORT.md
```diff
@@ -0,0 +1,58 @@
+Configurable transports
+-----------------------
+Almost all the http clients (AWS Cloud Provider is currently excluded) used throughout the service are configurable
+in a uniform matter.  Various components will take a `transport` setting to select which named client to use.  Each
```
aelse: s/matter/manner/

pkg/transport/client.go
```diff
@@ -0,0 +1,13 @@
+package transport
+
+import (
+	"net/http"
+)
+
+// Client is a holder of an http.Client.  In future it will have some high level logic
+// attached to it (PostProtobuf, PostJson, retries, metrics, etc).  The underlying Client
+// is exposed so that things that require a real http.Client (such as Cloudwatch) can
+// still utilize the TransportPool.
```
aelse: This is a very leaky abstraction. Everywhere `transport.Client` is used it's to call `Do` on the underlying `http.Client`. I'd suggest you might as well embrace this and embed `http.Client` here, but on its own that doesn't solve the cloudwatch case.

Alternatives:

- for this first cut I think you can remove `transport.Client`(using `http.Client` instead) and in future introduce what you need to support other features / transports
- implement `Do` on `transport.Client` to call the underlying Client, which at least eliminates some stutter

pkg/transport/pool.go
```diff
@@ -0,0 +1,95 @@
...
+const defaultTransportClientTimeout = 10 * time.Second
+const defaultTransportType = transportTypeHttp
+
+// TransportPool creates http.Clients as required, using the provided viper.Viper for configuration.
+type TransportPool struct {
+	config *viper.Viper
+	logger logrus.FieldLogger
+
+	mu      sync.Mutex
+	clients map[string]*Client
```
aelse: You can replace the mutex and map with `sync.Map` (available since 1.9) which is optimised for this write-once-read-many case

pkg/transport/client.go
tiedotguy: Since generally all (most) of the `x.Client.Do`s and surrounding code will be removed anyway, I'll switch to just storing `pool.Get(...).Client` for now.  It will also simplify the diff a bit, because the actual `client.Do` won't change at all then.

TRANSPORT.md
tiedotguy: :coneofshame:

pkg/transport/pool.go
tiedotguy: Generally speaking, I'm not a fan of the `sync.Map`, just because it's all weakly typed.  When it's not critical (such as this), I prefer simple.

In practice it's likely to only ever be 1 read + 1 write, and the lock only exists so I don't need to check if there's anything that's not safe.

tiedotguy:
I'll park this until #259 is merged, it'll make it a bit cleaner with the change log.

tiedotguy:
Needs another stamp after merging master.

tiedotguy:
Failed build is #251, I guess Travis doesn't have my new email, so it didn't notify me.

tiedotguy:
cc @MovieStoreGuy

pkg/statsd/handler_http_forwarder_v2.go
```diff
@@ -97,25 +102,25 @@ func NewHttpForwarderHandlerV2(logger logrus.FieldLogger, apiEndpoint, network s
...
 	if maxRequestElapsedTime <= 0 && maxRequestElapsedTime != -1 {
 		return nil, fmt.Errorf("max-request-elapsed-time must be positive")
 	}
 	if flushInterval <= 0 {
 		return nil, fmt.Errorf("flush-interval must be positive")
 	}
 
+	httpClient, err := pool.Get(transport)
+	if err != nil {
+		logger.WithError(err).Error("failed to create http client")
```
MovieStoreGuy: This isn't always going to be a http client right?

pkg/transport/transport_http.go
```diff
@@ -0,0 +1,107 @@
...
+	"time"
+
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/viper"
+)
+
+// There are other options on http.Transport, but these are the
+// ones which have been configured in gostatsd historically.
+
+const paramHttpDialerKeepAlive = "dialer-keep-alive"
```
MovieStoreGuy: Was there a reason for not adding a const block instead defining const in front of each?

pkg/statsd/handler_http_forwarder_v2.go
tiedotguy: It's a `transport.Client`, which has a `Client` member which is an `http.Client`.  Those types won't change, but the usage will change from `transport.Client.Client.Do()` to `transport.Client.PostXyz()` in most places.

pkg/transport/transport_http.go
tiedotguy: Not that I can tell.  I think I just started with only needing 1 constant and then copy/pasting it a bunch.

pkg/statsd/handler_http_forwarder_v2.go
MovieStoreGuy: Wouldn't this error be confusing if left since it could be any of the configured transports in future but we only report that we failed to create a http client?

tiedotguy: I'm not sure I understand what you mean.  Do you mean it's not obvious which transport has been configured and failed?

pkg/transport/transport_http.go
MovieStoreGuy: If we can move this to a const block, it matches the style used throughout the repo already.

tiedotguy: I'll clean it up in the PR that consumes it.  There's work based off this branch already.

pkg/statsd/handler_http_forwarder_v2.go
MovieStoreGuy: So if ths could was not to change, but you added another transport. Wouldn't this error be misleading?

tiedotguy: Gotcha.  Yeah, I'll change it to just "transport" in the next PR.

MovieStoreGuy:
@aelse you good with all this?

PR #261, opened 2019-09-17
Adding support for custom headers when sending requests.

MovieStoreGuy:
In order to forward requests to different target groups, adding the ability to add additional custom headers gives that opportunity at no cost. 

Since we do not have cluster ready yet, running a singular instance has proven problematic since we can't reason who is sending us data and rely on IP addresses to investigate further. Adding custom headers allows us to inspect these at the ALB and shard noisy neighbours onto their own deployment stack.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+17.2.1
+------
+- Support for `custom-headers` for http_forwarder_v2
+
 17.2.0
 ------
 - Support `log-raw-metric` configuration
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-17.2.1
+18.0.0
 ------
 - Support for `custom-headers` for http_forwarder_v2
 
--- a/README.md
+++ b/README.md
@@ -85,8 +85,8 @@ following configuration options:
 - `flush-interval`: duration for how long to batch metrics before flushing. Should be an order of magnitude less than
   the upstream flush interval. Defaults to `1s`.
 - `log-raw-metric`: print metrics received from network to stdout in JSON format.
-- `custom-headers` : a map of strings that are added to each request sent to allow for additional network routing / request inspection . 
-  Not required, default is empty.
+- `custom-headers` : a map of strings that are added to each request sent to allow for additional network routing / request inspection. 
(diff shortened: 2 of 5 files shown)


pkg/statsd/handler_http_forwarder_v2.go
```diff
@@ -368,18 +388,6 @@ func (hfh *HttpForwarderHandlerV2) serialize(message proto.Message) ([]byte, err
 	return buf, nil
 }
 
-// debug rendering
```
MovieStoreGuy: Nixed this as it wasn't be used and should have been cleaned up.

tiedotguy: Please restore this - it's deliberately here for debug purposes.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+17.2.1
```
tiedotguy: This changed a public API (`NewHttpForwarderHandlerV2`), so it's a major bump.

README.md
```diff
@@ -85,6 +85,8 @@ following configuration options:
 - `flush-interval`: duration for how long to batch metrics before flushing. Should be an order of magnitude less than
   the upstream flush interval. Defaults to `1s`.
 - `log-raw-metric`: print metrics received from network to stdout in JSON format.
+- `custom-headers` : a map of strings that are added to each request sent to allow for additional network routing / request inspection. 
+  Not required, default is empty. Example: `--custom-headers='{'region' : 'us-east-1', 'service' : 'event-producer'}'`
```
aelse: Single quotes outside, double quotes inside.

pkg/statsd/handler_http_forwarder_v2.go
MovieStoreGuy: It's shift in its spot but it is back for when you need it.

tiedotguy: I mean... I wanted it back so it doesn't conflict with my branch.

Now I have two conflicts :|

MovieStoreGuy: Ah, my bad.

MovieStoreGuy:
I believe everything should be cleaned up @aelse , would you mind giving it a once threw?

PR #263, opened 2019-09-18
Build gostatsd with go 1.13

aelse:
The go modules config was inconsistent with glide. This CL replaces glide with go modules and builds the public docker images with go 1.13 (also tested with go 1.12 and GO111MODULES=on).


tiedotguy:
I'm so glad someone who understands modules picked this up, so I can continue to drift along barely aware of how modules work for another day 👍

PR #270, opened 2019-09-25
Remove 8 duplicate getSubViper calls

### LINKED ISSUE #262: De-dupe getSubViper
This little function has been the bane of my career because I was never sure where to put it.

It's time for it to go.

tiedotguy:
Closes #262

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+19.0.1
+------
+- Internal refactor to de-dupe 8 getSubVipers
+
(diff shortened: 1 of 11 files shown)


ash2k:
Is it still needed? Maybe the bug that this function works around has been fixed in viper?

tiedotguy:
I'm not sure what the original bug was, however `v.Sub` returns `nil` if the key is missing or not a map.  It seems deliberate.

Issue #159, opened 2018-02-22
google pubsub support


joemiller:
We recently added support for publishing to google pubsub to the carbon-relay-ng project, and consuming from pubsub to the go-carbon project. It's working well enough that I'm considering also adding pubsub support to gostatsd. Right now our pipeline looks like:

```
clients -> [gostatsd x 3 (ingestion tier)] -> [gostatsd x1 (aggregator)] -> [graphite-relay-ng] -> [google pubsub] -> [go-carbon]
```
Instead of having the gostatsd aggregator tier funnel into graphite-relay-ng over TCP we could have gostatsd publish in graphite format to pubsub, bypassing the TCP pipe to graphite-relay-ng. I suspect this could add a little more robustness to the pipeline. For example, right now when we update the graphite-relay-ng pods (it's a kubernetes deployment) there is a minute or so of lost metrics from the gostatsd tier. In this setup the pipeline would look like this:

```
clients -> [gostatsd x 3 (ingestion tier)] -> [gostatsd x1 (aggregator)] -> [google pubsub] -> [go-carbon]
```

In addition to gostatsd publishing to google pubsub in graphite format as in the example above, it might also be desirable for the gostatsd ingestion tier to publish statsd format to pubsub, such as:

```
clients -> [gostatsd x 3 (ingestion tier)] -> [google pubsub] -> [gostatsd x1 (aggregator)] -> [google pubsub] -> [go-carbon]
```

The goal being fewer dropped metrics when the gostatsd-aggregator process is restarted or dies.

Questions for gostatsd team:

1. Would the project consider PRs that added google-pubsub support?
2. If so, is there a recommended approach? In looking at the existing graphite-TCP backend code I think I would end up copying significant amounts of code to format the payload into graphite linemode format. So would it be better to add a configuration option to the existing graphite backend to toggle between TCP and google-pubsub? Or, create a new separate 'google pubsub' backend?

ash2k:
1. I'll review PRs (when I have time), np here. Whether we want this or not - I think we do. However I'll let @tiedotguy decide, he runs it in prod.
2. If we want this functionality then I'd go with a separate backend and extract reusable pieces of code into a shared package - just like `pkg/backends/sender` which is used in multiple backends.

joemiller:
@ash2k Thanks for the pointer to the sender.

tiedotguy:
If it's useful, go for it!

[...]

Issue #84, opened 2017-01-05
Add Kafka backend


Krieke:

ash2k:
[PRs welcome](https://github.com/atlassian/gostatsd#contributors)!

aelse:
I'm going to close off this issue. A PR implementing a kafka backend (for aggregate or raw metrics) would probably be fine should someone wish to do the work.

Issue #14, opened 2016-05-21
Add influxdb backend


ash2k:

tiedotguy:
If there's anyone watching this issue and wants to provide feedback on the decisions, please do so.  The update to BACKENDS.md should capture everything, so knowledge of go and the code is not required (but feedback is always welcome)

Issue #10, opened 2016-05-21
Add kinesis backend


ash2k:

aelse:
I'm going to close this off as there isn't a need for it right now. A PR implementing this might be ok if someone were to do the work.

PR #272, opened 2019-10-24
Semi-RFC: Consolidate http behavior

tiedotguy:
This is the next change in the whole transport refactor saga, leading towards streaming (#10, #84, #159), InfluxDB (#14), and clustering.  Nothing here is consumed currently, it's just the primitives.  As it's getting rather large already, I wanted to minimize the review size.  Commits are relatively stand-alone.

This PR is not expected to be merged yet, because there's documentation mismatches (due to lack of consumption).  I'll be working on updating consumers in a subsequent PR, but didn't want to get too deep in to that in case there's feedback on the API (I'm not a huge fan of the `TransportOptions` for example)

### DIFF
--- a/pkg/transport/client_low.go
+++ b/pkg/transport/client_low.go
@@ -16,13 +16,17 @@ import (
 )
 
 type Client struct {
-	messagesSent     uint64 // atomic - messages successfully sent
-	messagesRetried  uint64 // atomic - retries (first attempt is not a retry, final failure is not a retry)
-	messagesDropped  uint64 // atomic - final failure, does not include semaphore timeout
-	messagesTimedOut uint64 // atomic - messages timed out waiting for a semaphore slot
+	messagesFailMarshal  uint64 // atomic - messages which failed to be marshalled
+	messagesFailCompress uint64 // atomic - messages which failed to be compressed
+	messagesSent         uint64 // atomic - messages successfully sent
+	messagesRetried      uint64 // atomic - retries (first attempt is not a retry, final failure is not a retry)
+	messagesDropped      uint64 // atomic - final failure, does not include semaphore timeout
+	messagesTimedOut     uint64 // atomic - messages timed out waiting for a semaphore slot
 
 	logger        logrus.FieldLogger
+	compress      bool
 	customHeaders map[string]string
+	debugBody     bool
 	requestSem    util.Semaphore
 	userAgent     string
 	backoff       util.BackoffFactory
--- a/pkg/transport/client_low.go
+++ b/pkg/transport/client_low.go
@@ -75,7 +75,7 @@ func (hc *Client) emitMetrics(statser stats.Statser) {
 
 // PostRaw will start a goroutine (semaphore allowing) which attempts to POST the provided data
 // to the provided URL.  It is fire-and-forget by design.
-func (hc *Client) PostRaw(ctx context.Context, url, contentType, encoding string, headers map[string]string, body []byte) {
+func (hc *Client) PostRaw(ctx context.Context, url string, contentType ContentType, encoding Encoding, headers map[string]string, body []byte) {
 	if !hc.requestSem.Acquire(ctx) {
 		atomic.AddUint64(&hc.messagesTimedOut.cur, 1)
 		return
@@ -87,7 +87,7 @@ func (hc *Client) PostRaw(ctx context.Context, url, contentType, encoding string
 	}()
 }
 
-func (hc *Client) do(ctx context.Context, url, contentType, encoding string, body []byte, headers map[string]string) {
+func (hc *Client) do(ctx context.Context, url string, contentType ContentType, encoding Encoding, body []byte, headers map[string]string) {
 	var err error
 
 	bo := hc.backoff()
@@ -117,7 +117,7 @@ func (hc *Client) do(ctx context.Context, url, contentType, encoding string, bod
 	hc.logger.WithError(err).Info("failed to send, giving up")
 }
 
-func (hc *Client) constructPost(ctx context.Context, body []byte, url, contentType, encoding string, headers map[string]string) func() error {
+func (hc *Client) constructPost(ctx context.Context, body []byte, url string, contentType ContentType, encoding Encoding, headers map[string]string) func() error {
 	return func() error {
 		req, err := http.NewRequest("POST", url, bytes.NewReader(body))
 		if err != nil {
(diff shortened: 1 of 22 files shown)


pkg/transport/client_low.go
```diff
@@ -0,0 +1,165 @@
...
+		req, err := http.NewRequest("POST", url, bytes.NewReader(body))
+		if err != nil {
+			return fmt.Errorf("unable to create http.Request: %v", err)
+		}
+
+		req = req.WithContext(ctx)
+
+		// Base headers
+		req.Header.Set("Content-Type", contentType)
+		req.Header.Set("Content-Encoding", encoding)
```
ash2k: Encoding and type could be enums for a little bit more type safety. Currently there is a risk to pass two strings in swapped positions.

pkg/transport/helpers.go
```diff
@@ -0,0 +1,79 @@
...
+	} else {
+		json = jsonConfig
+	}
+	stream := json.BorrowStream(buf)
+	stream.WriteVal(data)
+	err := stream.Flush()
+	if err != nil {
+		return nil, err
+	}
+	json.ReturnStream(stream)
```
ash2k: So if there was an error we don't return the stream? This line should be after `Flush`?

pkg/transport/pool.go
```diff
@@ -58,17 +100,34 @@ func (tp *TransportPool) newClient(name string) (*Client, error) {
...
+	userAgent := sub.GetString(paramTransportUserAgent)
 
 	if clientTimeout < 0 {
 		return nil, errors.New(paramTransportClientTimeout + " must not be negative") // 0 = no timeout
 	}
 
+	backoff, err := util.GetRetryFromViper(sub)
+	if err != nil {
+		tp.logger.WithField("name", name).WithError(err).Error("failed to load retry policy")
+		return nil, err
```
ash2k: Is there a reason to both log and return the error? Ideally should only return it

pkg/transport/pool.go
```diff
@@ -80,16 +139,37 @@ func (tp *TransportPool) newClient(name string) (*Client, error) {
 		return nil, err
 	}
 
+	headers := map[string]string{}
+	headerKeys := []string{}
```
ash2k: Could pre-allocate the exact size for both

pkg/util/backoff.go
```diff
@@ -0,0 +1,82 @@
...
+	v.SetDefault(paramRetryMaxTime, defaultRetryMaxTime)   // constant + exponential
+	v.SetDefault(paramRetryPolicy, defaultRetryPolicy)
+
+	retryInterval := v.GetDuration(paramRetryInterval) // constant
+	retryMaxCount := v.GetInt64(paramRetryMaxCount)    // constant + exponential
+	retryMaxTime := v.GetDuration(paramRetryMaxTime)   // constant + exponential
+	retryPolicy := v.GetString(paramRetryPolicy)
+
+	if retryInterval <= 0 {
+		return nil, errors.New(paramRetryInterval + " must be positive")
```
ash2k: Could use `fmt.Errorf()` here and below for consistency with other places where things are not strings. Not a big deal.

pkg/util/semaphore_test.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"context"
+	"sync"
+	"sync/atomic"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestSemaphoreUnlimited(t *testing.T) {
```
ash2k: `t.Parallel()`

pkg/transport/client_low.go
tiedotguy: Swapping is definitely a hazard, but I don't know how to prevent it.  Go lacks enums (near as I can tell), and will happily promote a string to an alias of it, breaking the type safety: https://play.golang.org/p/lzAsJ8gGB_O

I could do a sanity check for valid values, but that seems overkill.  Is there something I'm missing?

pkg/transport/helpers.go
tiedotguy: Good catch!

- [ ] this

pkg/transport/pool.go
tiedotguy: Not specifically.  Depending on the call stack it could actually end up dropping a few duplicate logs - they're all fatal though (fail early), so it's something I didn't put a lot of thought in to.

For preference I'd actually make it a `.Fatal()`. I've been slowly pushing to remove errors from the code (eg, this whole refactor is fire and forget) with a view to either fail-fast on startup, or handle runtime errors immediately when there's maximum context available.

Ultimately a lot of the error paths just pop out somewhere high level with a log line, and they lose context that could be supplied by an immediate log (such as the transport name in this case).

There's an argument to be made for more strongly typed errors to provide context, but a) I'm lazy, and b) the only way to expose that context is through `error.Error()` which means it's not exposed as structured log data.

pkg/transport/pool.go
tiedotguy: - [ ] this

pkg/util/backoff.go
tiedotguy: My main thought was it should turn it in to a static string at compile time rather than runtime.  I don't know if that will **actually** occur with either `"" + ""` or `fmt.Sprintf("%s", "")`, but that was the intent.

pkg/util/semaphore_test.go
tiedotguy: - [ ] this

pkg/transport/client_low.go
ash2k: Yeah, what we are seeing in that example is the behavior of "untyped literals". This does not compile though because `x` is typed vs untyped literal. https://play.golang.org/p/wiCUwuQqj5g

pkg/transport/pool.go
ash2k: Ok, I see your point. I think this code can be improved further if config parsing is moved into `NewTransportPool()` and viper is no longed referenced/used. I don't know if it's easy to do this. Also, just terminating the program in the middle of the pooling code would be weird. Testing and reusing this code becomes harder.

pkg/transport/client_low.go
tiedotguy: Ahh! I see now.  https://play.golang.org/p/xItV45OW49V

I'll fix it up in the cleanup commit (waiting on other feedback, I know the team is busy atm)

- [ ] this

ash2k: https://blog.golang.org/constants see the `String constants` section

tiedotguy:
Rebased off master, fixed the issues mentioned. Last commit is a bit derpy because I did a `fixup` instead of `reword` when cleaning up the "address issues from review" commit.

MovieStoreGuy:
Hey @tiedotguy , you currently have a merge conflict on this, would you be able to resolve this?

tiedotguy:
No merge conflicts, simply happy accidents.

This branch will return later.

PR #275, opened 2019-11-06
Add support for New Relic's new dimensional metrics pipeline

Kav91:
Adds support for New Relic's dimensional metrics via the new metric api.

https://docs.newrelic.com/docs/data-ingest-apis/get-data-new-relic/metric-api/introduction-metric-api

BACKENDS.md has been updated with additional information as well as relevant tests.

### DIFF
--- a/BACKENDS.md
+++ b/BACKENDS.md
@@ -63,10 +63,12 @@ When `mode` is `legacy`, the graphite backend will emit metrics with the followi
 
 New Relic Backend
 -----------------
-Supports three routes for flushing metrics to New Relic.
-- Directly to the Insights Collector - [Insights Event API](https://docs.newrelic.com/docs/insights/insights-data-sources/custom-data/send-custom-events-event-api)
-- Via the Infrastructure Agent's inbuilt HTTP Server
-- Directly to the Metric API - [Metric Event API](https://docs.newrelic.com/docs/data-ingest-apis/get-data-new-relic/metric-api/introduction-metric-api)
+Supports three flush-types for sending metrics to New Relic:
+- `infra` Via the Infrastructure Agent's inbuilt HTTP Server (default)
+- `insights` Directly to the Insights Collector - [Insights Event API](https://docs.newrelic.com/docs/insights/insights-data-sources/custom-data/send-custom-events-event-api)
+- `metrics` Directly to the Metric API - [Metric Event API](https://docs.newrelic.com/docs/data-ingest-apis/get-data-new-relic/metric-api/introduction-metric-api)
+
+An attempt will be made to automatically set the `flush-type` based on the address you have configured, [manual configuration](#manual-flush-type-configuration) is also available.
 
 ### [New Relic Insights Event API](https://docs.newrelic.com/docs/insights/insights-data-sources/custom-data/send-custom-events-event-api)
 Sending directly to the Event API alleviates the requirement of needing to have the New Relic Infrastructure Agent.
@@ -135,6 +137,20 @@ http_server_enabled: true
 http_server_host: 127.0.0.1 #(default host)
 http_server_port: 8001 #(default port)
 ```
+
+### Manual Flush Type Configuration
+The `flush-type` attribute can be configured with the following available  options - `insights`, `metrics` or `infra`.
+
+```
+[newrelic]
+    transport = "default"
+	flush-type = "insights" # <-- 
+    address = "https://another-collector.newrelic.com/v1/accounts/YOUR_ACCOUNT_ID/events"
+    api-key = "yourEventAPIInsertKey"
+```
+
+### Renaming Attributes
+This option is only available for `insights` and `infra` flush-types.
 
 Additional options are available to rename attributes if required.
 ```
--- a/pkg/backends/newrelic/newrelic.go
+++ b/pkg/backends/newrelic/newrelic.go
@@ -102,8 +102,14 @@ type NRInfraPayload struct {
 // NRMetricsPayload represents New Relic Metrics Payload format
 // https://docs.newrelic.com/docs/data-ingest-apis/get-data-new-relic/metric-api/report-metrics-metric-api#new-relic-guidelines
 type NRMetricsPayload struct {
-	Common  map[string]interface{} `json:"common"`
-	Metrics []interface{}          `json:"metrics"`
+	Common  NRMetricsCommon `json:"common"`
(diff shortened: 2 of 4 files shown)


BACKENDS.md
```diff
@@ -63,28 +63,69 @@ When `mode` is `legacy`, the graphite backend will emit metrics with the followi
...
-#Example configuration
+#Example Metric API configuration
+
+[newrelic]
+    transport = "default"
+    address-metrics = "https://metric-api.newrelic.com/metric/v1"
+    api-key = "yourEventAPIInsertKey"
+```
+
+> Make sure that `address-metrics` is set to `https://metric-api.newrelic.com/metric/v1` if your account hosts data in the EU data center.
```
carlosroman:
```suggestion
> Make sure that `address-metrics` is set to `https://metric-api.eu.newrelic.com/metric/v1` if your account hosts data in the EU data center.
```

pkg/backends/newrelic/newrelic.go
```diff
@@ -294,24 +311,29 @@ func (n *Client) constructPost(ctx context.Context, buffer *bytes.Buffer, data i
...
 	return func() error {
 		headers := map[string]string{
 			"Content-Type": "application/json",
 			"User-Agent":   n.userAgent,
 		}
 
 		// Insights Event API requires gzip or deflate compression
-		if n.flushType == flushTypeInsights && n.apiKey != "" {
+		// https://docs.newrelic.com/docs/insights/insights-data-sources/custom-data/introduction-event-api#h2-basic-workflow
+		// Metrics API requires gzip or identity
```
tiedotguy: Oh man, I'm going to have to add gzip to my http client refactoring!

Should be pretty trivial though :)

pkg/backends/newrelic/newrelic.go
```diff
@@ -88,14 +90,22 @@ type Client struct {
...
 	Name               string        `json:"name"`
 	ProtocolVersion    string        `json:"protocol_version"`
 	IntegrationVersion string        `json:"integration_version"`
 	Data               []interface{} `json:"data"`
 }
 
+// NRMetricsPayload represents New Relic Metrics Payload format
+// https://docs.newrelic.com/docs/data-ingest-apis/get-data-new-relic/metric-api/report-metrics-metric-api#new-relic-guidelines
+type NRMetricsPayload struct {
+	Common  map[string]interface{} `json:"common"`
```
tiedotguy: Can `Common` and `Metrics` be more well defined rather than a `map[string]interface{}`?

In the `NRInfraPayload` it uses metric names as keys, so it can't be well defined, but in `NRMetricsPayload` it has an actual `name` key, so it can.

Ideally everything that can possibly be well defined should be.

Kav91: I've added both a common and metric struct now. 
NRMetric struct is used within flush.go, I have intentionally left the NRMetricsPayload struct as []interface{} rather then []NRMetric so that the timeseries struct in flush.go can remain common across the flush types used.

pkg/backends/newrelic/newrelic.go
Kav91: So these need to be kept separate, the reason for that is our new dimensional metrics pipeline cannot accept arbitrary events on the same endpoint. 

When you use address-metrics and supply it the new metrics endpoint it will ship gostatsd metrics as NR dimensional metrics. However if you want to also send gostatsd events, you will need to configure our insights collector on that original address option as well.

We've documented this clearly in the BACKENDS.md.
Unfortunately at this moment there is no easier way to simplify this right now.

pkg/backends/newrelic/newrelic.go
```diff
@@ -88,14 +90,38 @@ type Client struct {
...
+	Attributes map[string]interface{} `json:"attributes"`
+	IntervalMs float64                `json:"interval.ms"`
+}
+
+// NRMetric metric for New Relic Metrics Format
+type NRMetric struct {
+	Name       string                 `json:"name"`
+	Value      interface{}            `json:"value,omitempty"`
+	Type       string                 `json:"type,omitempty"`
+	Timestamp  gostatsd.Nanotime      `json:"timestamp,omitempty"`
```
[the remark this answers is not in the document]
Kav91: Updated

Kav91:
Have documented the flush type options now.

Also background for metric-name, metric-type and the other configurable attributes they were made as such to support the insights & infra flush-types, the reason for that was users were particular in the way they wanted to query their attributes coming from other StatsD solutions, so this made it easier for them to switch.

With the new metrics pipeline, this is obviously a lot more standardized and built for things like StatsD. Ultimately we would like all NR users to use the dimensional metrics `metrics` flush type, as the default solution going forward.

There is some background work we'd like to complete before that, so will plan another PR once ready and hopefully we can remove the other options and tidy it all up.

@tiedotguy thanks as always for the review, and appreciate your understanding 👍

tiedotguy:
Ahh, I was under the impression the data was more structured.  Since I'm not familiar with NR specifically I lack the context on why it needs to be configurable :)

And thank you for bringing this in - looking through the forks, there's a handful of alternate backends which have been added by other companies.  Because my refactoring tends to be quite invasive, it would probably be a nightmare to bring them up to date.  With NR in master, it's trivial for me to keep it up to date as I go.

Incidentally, you might be interested in having a look at the histograms PR (#245), I think it will fit https://github.com/newrelic/newrelic-exporter-specs/blob/master/Guidelines.md#histograms with a bit of gentle massaging.  We're still not certain if it's going to make it in, or remain in permanently, but it's a thing.

tiedotguy:
@ash2k, if you've got the time, can I get another set of eyes over this sometime before I hit merge?

Kav91:
@tiedotguy @ash2k just a friendly poke, or if you have any idea when someone can take a look roughly?
We are looking to make this our go to option for StatsD once merged, and want to use the latest on master to build some images and binaries for our users 😄so any timing you can provide would be appreciated.

ash2k:
I don't know anything specific about NewRelic but I'll have a look tomorrow. Sorry for the delay.

Kav91:
hey @ash2k sorry to nudge again, any luck or are we looking at next week sometime? Thx

Kav91:
No worries, & thanks for taking the time to review again @ash2k & @tiedotguy 😄

tiedotguy:
Should be all released now, @Kav91 !

PR #279, opened 2019-11-29
[cloudwatch] Batch requests to Cloudwatch to reduce cost

tiedotguy:
This is a continuation of #204, just moved in to a branch on the main repository so I don't need to deal with any permissions issues.

It merges master to the original branch, and addresses my feedback.

cc @JorgenEvens


wooly:
Thanks @tiedotguy, will have a look at this next week for you.

tiedotguy:
Hi all, hope you had a happy and productive (or non-productive!) break.

Have you had an opportunity to give this a spin with a real CW backend?

pkg/backends/cloudwatch/cloudwatch_test.go
```diff
@@ -17,6 +18,22 @@ import (
...
+// enforce a timeout on tests.  Copied from fixtures_test.go.
+func testContext(t *testing.T) (context.Context, func()) {
+	ctxTest, completeTest := context.WithTimeout(context.Background(), 1100*time.Millisecond)
+	go func() {
+		after := time.NewTimer(1 * time.Second)
+		select {
+		case <-ctxTest.Done():
+			after.Stop()
+		case <-after.C:
+			require.Fail(t, "test timed out")
```
ash2k: I think this is a data race. Please see https://golang.org/pkg/testing/#T and https://github.com/golang/go/issues/15758 and https://github.com/golang/go/issues/24680

pkg/backends/cloudwatch/cloudwatch_test.go
```diff
@@ -189,3 +208,32 @@ func metricsOneOfEach() *gostatsd.MetricMap {
...
+	for i := 0; i < 200; i++ {
+		go func() {
+			cli.SendMetricsAsync(ctx, metricsOneOfEach(), func(errs []error) {
+				ch <- struct{}{}
+			})
+		}()
+	}
+	for i := 0; i < 200; i++ {
+		select {
+		case <-ctx.Done():
```
ash2k: `return`?

pkg/backends/cloudwatch/cloudwatch_test.go
ash2k: An approach to achieve the same result may be to move everything else into goroutines and block on timer in the test goroutine somehow. Then if it fires fail the test. This will not work if the other code also needs to be able to do the same. =\

pkg/backends/cloudwatch/cloudwatch.go
```diff
@@ -170,58 +198,128 @@ func (client Client) buildMetricData(metrics *gostatsd.MetricMap) (metricData []
...
+func (client *Client) flush() {
+	if !client.setupFlushTimer() {
 		return
 	}
 
-	go func() {
-		start := 0
+	// This is safe to read outside the timerMutex, as only the
+	// goroutine that created the initial timer can ever be here
+	<-client.timer.C
```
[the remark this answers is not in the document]
tiedotguy: I don't believe so:

Only one goroutine (call it "A") can be on line 230 at a time, and it's the goroutine that wrote to the timer. Essentially the flush timer is a mutex - if it's present, then the thing that created it will reach here, otherwise it won't. At line 235 the "flush timer as a mutex" is released.

I've called out in my comment above `clearFlushTimer` that it's possible for another goroutine (call it "B") to come along and take the data before "A" does, however that is protected by `queueMutex`, and just results in "A" having nothing to do.

That being said - I refactored this area specifically to make it more clear.  Even if it's safe, if it's still not clear, then I need to make it clearer. A reader of the code shouldn't have to think it through in that detail.

If you can confirm or refute my logic, I'll see about making it easier to follow / fixing it.

pkg/backends/cloudwatch/cloudwatch.go
```diff
@@ -170,58 +198,128 @@ func (client Client) buildMetricData(metrics *gostatsd.MetricMap) (metricData []
...
+	// Copy pending metric items and create new queue instance
+	// client.queue is protected by a Read-Write Lock such
+	// that execution is only blocked while replacing the queue
+	client.queueMutex.Lock()
+	queue := client.queue
+	client.queue = []*QueueItem{}
+	client.queueMutex.Unlock()
+
+	// Run actual function
+	metricData := []*cloudwatch.MetricDatum{}
```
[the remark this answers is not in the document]
tiedotguy: Good idea

- [ ] preallocate queue

pkg/backends/cloudwatch/cloudwatch.go
```diff
@@ -170,58 +198,128 @@ func (client Client) buildMetricData(metrics *gostatsd.MetricMap) (metricData []
...
+			Namespace:  &client.namespace,
+		})
+
+		errors = append(errors, err)
+	}
+
+	// TODO: Metrics about how many metrics were sent, and how many requests were required.  Maybe others.
+
+	for _, q := range queue {
+		go q.callback(errors)
```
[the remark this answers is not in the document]
tiedotguy: Better to be explicit and make sure it's ok.
- [ ] protect `q`

pkg/backends/cloudwatch/cloudwatch_test.go
tiedotguy: Well bugger. This pattern isn't new in the code base. I'll raise a new issue to refactor it out of the project. It "works" enough for now though, are you ok with leaving this instance in?

pkg/backends/cloudwatch/cloudwatch_test.go
tiedotguy: :coneofshame:

- [ ] return

pkg/backends/cloudwatch/cloudwatch.go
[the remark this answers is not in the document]
ash2k: I didn't have enough time to properly check the code here. It's probably fine if you say so :)

pkg/backends/cloudwatch/cloudwatch_test.go
ash2k: Yes, np.

pkg/backends/cloudwatch/cloudwatch.go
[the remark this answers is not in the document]
ash2k: It also may or may not be beneficial to have a slice of values rather than pointers. Of course if they are already pointers when they reach this code then it may not make any difference.

MovieStoreGuy:
Does this PR still need to remain open @tiedotguy ?

tiedotguy:
It'd be nice to get it in, but without a response for giving it a real test, I think no :(

We can always revisit it if needed, and there's a non-zero chance that it won't matter with future refactorings anyway.

PR #282, opened 2020-01-21
Update dependencies and Go version

ash2k:
Ran `go get -u -t ./...` to update all dependencies.

The second commit adds tool tracking via `go.mod`.


Makefile
```diff
@@ -10,27 +10,24 @@ GOBUILD_VERSION_ARGS_WITH_SYMS := -ldflags "-X $(VERSION_VAR)=$(REPO_VERSION) -X
...
 CLUSTER_PKG := github.com/atlassian/gostatsd/cmd/cluster
 PROTOBUF_VERSION ?= 3.6.1
 
 setup: setup-ci
-	go get github.com/githubnemo/CompileDaemon
-	go get github.com/jstemmer/go-junit-report
-	go get golang.org/x/tools/cmd/goimports
+	go install github.com/githubnemo/CompileDaemon \
+		github.com/jstemmer/go-junit-report \
+		golang.org/x/tools/cmd/goimports
```
ash2k: `go get` is a command to get a specific version of a dependency. If no version is specified it tries to get the `latest`. `go install` just installs the version of the binary that is tracked in `go.mod`/`go.sum`.

ash2k: In addition here we pull all dependencies using 1 command - can do more concurrent downloading and should be faster.

.travis.yml
```diff
@@ -6,7 +6,7 @@ services:
   - docker
 
 go:
-  - "1.13"  # Keep in sync with Makefile
+  - "1.13.6"  # Keep in sync with Makefile
```
tiedotguy: Since you're here, and you caught a sync issue in the README.md, can you update the comment here and in Makefile to also update README.md?

Issue #281, opened 2020-01-08
gostatsd as a cluster-aware Kubernetes DaemonSet


mwhittington21:
I'd like to propose an additional module inside gostatsd which would be able to enrich packets with Kubernetes metadata based on which IP address sent a packet.

I'm looking to use gostatsd inside a Kubernetes cluster, one per node, to allow applications running on the node to send metrics to a locally running gostatsd which will be able to augment packets with tags including metadata about which application sent the metric before forwarding it on to an upstream aggregation layer.

This enrichment would come in much the same way as the cloud providers work inside gostatsd today. If inside Kubernetes and with a config file or flag enabling the feature, then incoming metrics would be enriched with data cached from the Kubernetes API server. Without getting too much into the specifics I see it working similar to the pod IP lookup cache inside something like [kube2iam](https://github.com/jtblin/kube2iam).

If this is a feature that you believe could fit into gostatsd then I'm willing to submit a PR making it happen.

tiedotguy:
Sounds good to me. It **is** a cloud provider, really: a CP just an external map of ip -> tags.

My main suggestion would be to have a way to filter out tags (whitelist? prefix? you can also do regex because it's not high churn (expensive CPU)), as I know a lot of noise can be picked up via the EC2 tags.

That noise can be filtered out via filters in the pipeline, but it's better to do it once, rather than for every single metric.

mwhittington21:
Agreed on the filtering. Here's a a few paragraphs that I came up with when I was brainstorming how this would work end to end.

---

Kubernetes pod annotation keys matching a certain prefix (e.g. `gostatsd.github.com/`) could have the second part of their annotation key used as a tag name for the gostatsd. The value of the annotation would then be used as the tag value.

So for example an annotation of `gostatsd.github.com/foo` with the value `bar` would have the gostatsd tag any metrics received from that pod with `foo:bar`.

We would need the following things to be customisable, either through command line flags or the config file:
- a whitelist of default tags to look for inside the annotations of a pod
  - optional - we can always self-discover from the prefix, but this may not be desireable as it can promote misuse
- a prefix for annotations which will be used to find the default tags
  - this would not be required if the whitelist is provided
- the location of a creds file to auth to k8s (if unsupplied then use the in-cluster auth via the mounted token)

[...]

PR #283, opened 2020-02-10
k8s cloud provider

mwhittington21:
Implementation of #281 

I still need to write the docs for this but I wanted to get it up here for comment ASAP.

Things to know about the k8s provider implementation provided here:
* Sets up a watch of pod objects, which can be either targeted at the pod's current node, or at the entire cluster
* Indexes these pods by IP address so we can look them up at-will
* Retrieves tags that are allowed via lists of regexes for both pod annotations and labels
* For the first cut, we just immediately forward all metrics that don't have a pod IP. This is our solution to the race condition of a pod starting to run and send metrics before the API server has sent out information about the Pod's IP address.
  * Not ideal. Future work could be done to fix this but as things stand today this will require a refactor of the metrics pipeline to allow waiting for some metrics without waiting on the whole batch waiting to be processed.
* Has a bunch of k8s options to allow running locally or in-cluster, and to set the user agent so that you can track API server requests back to gostatsd

Things I think are outside the scope of this first implementation:
* The above mentioned waiting for pod IPs to arrive before forwarding metrics
* Deduplicating labels and annotations that have the same key
* Extra k8s ecosystem goodies like prometheus metrics and k8s events
* Allowing multiple cloud providers to run simultaneously (e.g. AWS and k8s)
*


pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,378 @@
...
+	return "", errors.New("no eth0 interface address found")
+}
+
+func podByIpIndexFunc(obj interface{}) ([]string, error) {
+	pod := obj.(*core_v1.Pod)
+	ip := pod.Status.PodIP
+	// Pod must have an IP, be running, and also not have the same IP as the host
+	// If a pod is a HostNetwork pod then there could be multiple with the same IP sending stats, which breaks this
+	// abstraction
+	if ip == "" ||
```
awprice: Worth simplifying this if statement?

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,378 @@
...
+	}
+
+	// Set up informers
+	indexers := cache.Indexers{
+		PodsByIPIndexName: podByIpIndexFunc,
+	}
+
+	watchCluster := a.GetBool(ParamWatchCluster)
+	customWatchOptions := func(*meta_v1.ListOptions) {}
+	if watchCluster != true {
```
awprice: `if !watchCluster {` ?

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,378 @@
...
+		meta_v1.NamespaceAll,
+		a.GetDuration(ParamResyncPeriod),
+		indexers,
+		customWatchOptions)
+
+	// TODO: we should emit prometheus metrics to fit in with the k8s ecosystem
+	// TODO: we should emit events to the k8s API to fit in with the k8s ecosystem
+
+	return &Provider{
+		ClientSet:           &clientset,
```
awprice: Is this ever used?

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,378 @@
...
+}
+
+func getSelfIPAddress() (gostatsd.IP, error) {
+	// In Kubernetes, the IP address of eth0 will be the IP address of the pod running this application
+	interfaces, err := net.Interfaces()
+	if err != nil {
+		return "", err
+	}
+	for _, i := range interfaces {
+		if i.Name != "eth0" {
```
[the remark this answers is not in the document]
mwhittington21: So the main other option would be to supply via the downwards API, but I tried to avoid needing that at all since it's a bit more of a pain to enforce.

We could just make the executive decision to never allow enriching our own pod metrics via this mechanism, which would probably be fine. Then we wouldn't need our own IP.

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,378 @@
...
+	DefaultResyncPeriod = 5 * time.Minute
+	// DefaultWatchCluster is the default watch mode for pods. By default we watch the entire cluster.
+	DefaultWatchCluster = true
+)
+
+// Provider represents a k8s provider.
+type Provider struct {
+	logger logrus.FieldLogger
+
+	ClientSet           *kubernetes.Interface
```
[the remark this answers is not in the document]
mwhittington21: Yep, good catch. I didn't use it anywhere so didn't catch that.

pkg/cloudproviders/k8s/k8s.go
mwhittington21: No, I'll remove it.

pkg/statsd/statsd_defaults_and_params.go
```diff
@@ -4,6 +4,9 @@ import (
 	"runtime"
 	"time"
 
+	"github.com/atlassian/gostatsd/pkg/cloudproviders/aws"
+	"github.com/atlassian/gostatsd/pkg/cloudproviders/k8s"
```
Jacobious52: nit: imports

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,380 @@
...
+		}
+		if len(objs) < 1 {
+			p.logger.Debugf("Could not find IP in cache, continuing")
+			instanceIPs[lookupIP] = nil
+			continue
+		}
+		if len(objs) > 1 {
+			p.logger.Warn("More than one pod in cache for IP '%s'. Using first stored", lookupIP)
+		}
+		pod := objs[0].(*core_v1.Pod)
```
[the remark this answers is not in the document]
ash2k: If something unexpected comes out from the informer, panicking is the right thing to do :)

pkg/backends/newrelic/newrelic.go
```diff
@@ -60,7 +60,7 @@ type Client struct {
 	flushType      string
 	apiKey         string
 	tagPrefix      string
-	// Options to define your own field names to support other StatsD implementations
+	// CloudProviderOptions to define your own field names to support other StatsD implementations
```
ash2k: Undo :)

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,380 @@
...
+)
+
+var (
+	// DefaultAnnotationTagWhitelist is the default ParamAnnotationTagWhitelist. Everything beginning with a standard
+	// gostatsd string is included by default.
+	DefaultAnnotationTagWhitelist = []string{
+		fmt.Sprintf("^%s", AnnotationPrefix),
+	}
+	// DefaultLabelTagWhitelist is the default ParamLabelTagWhitelist. Every label is ignored by default.
+	DefaultLabelTagWhitelist []string
```
ash2k: I think we should start making mutable things private or removing them completely (i.e. replace with a method that returns things).

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,380 @@
...
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	core_v1inf "k8s.io/client-go/informers/core/v1"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+)
+
+var (
+	// DefaultAnnotationTagWhitelist is the default ParamAnnotationTagWhitelist. Everything beginning with a standard
+	// gostatsd string is included by default.
+	DefaultAnnotationTagWhitelist = []string{
+		fmt.Sprintf("^%s", AnnotationPrefix),
```
ash2k: Just concatenate?

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,380 @@
...
+
+const (
+	// ProviderName is the name of the k8s cloud provider.
+	ProviderName = "k8s"
+	// PodsByIPIndexName is the name of the index function storing pods by IP.
+	PodsByIPIndexName = "PodByIP"
+	// AnnotationPrefix is the annotation prefix that is turned into tags by default.
+	AnnotationPrefix = "gostatsd.atlassian.com/"
+	// NodeNameEnvVar is the name of the environment variable that the k8s cloud provider will find it's node name in.
+	NodeNameEnvVar = "GOSTATSD_NODE_NAME"
```
ash2k: All our configuration comes from viper. Why environment variables?

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,380 @@
...
+	logger logrus.FieldLogger
+
+	PodsInf             cache.SharedIndexInformer
+	annotationWhitelist []regexp.Regexp
+	labelWhitelist      []regexp.Regexp
+}
+
+func (p *Provider) EstimatedTags() int {
+	// There is no real way to estimate this for k8s provider as any pod can have arbitrary labels/annotations
+	return 8
```
ash2k: Leave it at 0

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,380 @@
...
+	DefaultResyncPeriod = 5 * time.Minute
+	// DefaultWatchCluster is the default watch mode for pods. By default we watch the entire cluster.
+	DefaultWatchCluster = true
+)
+
+// Provider represents a k8s provider.
+type Provider struct {
+	logger logrus.FieldLogger
+
+	PodsInf             cache.SharedIndexInformer
```
ash2k: Can be private?

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,380 @@
...
+	annotationWhitelist []regexp.Regexp
+	labelWhitelist      []regexp.Regexp
```
ash2k: Do we need lists here? A regex can have branches `(A branch|B branch|something else)` - is that not good enough?

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,380 @@
...
+			p.logger.Warn("More than one pod in cache for IP '%s'. Using first stored", lookupIP)
+		}
+		pod := objs[0].(*core_v1.Pod)
+
+		// Turn the pod metadata into tags
+		tags := gostatsd.Tags{}
+		// TODO: deduplicate labels and annotations in their tag format, rather than overwriting
+		for k, v := range pod.ObjectMeta.Labels {
+			if tagKeyMatchesWhitelist(k, p.labelWhitelist) {
+				tags = append(tags, fmt.Sprintf("%s:%s", k, v))
```
ash2k: String concatenation is more performant than formatting. It does matter for our scale.

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,380 @@
...
+			}
+		}
+		for k, v := range pod.ObjectMeta.Annotations {
+			if !tagKeyMatchesWhitelist(k, p.annotationWhitelist) {
+				continue
+			}
+			// Annotations are often of the format "purpose.company.com/annotationName" so we want to expectedKey off the
+			// start when processing tag names
```
ash2k: I think it'd be better to use capture groups instead of making assumptions about the key structure. WDYT?

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,380 @@
...
+			tagName := k
+			splitTagName := strings.SplitN(k, "/", 2)
+			if len(splitTagName) > 1 {
+				tagName = splitTagName[1]
+			}
+			tags = append(tags, fmt.Sprintf("%s:%s", tagName, v))
```
ash2k: Concatenation please

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,380 @@
...
+	// There is no real way to estimate this for k8s provider as any pod can have arbitrary labels/annotations
+	return 8
+}
+
+// Instance returns pod details from k8s API server watches.
+// ip -> nil pointer if pod was not found.
+// map is returned even in case of errors because it may contain partial data.
+// An "instance", as far as k8s cloud provider is concerned, is an individual pod running in the cluster
+func (p *Provider) Instance(ctx context.Context, IP ...gostatsd.IP) (map[gostatsd.IP]*gostatsd.Instance, error) {
+	instanceIPs := map[gostatsd.IP]*gostatsd.Instance{}
```
ash2k: Should pre-allocate the map based on the number of IPs.

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,380 @@
...
+			// Annotations are often of the format "purpose.company.com/annotationName" so we want to expectedKey off the
+			// start when processing tag names
+			tagName := k
+			splitTagName := strings.SplitN(k, "/", 2)
+			if len(splitTagName) > 1 {
+				tagName = splitTagName[1]
+			}
+			tags = append(tags, fmt.Sprintf("%s:%s", tagName, v))
+		}
+		instanceID := fmt.Sprintf("%s/%s", pod.Namespace, pod.Name)
```
ash2k: Concat please

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,380 @@
...
+		instanceIPs[lookupIP] = &gostatsd.Instance{
```
ash2k: Not in this PR but I wonder if we should change this whole "API" to use values and use the `ID` field as the flag? Making some garbage for the collector to collect here...

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,380 @@
...
+	a.SetDefault(ParamKubeconfigContext, DefaultKubeconfigContext)
+	a.SetDefault(ParamKubeconfigPath, DefaultKubeconfigPath)
+
+	// Set up the k8s client
+	var restConfig *rest.Config
+	var err error
+	kubeconfigPath := a.GetString(ParamKubeconfigPath)
+	kubeconfigContext := a.GetString(ParamKubeconfigContext)
+	if kubeconfigPath != "" {
+		configOverrides := &clientcmd.ConfigOverrides{ClusterInfo: clientcmdapi.Cluster{Server: ""}}
```
ash2k: `ClusterInfo` does not need to be set? It's a noop, right?

pkg/cloudproviders/k8s/k8s.go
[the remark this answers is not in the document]
mwhittington21: I would much rather the app shut down and restart with a fresh cache if something unexpected comes out from the informer. It's also much more obvious that something rather bad happened.

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,380 @@
...
+		customWatchOptions)
+
+	// TODO: we should emit prometheus metrics to fit in with the k8s ecosystem
+	// TODO: we should emit events to the k8s API to fit in with the k8s ecosystem
+
+	return &Provider{
+		PodsInf:             podsInf,
+		annotationWhitelist: annotationWhitelist,
+		labelWhitelist:      labelWhitelist,
+		logger:              options.Logger,
```
ash2k: Prefer setting fields in literals in the order they are defined in the struct for consistency reasons.

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,380 @@
...
+	var wg wait.Group
+	defer wg.Wait()
+
+	informerCtx, informerCancel := context.WithCancel(ctx)
+	defer informerCancel()
+
+	wg.StartWithChannel(informerCtx.Done(), p.PodsInf.Run) // Fork
+
+	// Work until asked to stop
+	<-ctx.Done()
```
ash2k: Replace everything here with `p.PodsInf.Run(ctx.Done())`? 😁

pkg/cloudproviders/k8s/k8s_test.go
```diff
@@ -0,0 +1,283 @@
...
+	go func() {
+		r.Run(ctx) // fork
+	}()
+
+	test(t, &testFixtures{
+		fakeClient:    fakeClient,
+		cloudProvider: cloudProvider,
+		podsWatch:     podsWatch,
+	})
+	return nil
```
ash2k: Could just call `require.*` and `t.FailNow()`/etc methods and then it would not need to return anything.

pkg/statsd/handler_cloud.go
```diff
@@ -41,6 +50,92 @@ type CacheOptions struct {
...
 }
 
+// CloudHandlerFactory creates a CloudHandler based on some predefined options.
+type CloudHandlerFactory struct {
+	cloudProviderName string
+	version           string
+	CloudProvider     gostatsd.CloudProvider
+	logger            logrus.FieldLogger
+	Limiter           *rate.Limiter
+	CacheOptions      *CacheOptions
```
ash2k: This doesn't need to be a pointer, better as just a value.
Also all fields should be private OR make them all public and remove the constructor, I guess.

pkg/statsd/handler_cloud.go
```diff
@@ -41,6 +50,92 @@ type CacheOptions struct {
...
+}
+
+// InitCloudProvider initialises the cloud provider given some already parsed defaults.
+// This is mainly split out to enable testing the config parsing separately from the cloud providers.
+func (f *CloudHandlerFactory) InitCloudProvider(v *viper.Viper) error {
+	options := gostatsd.CloudProviderOptions{Viper: v, Logger: f.logger, Version: f.version}
+	switch f.cloudProviderName {
+	case aws.ProviderName:
+		// AWS has no special options required
+		break
```
ash2k: Don't need `break` here :)

pkg/statsd/handler_cloud.go
```diff
@@ -226,13 +321,25 @@ func (ch *CloudHandler) Run(ctx context.Context) {
 		logger:        ch.logger,
 	}
 
-	var wg wait.Group
-	defer wg.Wait() // Wait for lookupDispatcher to stop
+	// Stager will perform ordered, graceful shutdown. Stage by stage in reverse startup order.
+	stgr := stager.New()
+	defer stgr.Shutdown() // Wait for CloudProvider and lookupDispatcher to stop
 
 	ctx, cancel := context.WithCancel(ctx)
```
ash2k: This extra context is not needed if you use stager here.

pkg/statsd/handler_cloud_lookup.go
```diff
@@ -47,7 +47,7 @@ func (ld *lookupDispatcher) run(ctx context.Context) {
 			if err != context.Canceled && err != context.DeadlineExceeded {
 				// This could be an error caused by context signaling done.
 				// Or something nasty but it is very unlikely.
-				ld.logger.Warnf("Error from limiter: %v", err)
+				ld.logger.Warnf("Error from Limiter: %v", err)
```
ash2k: Should probably revert?

pkg/statsd/statsd_defaults_and_params.go
```diff
@@ -32,6 +35,43 @@ var DefaultTags = gostatsd.Tags{}
 // DefaultInternalTags is the default list of additional tags on internal metrics
 var DefaultInternalTags = gostatsd.Tags{}
 
+// DefaultCloudProviderCacheValues contains the default cache values for each cloud provider.
+var DefaultCloudProviderCacheValues = map[string]CacheOptions{
+	k8s.ProviderName: {
+		// Basically the CacheRefreshPeriod is the granularity at which we care about batching data.
+		// If 5 requests come in inside 50 milliseconds, and this is set to 50 milliseconds, we only need to look
+		// up the informer once, and do the regexes once. This saves us resources at the cost of data precision.
+		CacheRefreshPeriod:        50 * time.Millisecond,
```
ash2k: I think we should just leave the defaults for now and refactor the things to avoid this problem all together. I'm working on that refactoring in a separate branch to merge later.

pkg/cloudproviders/k8s/k8s.go
mwhittington21: I'm all for encapsulation, I was just following the existing standard. Separate PR to encapsulate?

ash2k:
> The above mentioned waiting for pod IPs to arrive before forwarding metrics
I'm working on the refactoring that will allow us to do this. Once we merge this PR I'll hopefully have the refactoring ready. Will need to rebase and then open a PR.

pkg/cloudproviders/k8s/k8s.go
mwhittington21: It's a Kubernetes thing. I don't think we can easily do viper for this.

https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/

I thought very hard about how I could put this into Viper before I committed to this. The only way I can think of would be to put an initialisation script into the container that would read the environment variable and insert it into the config.toml instead of a placeholder. This would put additional burden on the operator as the binary itself is no longer standalone for using inside k8s.

pkg/cloudproviders/k8s/k8s.go
mwhittington21: It was separated because labels and annotations often have very different formats, and to logically allow two places for people to specify their regexes in the Viper config.

Annotation example: `gostatsd.atlassian.com/mytag=value`
Label example: `app=myapp`

pkg/cloudproviders/k8s/k8s.go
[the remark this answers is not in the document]
awprice: I'm aware it's going to add additional complexity, but since we will have the node name, could we grab the IP from the node object in the cluster?

pkg/cloudproviders/k8s/k8s.go
mwhittington21: That's a good idea, I actually hadn't thought about that at all. I'll have a go at implementing that.

pkg/cloudproviders/k8s/k8s.go
mwhittington21: Yeah I think that would fit in with the cache changes, or close to.

pkg/cloudproviders/k8s/k8s.go
mwhittington21: After testing it seems you're correct. Will remove.

pkg/cloudproviders/k8s/k8s.go
mwhittington21: My goroutine-fu is still rather bad 😅. Thanks!

pkg/cloudproviders/k8s/k8s.go
ash2k: Up to you. Separate PR is totally fine. 👍

pkg/cloudproviders/k8s/k8s.go
ash2k: Ok, I see. Let's keep it like this then. We can always change it.

pkg/cloudproviders/k8s/k8s_test.go
mwhittington21: I did this while I was working out how to wrap tests properly for table driven testing. With the current setup this does work much better, I'll change it. Changing it does complicate TestWatchNodeFailsNoNodeName a tiny bit though.

pkg/cloudproviders/k8s/k8s.go
ash2k: Sorry, let me rephrase that. Can we do this instead:
```go
	annotationFilter regexp.Regexp
	labelFilter      regexp.Regexp
```

pkg/cloudproviders/k8s/k8s.go
ash2k: However I think it'd be better to not expose the function to get the value and only use it in the factory. This will prevent spreading the knowledge of where this value is coming from all over the place.

pkg/statsd/handler_cloud.go
mwhittington21: I think we do - unless I just remove that example clause. Otherwise if someone is using the AWS provider then we'll override them with k8s options. These options are harmless for now but it could introduce a bug down the road.

The two options I see here are:

1. keep it the way it is
2. remove the AWS clause completely since it's just a big comment, basically

pkg/cloudproviders/k8s/k8s.go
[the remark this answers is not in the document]
ash2k: I think it should either be via downward API (my preferred way) or via a configuration field/env var. Fetching it from `Node` objects will require adding extra permissions to `ServiceAccount`. It's nice to keep it minimal.

pkg/statsd/handler_cloud_lookup.go
mwhittington21: Done. Still getting used to IntelliJ's refactoring in Go 😖

pkg/statsd/statsd_defaults_and_params.go
mwhittington21: I disagree. The purpose of these cache overrides is that we can use what's here indefinitely if we absolutely have to. If we use the cache values of the AWS one then if a pod leaves and arrives again with the same IP quickly then we'll cache metrics for far too long.

I don't see any harm in splitting this, basically.

tiedotguy: There's a `time.Ticker` running at this interval, is that acceptable? It will also flush the entire CP cache every time, effectively disabling it, but also creating memory churn and garbage as it tries to repopulate it. This is fine on a local instance, but may hurt something which monitors an entire cluster.

The batching is in `lookupDispatcher`, and it waits for more IPs for 10ms (hard coded) after the first IP comes in, then processes what it has.  Generally the CP caching and lookup behavior can be pretty tough to follow because of all the batching and async stuff :(

To be clear:
- `CacheRefreshPeriod`: how often the `CloudHandler` will check if entries should be evicted or refreshed
- `CacheEvictAfterIdlePeriod`: when the refresh runs, if an entry hasn't been accessed in this interval, evict it
- `Cache[Negative]TTL`: if an entry is kept, and it's been this long since it was last created/refreshed (**not** last time the cached value was accessed), refresh it from the backing CP.

ash2k: ok 👍

pkg/statsd/handler_cloud.go
ash2k: https://play.golang.org/p/pelegl1WgOt

pkg/cloudproviders/k8s/k8s.go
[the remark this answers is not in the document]
mwhittington21: Good thoughts about RBAC, I hadn't thought that far ahead.

pkg/cloudproviders/k8s/k8s.go
mwhittington21: Definitely agree with you about not spreading the knowledge beyond the factory.

Actually there is one possible way we could fix this. We could allow the config file to point to a file that contains the node name, and then create that file with a value from the downwards api. Both ways kind of depart from viper a bit though.

Do you think this would be a better fit @ash2k ?

pkg/cloudproviders/k8s/k8s.go
mwhittington21: Yeah we could do that. I'll make it happen.

pkg/cloudproviders/k8s/k8s.go
ash2k: I think it's better to keep it as is.

pkg/cloudproviders/k8s/k8s_test.go
```diff
@@ -0,0 +1,283 @@
...
+				}
+
+				ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
+				defer cancel()
+
+				instanceData, err := fixtures.cloudProvider.Instance(ctx, ipAddr)
+				require.NoError(ttt, err)
+
+				instance := instanceData[ipAddr]
+				assert.NotNil(ttt, instance)
```
tiedotguy: I generally favor `require` always, however in this case you're depending on `instance` being not-nil (otherwise you get the panic), so this should absolutely be `required`.

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,380 @@
...
+	DefaultLabelTagWhitelist []string
+)
+
+const (
+	// ProviderName is the name of the k8s cloud provider.
+	ProviderName = "k8s"
+	// PodsByIPIndexName is the name of the index function storing pods by IP.
+	PodsByIPIndexName = "PodByIP"
+	// AnnotationPrefix is the annotation prefix that is turned into tags by default.
+	AnnotationPrefix = "gostatsd.atlassian.com/"
```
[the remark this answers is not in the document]
ash2k: Good catch! I think rather than making this constant a regex it should be escaped where it is used as part of a regex using https://golang.org/pkg/regexp/#QuoteMeta
WDYT?

pkg/cloudproviders/k8s/k8s.go
tiedotguy: I disagree with both these approaches. If it's needed, it's on the service owner to provide it to viper. A wrapper script for config is trivial. By definition we aren't a standalone binary in a k8s environment, because there's all the k8s stuff to define what's even being ran.

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,380 @@
...
+	}
+
+	podsInf := core_v1inf.NewFilteredPodInformer(
+		clientset,
+		meta_v1.NamespaceAll,
+		a.GetDuration(ParamResyncPeriod),
+		indexers,
+		customWatchOptions)
+
+	// TODO: we should emit prometheus metrics to fit in with the k8s ecosystem
```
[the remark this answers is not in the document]
ash2k: Can statser use a separate/different backend than the main pipeline? That would be ideal, I guess.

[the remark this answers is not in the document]
tiedotguy: Yep, currently there's only null, internal (sends via the pipeline, but has a different set of configurable tags), and logger implemented, but it would be fairly trivial to add a prom statser I expect, since you're dealing with pre-aggregated data (an actual prom backend would be more complicated because it's aggregated).

The only down side is the configuration around all that is still very minimal (it's in the global config namespace, there's no facility for multiple statsers, etc)

tiedotguy:
The test failures seem to be timing racy, not data racy. I think the main thing `test-race` is doing is introducing slowdown with its overhead.  If I run non-race with `-count 1000`, it triggers, although far less frequently.

I'd suggest giving the whole thing a bit of a soak test with `-count 1000` (and maybe `-race -count 1000`) once you figure out the failure.  The panic in the build failure is masking the underlying issue, which is `assert` vs `require`. The assert should print, but I guess the panic is causing it to not output.

pkg/cloudproviders/k8s/k8s.go
mwhittington21: @tiedotguy that's a fair stance. We could use docs to show usage patterns, but I still disagree with using a wrapper script for config unless we're providing it inside a k8s-built container for the user.

How about we allow specifying the node name in config, which overrides environment variable?

pkg/statsd/handler_cloud.go
mwhittington21: Ah. My bad for assuming, sorry Mikhail!

pkg/cloudproviders/k8s/k8s.go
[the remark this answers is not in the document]
mwhittington21: This one is interesting and I might need to have a conversation with you in person about this.

I absolutely agree with you, when I was doing this I got all sorts of alarm bells in my head. However, it is by far the simplest and least error prone way to supply this information to the running binary from a k8s perspective.

As discussed under @ash2k's comment above, we could make a script that takes this variable and inserts it into the Viper config via find/replace, but if we aren't providing that for users I feel that it makes this feature especially confusing to use. Maybe it's just a case of needing the right docs, but maybe this approach does have legs. It does seem weird only getting one or two values via environment variable while everything else comes from Viper.

If I could read the environment variable via Viper and allow overriding from the config, then maybe we could get past this. I'll investigate that.

pkg/cloudproviders/k8s/k8s.go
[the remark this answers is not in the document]
mwhittington21: Great catch. I think keeping the constant a string for ease of reading and then escaping it on-use is the way. Thanks both of you!

pkg/cloudproviders/k8s/k8s.go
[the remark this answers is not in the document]
mwhittington21: I agree with what you're saying here. To clarify, the context of this TODO is just that to fit into the k8s ecosystem properly we should emit Prometheus metrics. I think that the approach you mention is the way to do it, which is why I haven't tackled it here as it's a bit of a big job.

Basically it would just be adding a prometheus backend/statser and then exposing that to the cluster so that k8s operators can monitor it the same way as everything else in the ecosystem 🙂

pkg/cloudproviders/k8s/k8s.go
tiedotguy: I would say make it required (log a descriptive error then fail, rather than `SetDefault`), and document it with a small snippet of what the config would look like. People should be reading documentation anyway, and the "fail fast" tells them something is missing even if they don't.

We can provide a sample script easily enough in the repo - I think it would be cool to include a full kubernetes yaml file as well to bring everything up, and this can just be in there with it.

I didn't have any definite plans to head in to the office, but I can swing past on Friday if you like (if you need this earlier I can make that happen too, Friday is just the typical day because it's a bit more relaxed)

pkg/statsd/statsd_defaults_and_params.go
mwhittington21: Yeah I had real trouble diving into this codebase in general due to all the async stuff. I think I've done OK but I'm really relying on you and Mikhail to pick out any weird edge cases or performance things for me since I don't have a lot of context.

Perhaps negative TTL and CacheRefreshPeriod are the things I need to tweak here then? Something like `CacheRefreshPeriod=1s`, `CacheEvictAfterIdlePeriod=60s`, `CacheNegativeTTL=1s`. This would keep everything refreshed and evict things after a while.

I might make a test to try and get the situation to happen at AWS cache values and then try and fix it with more aggressive cache values to prove the fix works.

We can probably deal with metrics not having sub second granularity if a pod somehow manages to stop and another one start with the same IP immediately (very rare edge case).

WDYT?

pkg/cloudproviders/k8s/k8s_test.go
mwhittington21: I learned that the hard way before I got around to reading this 😔.
✅ Always `require` on things that will break further down.

pkg/cloudproviders/k8s/k8s.go
mwhittington21: F2F would probably be the easiest, Friday is fine 🙂

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,380 @@
...
+	podSelfIP, err := getSelfIPAddress()
+	if err != nil {
+		return gostatsd.UnknownIP, err
+	}
+	return podSelfIP, nil
+}
+
+// NewProviderFromOptions returns a new k8s provider.
+func NewProviderFromOptions(options gostatsd.CloudProviderOptions) (gostatsd.CloudProvider, error) {
+	a := util.GetSubViper(options.Viper, "k8s")
```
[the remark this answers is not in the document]
mwhittington21: I just copied what was in the AWS provider. It only now hit me that `a` stands for AWS...so should I use `k` instead?

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,380 @@
...
+			returnErr = err
+			continue
+		}
+		if len(objs) < 1 {
+			p.logger.Debugf("Could not find IP in cache, continuing")
+			instanceIPs[lookupIP] = nil
+			continue
+		}
+		if len(objs) > 1 {
+			p.logger.Warn("More than one pod in cache for IP '%s'. Using first stored", lookupIP)
```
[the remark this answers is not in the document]
mwhittington21: This is how you know I haven't worked in Observability before 🙃

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,380 @@
...
+	annotationWhitelist []regexp.Regexp
+	labelWhitelist      []regexp.Regexp
+}
+
+func (p *Provider) EstimatedTags() int {
+	// There is no real way to estimate this for k8s provider as any pod can have arbitrary labels/annotations
+	return 8
+}
+
+// Instance returns pod details from k8s API server watches.
```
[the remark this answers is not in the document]
mwhittington21: A Pod shares a network namespace, so all containers in the pod share an IP address.

tiedotguy:
I've been poking at the build failure (I love a good race condition!), and haven't been able to identify a good fix for it. My approach would be to replace...
```
  for _, p := range testCase.pods {
    fixtures.podsWatch.Add(p)
  }
```

with
```
  for _, p := range testCase.pods {
    fixtures.podsWatch.Add(p)
  }
  for i := 1; i < 100 && len(testCase.pods) < podsInCache(); i++ {
    time.Sleep(10 * time.Millisecond)
  }
  require.Equal(ttt, len(testCase.pods), podsInCache())
```

Implementing `podsInCache()` is left as an exercise for the reader, because damned if I can figure out how to get it :(

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,312 @@
...
+}
+
+func stringListToRegex(strList []string) (*regexp.Regexp, error) {
+	if len(strList) == 0 {
+		// This regex should always fail to match anything
+		return regexp.Compile("(k^)")
+	}
+
+	combinedRegexes := "("
+	for i, regexStr := range strList {
```
tiedotguy: You can use `strings.Join(strList, "|")`, it does the whole measure + pre-allocation and such.

pkg/statsd/statsd_defaults_and_params.go
tiedotguy: I think a low-ish (15 second) `CacheRefreshPeriod` is fine, keep `CacheEvictAfterIdlePeriod` at default (DefaultCacheEvictAfterIdlePeriod = 10 min), and keep the TTLs at 1ms (or even 0ms).

Since there's no network cost to doing a refresh, and we're not refreshing multiple times a second, it should be fine to refresh the cache every time, even though there's a bit of a hit on the CPU.  I doubt it would be noticeable.

pkg/cloudproviders/k8s/k8s.go
tiedotguy: Can do!

pkg/cloudproviders/k8s/k8s.go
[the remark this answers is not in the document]
tiedotguy: I raised a pair of issues, for backend and internal statser :)

mwhittington21:
> I've been poking at the build failure (I love a good race condition!), and haven't been able to identify a good fix for it. My approach would be to replace...
> 
> ```
>   for _, p := range testCase.pods {
>     fixtures.podsWatch.Add(p)
>   }
> ```
> 
> with
> 
> ```
>   for _, p := range testCase.pods {
>     fixtures.podsWatch.Add(p)
>   }
>   for i := 1; i < 100 && len(testCase.pods) < podsInCache(); i++ {
>     time.Sleep(10 * time.Millisecond)
>   }
>   require.Equal(ttt, len(testCase.pods), podsInCache())
> ```
> 
> Implementing `podsInCache()` is left as an exercise for the reader, because damned if I can figure out how to get it :(

This worked great! Although I'm tired and didn't realise that the logic in the `<` statement in the if was the wrong way around for far too long :(. Using `podsInCache() < len(testCase.pods)` worked fine.

tiedotguy:
Nice! I'm not a fan of busy loops, but I'm even less of a fan of a 6 core system doing a `sleep(1 second)` just because the build system has 1/3 of a CPU...

Derp, sorry about the logic mixup!

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,304 @@
...
+		annotationWhitelist: annotationWhitelist,
+		labelWhitelist:      labelWhitelist,
+	}, nil
+}
+
+func (p *Provider) Run(ctx context.Context) {
+	p.podsInf.Run(ctx.Done())
+}
+
+func stringListToRegex(strList []string) (*regexp.Regexp, error) {
```
ash2k: When I suggested to make it a single regex I meant that the user should specify a single regex. We can take in a single string, simplifying our code and also making it more performant (1 match vs N), without limiting the user's ability to achieve what they want. I'm sorry I was not clear enough.

mwhittington21: Will capture groups actually work with this approach though? I remember you mentioned we could use capture groups to simplify how the tag is captured and processed, but I'm not sure how we would reference these in a multi-regex.

ash2k: When capturing is not needed/desired, a non-capturing group can be used. See https://github.com/google/re2/wiki/Syntax

Example: `(?:one|two|three)/(\w+)`

ash2k: We can also just mandate the use of a named capturing group. That might be a better option, actually. Who knows, maybe later we'll want to capture more things and then depending on order becomes problematic.

pkg/cloudproviders/k8s/k8s.go
tiedotguy: I was re-reading the thread with fresh eyes, trying to figure out what it is I didn't like about it. Essentially it was gostatsd changing its configuration model to conform, but I wasn't really considering it from the k8s API side.  The solution is ultimately a bridge between the two, which is what you suggested here:

> Actually there is one possible way we could fix this. We could allow the config file to point to a file that contains the node name, and then create that file with a value from the downwards api. Both ways kind of depart from viper a bit though.

Looking at it from that perspective, I've changed my opinion and think this is the right approach, however I would suggest using an env var defined by Viper (with a sane default), rather than a file, mainly because It's simpler code.

Thoughts?

mwhittington21: Yeah that sounds good to me. I'll try and implement it. To clarify, we're having a Viper defined env var that can also be overridden by config in the `.toml` right?

tiedotguy: Yeah, just like..

```
v.SetDefault("k8s-host-name-var", "GOSTATSD_BLAH")
...
hostName := GetEnv(v.GetString("k8s-host-name-var"))
```

... or something.

mwhittington21: I was thinking something more like
```
// replacer is to allow this to be a k8s sub option that conforms to env var naming standards
replacer := strings.NewReplacer(".", "_")
v.SetEnvKeyReplacer(replacer)
// this default could be sensible, or empty and fail-fast
v.SetDefault("k8s.node-name", "")
v.AutomaticEnv()

// this should retrieve from env var or config, depending on what's set and whatever the override order is
nodeName := v.Get("k8s.node-name")
```

Thoughts?

mwhittington21: The one downside is that particular variable in the config file would have to be named with `_` instead of `-` because we're using our replacer on the `.` to `_`

mwhittington21: Never mind, this probably won't work since there's a top level replacer already. I'll see what I can do to make something good within the constraints.

mwhittington21: Got something working here, just need to write tests for it now 🙂

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,380 @@
...
+	// Get our pod's IP address from the network stack
+	podSelfIP, err := getSelfIPAddress()
+	if err != nil {
+		return gostatsd.UnknownIP, err
+	}
+	return podSelfIP, nil
+}
+
+// NewProviderFromOptions returns a new k8s provider.
+func NewProviderFromOptions(options gostatsd.CloudProviderOptions) (gostatsd.CloudProvider, error) {
```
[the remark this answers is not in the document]
mwhittington21: I've refactored this to have config processing happen in the `NewProviderFromViper` method, and any initialisation happen in the `NewProvider` method. The tests have much cleaner separation of concerns now.

pkg/cloudproviders/k8s/k8s.go
mwhittington21: Implemented capture groups without too much trouble. All it will need is some solid docs examples so people don't get confused.

pkg/cloudproviders/k8s/k8s.go
mwhittington21: I've gone with a sane default. If the regex matches and the "tag" capture group matches, use the tag. Otherwise use the regex match.

pkg/statsd/handler_cloud.go
```diff
@@ -226,13 +321,22 @@ func (ch *CloudHandler) Run(ctx context.Context) {
 		logger:        ch.logger,
 	}
 
-	var wg wait.Group
-	defer wg.Wait() // Wait for lookupDispatcher to stop
+	// Stager will perform ordered, graceful shutdown. Stage by stage in reverse startup order.
+	stgr := stager.New()
+	defer stgr.Shutdown() // Wait for CloudProvider and lookupDispatcher to stop
+
+	stage := stgr.NextStageWithContext(ctx)
```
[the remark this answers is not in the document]
mwhittington21: Thanks. Still really struggling with all the goroutine stuff.

mwhittington21:
@ash2k @tiedotguy

Okay I've addressed all of the proposed changes above that weren't to be moved into other PRs. Let me know if there's anything else you'd like to see.

I'll start working on a docs PR now, branched off this one.

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,348 @@
...
+// NeverMatchRegex is a regex string that will never match anything
+const NeverMatchRegex = "k^"
+
+var (
+	// DefaultAnnotationTagRegex is the default ParamAnnotationTagRegex. Everything beginning with a standard
+	// gostatsd string is included by default. The default sets the tag name as the part after the "/", for example
+	// gostatsd.atlassian.com/tag1 would return "tag1" as the tag name after matching.
+	DefaultAnnotationTagRegex = "^" + regexp.QuoteMeta(AnnotationPrefix) + DefaultTagCaptureRegex
+	// DefaultLabelTagRegex is the default ParamLabelTagRegex. Every label is ignored by default.
+	DefaultLabelTagRegex = NeverMatchRegex // regex will never match anything
```
ash2k: I wonder if we should use an empty string as such flag? Has two benefits:
- Less ugly than `k^`
- Simple string check (or `nil` check - see below) is likely much faster than running a regex, even a simple one.
- The check can guard the whole loop on labels - no need to try to match each label if we already know none will match. Big performance gain.

Can be passed as a `nil` pointer to `Regexp` to indicate that it's `off`.

WDYT?

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,348 @@
...
+
+const (
+	// ProviderName is the name of the k8s cloud provider.
+	ProviderName = "k8s"
+	// PodsByIPIndexName is the name of the index function storing pods by IP.
+	PodsByIPIndexName = "PodByIP"
+	// AnnotationPrefix is the annotation prefix that is turned into tags by default.
+	AnnotationPrefix = "gostatsd.atlassian.com/"
+	// DefaultTagCaptureRegex is a regex subexpression to capture from now until the end of the regex as the tag name.
+	DefaultTagCaptureRegex = "(?P<tag>.*)$"
```
ash2k: I suggest to make `tag` a constant to avoid hardcoding it in more than one place.

ash2k: If you end it with `$`, I'd suggest to also start with `^` to make the intent clearer.

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,348 @@
...
+		customWatchOptions)
+
+	// TODO: we should emit prometheus metrics to fit in with the k8s ecosystem
+	// TODO: we should emit events to the k8s API to fit in with the k8s ecosystem
+
+	return &Provider{
+		logger:          logger,
+		podsInf:         podsInf,
+		annotationRegex: regexp.MustCompile(annotationRegex),
+		labelRegex:      regexp.MustCompile(labelRegex),
```
ash2k: I'd change this method to accept the regex type, not strings. "Accept direct dependencies" is the "rule" of dependency inversion/injection. It's the job of the caller to construct them.

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,348 @@
...
+	v.SetDefault(ParamUserAgent, DefaultUserAgent+"/"+version)
+	v.SetDefault(ParamWatchCluster, DefaultWatchCluster)
+}
+
+// getTagNameFromRegex gets a tag name from the regex. This is either the entire input string, or a subset matching
+// the "tag" capture group if it exists and matched. Returns "" if the regex matched nothing.
+func getTagNameFromRegex(re *regexp.Regexp, s string) string {
+	match := re.FindStringSubmatch(s)
+
+	var matches = make(map[string]string, len(match))
```
ash2k: We don't need this map do we? Can just do `switch` inside of the loop and look for interesting group names. WDYT?

pkg/cloudproviders/k8s/k8s_test.go
```diff
@@ -0,0 +1,341 @@
...
+		v.GetString(ParamLabelTagRegex),
+	)
+	require.NoError(t, err)
+
+	r, ok := cloudProvider.(gostatsd.Runner)
+	require.True(t, ok)
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	go func() {
+		r.Run(ctx) // fork
```
ash2k: This is fine, but FYI `go r.Run(ctx)` would do the same.

pkg/statsd/handler_cloud.go
```diff
@@ -33,14 +37,89 @@ func (ih *instanceHolder) lastAccess() int64 {
 	return atomic.LoadInt64(&ih.lastAccessNano)
 }
 
-// CacheOptions holds cache behaviour configuration.
+// cacheOptions holds cache behaviour configuration.
```
ash2k: Revert :)

pkg/util/viper.go
```diff
@@ -1,13 +1,33 @@
...
 
+// EnvPrefix is the prefix of the inspected environment variables.
+const EnvPrefix = "GSD" //Go Stats D
+
 func GetSubViper(v *viper.Viper, key string) *viper.Viper {
 	n := v.Sub(key)
 	if n == nil {
 		n = viper.New()
 	}
+	InitViper(n, key)
```
tiedotguy: I'm so glad `GetSubViper` was factored out :)

pkg/cloudproviders/k8s/k8s.go
mwhittington21: If we're concerned about GC then yeah I can change it for that reason. It does complicate the code a little though since there's two slices in question and we'd have to loop through them in reverse to find the subexp match before the entire regex.

pkg/cloudproviders/k8s/k8s.go
mwhittington21: The intent was really just to be a reusable thing to capture a tag at any point inside a regex. I split it out from the main regex for documentation purposes really. I will remove the $ and just add that into the combination though to make it more generic.

pkg/cloudproviders/k8s/k8s.go
mwhittington21: Makes sense, I'll do that.

mwhittington21:
Okay I've addressed the latest batch of comments and fixed some more bugs / cleaned things up a bit.

mwhittington21:
Added docs to the PR.

examples/cloudproviders/k8s/daemonset.yaml
```diff
@@ -0,0 +1,116 @@
...
+    internal-tags=['in_cluster:true']
+    statser-type='internal'
+    flush-interval='1s'
+    bad-lines-per-minute=1000
+    metrics-addr=':8125'
+    filters='no-hostname'
+
+    [k8s]
+    # Include any annotation beginning with myapp.company.com as a tag, using the part after the / as the tag name
+    annotation-tag-regex = '^myapp.company.com/(?P<tag>.*)$'
```
ash2k: Should probably escape the dots :)

examples/cloudproviders/k8s/rbac.yaml
```diff
@@ -0,0 +1,26 @@
...
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: gostatsd
+rules:
+  - apiGroups: [""]
+    resources:
+      - pods
+    verbs:
+      - get
```
ash2k: If we only use watch then we don't need get.

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,375 @@
...
+		logger:          logger,
+		podsInf:         podsInf,
+		factory:         factory,
+		annotationRegex: annotationRegex,
+		labelRegex:      labelRegex,
+	}, nil
+}
+
+func (p *Provider) Run(ctx context.Context) {
+	go p.factory.Start(ctx.Done()) // fork
```
ash2k: The factory is annoying because it's impossible to wait until all the goroutines it spawns finish. And the reason for that is that `Start()` does not block, it spawns informers' `Run()` methods in goroutines and returns control. 
I suggest you directly construct the informer and use it here OR if you decide to stick with the factory at least remove that `go` statement as it's not needed.

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,375 @@
...
+	// wait for the initial synchronization of the local cache.
+	// TODO: why is the error message firing for our tests?
```
ash2k: Perhaps you don't have the watch part mocked properly? IIRC some extra plumbing may be necessary for it to work. Smith probably has an example of this working.

pkg/statsd/handler_cloud.go
```diff
@@ -226,13 +305,21 @@ func (ch *CloudHandler) Run(ctx context.Context) {
...
+
+	stage := stgr.NextStage()
 
-	ctx, cancel := context.WithCancel(ctx)
-	defer cancel() // Tell lookupDispatcher to stop
+	// If the cloud provider implements runnable, it has things to do. It must start before anything that
+	// accesses it
+	if ch.cloud != nil {
+		if r, ok := ch.cloud.(gostatsd.Runner); ok {
+			r.Run(ctx) // synchronous - must fork inside if necessary
```
ash2k: By convention `Run()` of the `Runner` interface is a blocking method. If should be launched in a stage of stager.

examples/cloudproviders/k8s/rbac.yaml
mwhittington21: True, I'll remove it.

pkg/cloudproviders/k8s/k8s.go
mwhittington21: I actually fixed this I just forgot to remove the TODO. My bad.

pkg/cloudproviders/k8s/k8s.go
mwhittington21: What I've done here is make the Run() synchronous (i.e. we don't call it with a goroutine when spawning a cloud provider - see handler_cloud.go:318). This way we can asynchronously start whatever we want inside this context, and block until the cache returns. This ensures that the lookup can never happen before the k8s provider is ready, and we'll actually get a good debugging trace if the cache won't populate for some reason.

WDYT about the approach? It did feel a little weird changing the delineation point for asynchronous calls.

pkg/statsd/handler_cloud.go
mwhittington21: I get that it is convention but to allow waiting for the cache in a nice, cloud provider agnostic way, I thought tying in the cache wait would allow startup to happen in a nicer order. If you can think of a better way to achieve this I would love to change this because I agree on the convention.

pkg/cloudproviders/k8s/k8s.go
ash2k: `Run()` is supposed to block until the context is cancelled and when it returns, the "thing" is not supposed to be running and hence is not supposed to be used (this is not documented on the interface, sorry). In this PR the semantics of the contract are being changed and only in a single place. Rather than introducing an exception (and an undocumented one especially), maybe introduce another method to the `CloudProvider` interface that allows waiting for the provider to become ready. Something like `WaitUntilReady(context.Context) error` that blocks until provider is ready to be used. This is explicit and clear.
On the other hand, when gostatsd is redeployed all producers are not going to be waiting as there is no backpressure in UDP. And even for TCP, the socket hasn't been opened at this point yet so...
So regardless of what we do the metrics are either going to be dropped or sent untagged. I guess we have that choice here. We can design the thing to block/delay metrics processing for up to N seconds and then either drop metrics for which the tags couldn't be obtained or send them though down the pipeline. Can be configurable even.
So far the approach used has been to block/delay metrics for a very short while (while cache is populated for the IP) or not block at all and then push them further. I think we should go one step further and control how much blocking/waiting we want. We'll be able to do it after my refactoring (not in the first PR). If we agree on this design we can just leave things without any blocking/waiting and just do the smarts later.

p.s. another thing to consider - what if informer has not yet synced fully but already has the needed information? If done via `WaitForCacheToSync()` things will be blocked for longer than necessary. Via blocking index-like functionality we can have smarter waiting rather than all or nothing.

mwhittington21: I think we should just leave it and do the smarts later, as you say. Totally agree on the contract stuff, thanks for clarifying.

The main thing that was attracting me to waiting for the cache sync was debugging, since if the cache does not initially sync then something is clearly wrong. I wonder if there's a way to surface that sort of error nicely but asynchronously. I guess we could put a context with timeout in the run method, but just not block on it. Then if it times out we can print an error message saying it took a long time to sync the cache and maybe there is a problem.

Thoughts?

ash2k: I think the initial sync problems vs sync problems that may happen later shouldn't be viewed as separate things. If what we have here works we can keep it for now and I'll refactor this in my PR.

Or add a method to the interface. That is a nice way to surface problems with initial sync.

CLOUDPROVIDERS.md
```diff
@@ -0,0 +1,98 @@
...
+--------------------
+Cloud providers are a way to automatically enrich metrics with metadata from a cloud vendor.
+
+Cloud providers must be configured through the usage of a configuration file (toml, yaml and json are supported), passed via
+`--config-path`.
+
+There are currently two supported cloud providers:
+
+* `aws` which retrieves tags from AWS instance tags via AWS API calls.
+* `k8s` which retrieves tags from pod labels and annotations.
```
tiedotguy:
```suggestion
* `k8s` which retrieves tags from kubernetes pod labels and annotations.
```

CLOUDPROVIDERS.md
```diff
@@ -0,0 +1,98 @@
...
+node-name = ''
+resync-period = '5m'
+user-agent = 'gostatsd'
+watch-cluster = true
+```
+
+The configuration settings are as follows:
+- `kube-api-qps`: the number of queries per second gostatsd can make to the kubernetes API server before being rate limited
+- `kube-api-burst`: the number of queries per second gostatsd can burst above the limit when necessary
+- `annotation-tag-regex`: a regex that is compared to every pod's annotations. Any matching annotations matching the
```
tiedotguy:
```suggestion
- `annotation-tag-regex`: a regex that is compared to every pod's annotations. Any annotations matching the
```

CLOUDPROVIDERS.md
```diff
@@ -0,0 +1,98 @@
...
+regex have their value used as a metric tag value. The key of the metric is either the entire annotation key, or a subset
```
tiedotguy:
```suggestion
regex have their value used as a metric tag value. The key of the metric tag is either the entire annotation key, or a subset
```

I think?

CLOUDPROVIDERS.md
```diff
@@ -0,0 +1,98 @@
...
+All configuration is in a stanza named after the backend, and takes simple key value pairs.
+
+**Cloud providers should be disabled on the aggregation server when using http forwarding, as the source IP isn't
+propagated, and that information should be collected on the ingestion server.**
+
+aws
+---
+### TODO
+
+k8s
```
tiedotguy: We should include a line about the host behavior and pod name, and the implications of it.  It's called out in the examples, but there's no context there for it.

examples/cloudproviders/k8s/daemonset.yaml
```diff
@@ -0,0 +1,116 @@
...
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        app: gostatsd
+      annotations:
+        myapp.company.com/imageVersion: "latest"
+    spec:
+      serviceAccountName: gostatsd
+      # We need the host network so we have an explicitly addressable pod from every pod running on the node.
```
tiedotguy: Something I hadn't considered - one of the big problems with statsd is that almost all clients do "resolve one, refresh never", which is a massive pain.  What are the implications of updating the DaemonSet?  Will the IP remain constant because of `hostNetwork` being true?

pkg/cloudproviders/k8s/k8s.go
```diff
@@ -0,0 +1,375 @@
...
+
+// Instance returns pod details from k8s API server watches.
+// ip -> nil pointer if pod was not found.
+// map is returned even in case of errors because it may contain partial data.
+// An "instance", as far as k8s cloud provider is concerned, is an individual pod running in the cluster
+func (p *Provider) Instance(ctx context.Context, IP ...gostatsd.IP) (map[gostatsd.IP]*gostatsd.Instance, error) {
+	instanceIPs := make(map[gostatsd.IP]*gostatsd.Instance, len(IP))
+	var returnErr error
+
+	p.logger.WithField("cacheKeys", p.podsInf.GetIndexer().ListKeys()).Debug("pods in cache")
```
tiedotguy: `ListKeys()` is always evaluated, is this an expensive call?

tiedotguy:
Looks good! I assume we're not cutting a new release until @ash2k has his refactoring work in?

examples/cloudproviders/k8s/daemonset.yaml
mwhittington21: Yes, the IP remains constant. The old pod is terminated before the new one can be started because both pods cannot share the same IP:port combination.

pkg/cloudproviders/k8s/k8s.go
mwhittington21: Yeah it probably isn't the cheapest. I will fix this.

pkg/cloudproviders/k8s/k8s.go
mwhittington21: I've just set it back to async calls as per the contract and removed cache checking. We can revisit this later.

pkg/cloudproviders/k8s/k8s.go
mwhittington21: It was useful for debugging but I've removed it completely since I couldn't think of a nicer way.

mwhittington21:
Last round of nits have been addressed!

ash2k:
@tiedotguy I think we don't have to wait for my refactoring, this is good and working code 🤞

tiedotguy:
@ash2k if you give it an approve, I'll do a quick update of the changelog with pending changes, and cut a release.

ash2k:
@mwhittington21 feel free to merge it! I suggest to squash it - we don't need all the 23 commits :)

PR #286, opened 2020-02-19
add regex matching support

ermeaney:
We had a use case were we need more flexible filtering, and I added regex support for matching.

I am sure there is some performance hit, but we were willing to accept that

### DIFF
--- a/FILTERING.md
+++ b/FILTERING.md
@@ -35,6 +35,8 @@ Examples:
 - !abc* - matches "xyz" but not "abc" or "abcd"
 - regex:.*abc.* - matches "xyz.abc.123" but not "xyz.123"
 - !regex:.*abc.* - matches "xyz.123" but not "xyz.abc.123"
+- !regex:^abc.* - matches "abc.123" and "abcd.123" but not "xyz.abc.123"
+- !regex:.*\.count$ - matches "abc.123.count" and "abcd.123.count" but not "xyz.abc.count.123"
 
 ## Filter examples
 
--- a/FILTERING.md
+++ b/FILTERING.md
@@ -24,12 +24,16 @@ A filter block contains up to 6 keys.  3 for filtering rules, and 3 for actions 
 ## Matching
 A match is defined as a case sensitive string with an optional ! prefix to invert the meaning, and an optional * suffix
 to indicate it is a prefix match.  Note: it is not a wildcard, it is a prefix match only.
+## Regex matching
+If a match is prefixed with `regex:` (after the `!` if you want it inverted) then the rest of the pattern is a golang regex. The trailing `*` behavior is diffrent as it is part of the regex and not a prefix match.
 
 Examples:
 - abc - matches the "abc", but not ABC or abcd
 - abc* - matches "abc" and "abcd"
 - !abc - matches "xyz" and "abcd" but not "abc"
 - !abc* - matches "xyz" but not "abc" or "abcd"
+- regex:.*abc.* - matches "xyz.abc.123" but not "xyz.123"
+- !regex:.*abc.* - matches "xyz.123" but not "xyz.abc.123"
 
 ## Filter examples
 
--- a/FILTERING.md
+++ b/FILTERING.md
@@ -24,8 +24,9 @@ A filter block contains up to 6 keys.  3 for filtering rules, and 3 for actions 
 ## Matching
 A match is defined as a case sensitive string with an optional ! prefix to invert the meaning, and an optional * suffix
 to indicate it is a prefix match.  Note: it is not a wildcard, it is a prefix match only.
+
 ## Regex matching
-If a match is prefixed with `regex:` (after the `!` if you want it inverted) then the rest of the pattern is a golang regex. The trailing `*` behavior is diffrent as it is part of the regex and not a prefix match.
+If a match is prefixed with `regex:` (after the `!` if you want it inverted) then the rest of the pattern is a golang regex. The trailing `*` behavior is diffrent as it is part of the regex and not a prefix match. See [re2](https://github.com/google/re2/wiki/Syntax) for syntax.
 
 Examples:
 - abc - matches the "abc", but not ABC or abcd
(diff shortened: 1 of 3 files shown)


FILTERING.md
```diff
@@ -24,12 +24,16 @@ A filter block contains up to 6 keys.  3 for filtering rules, and 3 for actions
...
 to indicate it is a prefix match.  Note: it is not a wildcard, it is a prefix match only.
+## Regex matching
+If a match is prefixed with `regex:` (after the `!` if you want it inverted) then the rest of the pattern is a golang regex. The trailing `*` behavior is diffrent as it is part of the regex and not a prefix match.
 
 Examples:
 - abc - matches the "abc", but not ABC or abcd
 - abc* - matches "abc" and "abcd"
 - !abc - matches "xyz" and "abcd" but not "abc"
 - !abc* - matches "xyz" but not "abc" or "abcd"
+- regex:.*abc.* - matches "xyz.abc.123" but not "xyz.123"
```
tiedotguy: `MatchString` will look anywhere in the string, which makes the `.*` redundant, and may produce unexpected results if removed (because it still matches).  The examples should include `abc` and `^abc$` as a bare minimum to show anchoring behavior.

matcher.go
```diff
@@ -1,43 +1,61 @@
...
-	prefix := false
-	if strings.HasSuffix(s, "*") {
-		prefix = true
-		s = s[0 : len(s)-1]
+	regex := false
+	var compiledRegex *regexp.Regexp = nil
+	if strings.HasPrefix(s, "regex:") {
+		regex = true
+		s = s[6:]
+		compiledRegex, _ = regexp.Compile(s)
```
tiedotguy: Don't ignore errors - it would be better to propagate the error out, but that's a bit of a pain to plumb through the system. You can use `regexp.MustCompile`.  Since this is initialization only code, it's ok (but not ideal) to fail fast with a panic.

matcher.go
```diff
@@ -1,43 +1,61 @@
...
 	}
-
+	prefix := false
+        if !regex {
```
tiedotguy: Sadly, go uses tabs not spaces (also on line 39)

tiedotguy:
Please notify @ash2k when the CLA has been signed, and he can confirm it.

matcher.go
```diff
@@ -1,43 +1,61 @@
...
 	}
 }
 
 // Match indicates if the provided string matches the criteria for this StringMatch
 func (sm StringMatch) Match(s string) bool {
-	if sm.prefixMatch {
-		return strings.HasPrefix(s, sm.test) != sm.invertMatch
+	if sm.isRegex {
+		return (sm.regex.MatchString(s)) != sm.invertMatch
+	} else {
```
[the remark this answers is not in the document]
ash2k: `else` may be omitted (Go style) because of the return in the `if` block.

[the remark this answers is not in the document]
tiedotguy: In this particular case, I'd personally keep the `else` in, because I think the code flows better. Although since you've got me thinking about it now, a `switch` is probably even clearer because it's more explicit and regular in the flow:
```
switch {
case sm.isRegex:
  return sm.regex.MatchString(s) != sm.invertMatch
case sm.prefixMatch:
  return strings.HasPrefix(s, sm.test) != sm.invertMatch
default: // exact match
  return (s == sm.test) != sm.invertMatch
}
```

FILTERING.md
```diff
@@ -24,12 +24,16 @@ A filter block contains up to 6 keys.  3 for filtering rules, and 3 for actions
 ## Matching
 A match is defined as a case sensitive string with an optional ! prefix to invert the meaning, and an optional * suffix
 to indicate it is a prefix match.  Note: it is not a wildcard, it is a prefix match only.
+## Regex matching
```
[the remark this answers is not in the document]
ermeaney: Fixed

FILTERING.md
```diff
@@ -24,12 +24,16 @@ A filter block contains up to 6 keys.  3 for filtering rules, and 3 for actions
...
+If a match is prefixed with `regex:` (after the `!` if you want it inverted) then the rest of the pattern is a golang regex. The trailing `*` behavior is diffrent as it is part of the regex and not a prefix match.
```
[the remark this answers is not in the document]
ermeaney: done

ermeaney:
Signed the CLA, and fixed all the things I was asked for.

BTW, the switch vs the if/else does look a lot cleaner, thanks for that

ermeaney:
@ash2k , sorry missed the spot where I was supposed to let you know that I signed the CLA via docusign

ash2k:
I'm confirming that the CLA has been signed. Thanks!

tiedotguy:
Thanks @ash2k !

matcher.go
```diff
@@ -1,43 +1,61 @@
...
+	"regexp"
 	"strings"
 )
 
 type StringMatch struct {
 	test        string
 	invertMatch bool
 	prefixMatch bool
+	isRegex     bool
+	regex       *regexp.Regexp
```
ash2k: This field can be used as a flag instead of the bool above. It's not a regex if the pointer is `nil`.

matcher.go
```diff
@@ -1,43 +1,61 @@
 package gostatsd
 
 import (
+	"regexp"
 	"strings"
 )
 
 type StringMatch struct {
```
ash2k: I wonder if it'd be better to turn this into an interface with multiple simple implementations. Perhaps it's not worth the effort at this point, just a thought.

tiedotguy: I'd been thinking that. Definitely something to benchmark and not do blindly, as this is one of the hottest code paths in the service.

tiedotguy:
It looks ok to me. The docs are slightly wrong (the last two examples don't consider the inversion). Just to get this through, if you give it an approve @ash2k, I'll merge it, then update the changelog, docs, and using `regex` as a flag in the release PR.

tiedotguy:
@ermeaney this is being released now as 20.1.0, thanks for the contribution!

PR #290, opened 2020-02-25
Cloud provider refactor

ash2k:
The idea is to split `CloudHandler` into two pieces by introducing an interface in the middle.
Currently `CloudHandler` does these things:
1. maintains a (negative) cache of ip->instance info
1. maintains a map of ip->[]metrics that need to be enriched per ip while the lookup it taking place

I'm introducing `CachedInstances` interface to abstract the caching bit (1 in the list above) and leave only 2 in the `CloudHandler`.

The next step/PR is to refactor the k8s cloud provider into an implementation of `CachedInstances`.

This PR is not ready to review in full (tests don't even compile as I haven't touched them yet) but to give you a preview and collect early feedback on the approach and major impl details. WDYT?

### DIFF
--- a/pkg/cachedinstances/cloudprovider/cached_cloud_provider.go
+++ b/pkg/cachedinstances/cloudprovider/cached_cloud_provider.go
@@ -1,0 +1,261 @@
+package cloudprovider
+
+import (
+	"context"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/ash2k/stager/wait"
+	"github.com/atlassian/gostatsd"
+	"github.com/atlassian/gostatsd/pkg/stats"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/time/rate"
+)
+
+func NewCachedCloudProvider(logger logrus.FieldLogger, limiter *rate.Limiter, cloudProvider gostatsd.CloudProvider, cacheOpts gostatsd.CacheOptions) *CachedCloudProvider {
+	return &CachedCloudProvider{
+		logger:         logger,
+		limiter:        limiter,
+		cloudProvider:  cloudProvider,
+		cacheOpts:      cacheOpts,
+		ipSinkSource:   make(chan gostatsd.IP),
+		infoSinkSource: make(chan gostatsd.InstanceInfo),
+		emitChan:       make(chan stats.Statser),
+		cache:          make(map[gostatsd.IP]*instanceHolder),
+	}
+}
+
+type CachedCloudProvider struct {
+	// These fields may only be read or written by the main CachedCloudProvider.Run goroutine
+	statsCacheRefreshPositive uint64 // Cumulative number of positive refreshes (ie, a refresh which succeeded)
+	statsCacheRefreshNegative uint64 // Cumulative number of negative refreshes (ie, a refresh which failed and used old data)
+	statsCachePositive        uint64 // Absolute number of positive entries in cache
+	statsCacheNegative        uint64 // Absolute number of negative entries in cache
+
+	logger         logrus.FieldLogger
+	limiter        *rate.Limiter
+	cloudProvider  gostatsd.CloudProvider
+	cacheOpts      gostatsd.CacheOptions
+	ipSinkSource   chan gostatsd.IP
+	infoSinkSource chan gostatsd.InstanceInfo
+
+	// emitChan triggers a write of all the current stats when it is given a Statser
+	emitChan     chan stats.Statser
+	rw           sync.RWMutex // Protects cache
+	cache        map[gostatsd.IP]*instanceHolder
+	toLookupIPs  []gostatsd.IP
+	toReturnInfo []gostatsd.InstanceInfo
+}
+
+func (ccp *CachedCloudProvider) Run(ctx context.Context) {
+	var (
+		toLookupC     chan<- gostatsd.IP
+		toLookupIP    gostatsd.IP
+		toReturnInfoC chan<- gostatsd.InstanceInfo
+		toReturnInfo  gostatsd.InstanceInfo
+		wg            wait.Group
+	)
+	// this goroutine needs to populate/update the cache so an intermediate InstanceInfo channel is used below that allows
+	// to intercept, update the cache and then push the information through to the cache consumer.
+	ownInfoSource := make(chan gostatsd.InstanceInfo)
+	ld := cloudProviderLookupDispatcher{
+		logger:        ccp.logger,
+		limiter:       ccp.limiter,
+		cloudProvider: ccp.cloudProvider,
+		ipSource:      ccp.ipSinkSource, // our sink is their source
+		infoSink:      ownInfoSource,    // their sink is our source
+	}
+
+	defer wg.Wait() // Wait for cloudProviderLookupDispatcher to stop
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel() // Tell cloudProviderLookupDispatcher to stop
+
+	wg.StartWithContext(ctx, ld.run)
+
+	refreshTicker := time.NewTicker(ccp.cacheOpts.CacheRefreshPeriod)
+	defer refreshTicker.Stop()
+	// No locking for ccp.cache READ access required - this goroutine owns the object and only it mutates it.
+	// So reads from the same goroutine are always safe (no concurrent mutations).
+	// When we mutate the cache, we hold the exclusive (write) lock to avoid concurrent reads.
+	// When we read from the cache from other goroutines in the Peek() method, we obtain the read lock.
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case toLookupC <- toLookupIP:
+			toLookupIP = gostatsd.UnknownIP // enable GC
+			toLookupC = nil                 // ip has been sent; if there is nothing to send, the case is disabled
+		case toReturnInfoC <- toReturnInfo:
+			toReturnInfo = gostatsd.InstanceInfo{} // enable GC
+			toReturnInfoC = nil                    // info has been sent; if there is nothing to send, the case is disabled
+		case info := <-ownInfoSource:
+			ccp.handleInstanceInfo(info)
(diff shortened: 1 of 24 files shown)


mwhittington21:
Seems sane to me! Splitting out the caching behaviour from a default cloud provider makes a lot of sense.

ash2k:
I'll get back to this PR, hopefully soon.

pkg/cachedinstances/factory.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/atlassian/gostatsd"
+	"github.com/atlassian/gostatsd/pkg/cachedinstances/cloudprovider"
+	"github.com/atlassian/gostatsd/pkg/cloudproviders/aws"
+	"github.com/atlassian/gostatsd/pkg/cloudproviders/k8s"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/viper"
+	"golang.org/x/time/rate"
+)
+
+var (
```
ash2k, line 16: Two types of defaults will be gone once k8s provider is no longer a provider but a direct implementation of `CachedInstances` interface (next PR).

pkg/statsd/statsd.go
```diff
@@ -88,12 +86,6 @@ func socketFactory(metricsAddr string, connPerReader bool) SocketFactory {
 func (s *Server) createStandaloneSink() (gostatsd.PipelineHandler, []gostatsd.Runnable, error) {
 	var runnables []gostatsd.Runnable
 
-	for _, backend := range s.Backends {
```
ash2k, line 91: Server receives backends as a parameter - it is not the owner, it didn't create them. So it should not be managing their lifecycle. That is why I moved this up one level into whatever constructs the `Server`. Same applies to `CachedInstances` and cloud providers.

types.go
```diff
@@ -48,11 +48,25 @@ type TimerSubtypes struct {
...
 type Runner interface {
-	Run(ctx context.Context)
+	Run(context.Context)
+}
+
+type MetricsRunner interface {
+	RunMetricsContext(context.Context)
+}
+
+func MaybeAppendRunnable(runnables []Runnable, maybeRunner interface{}) []Runnable {
```
ash2k, line 62: Single helper func to ensure we always start both the main work loop and the metrics loop too.

ash2k:
This is ready for review.

tiedotguy:
Awesome! I'll try and give it a thorough review tomorrow.

pkg/cachedinstances/cloudprovider/cached_cloud_provider.go
```diff
@@ -0,0 +1,261 @@
...
+			toReturnInfo = gostatsd.InstanceInfo{} // enable GC
+			toReturnInfoC = nil                    // info has been sent; if there is nothing to send, the case is disabled
+		case info := <-ownInfoSource:
+			ch.handleInstanceInfo(info)
+		case t := <-refreshTicker.C:
+			ch.doRefresh(t)
+		case statser := <-ch.emitChan:
+			ch.emit(statser)
+		}
+		if toLookupC == nil && len(ch.toLookupIPs) > 0 {
```
tiedotguy, line 100: `ch.toLookupIPs` is only ever populated by `ch.doRefresh()`.  `ch.doRefresh()` could just return a slice of IPs to refresh, then this could pass that entire thing as a single slice to `ch.ipSinkSource`, rather than feeding it through one at a time.

pkg/cachedinstances/cloudprovider/cached_cloud_provider.go
```diff
@@ -0,0 +1,261 @@
...
+		case toLookupC <- toLookupIP:
+			toLookupIP = gostatsd.UnknownIP // enable GC
+			toLookupC = nil                 // ip has been sent; if there is nothing to send, the case is disabled
+		case toReturnInfoC <- toReturnInfo:
+			toReturnInfo = gostatsd.InstanceInfo{} // enable GC
+			toReturnInfoC = nil                    // info has been sent; if there is nothing to send, the case is disabled
+		case info := <-ownInfoSource:
+			ch.handleInstanceInfo(info)
+		case t := <-refreshTicker.C:
+			ch.doRefresh(t)
```
tiedotguy, line 96: I would suggest skipping refreshing if we haven't managed to clear the previous refresh yet. I could see a situation where we get in to a bad state (due to cpu or network, not necessarily a bug), where we're refreshing faster than we can process, causing either unbound growth in `ch.toLookupIPs` or having something to write before we're ready if the below suggestion is done.

I'm not so sure if the same pattern is useful in `toReturnInfo`, it complicates the interface a bunch.

tl;dr backpressure on `ch.toLookupIPs`

ash2k: I agree and I thought about that. But I think it should be done separately. This PR does not make it better or worse, right? Also, don't want to make it even bigger.

pkg/cachedinstances/cloudprovider/cached_cloud_provider.go
ash2k: What are the benefits? This will complicate logic quite a bit. `cloudProviderLookupDispatcher.run()` would need to be handle all the extra complexity of collecting just enough IPs in a buffer and then processing them. I tried such approach somewhere during this refactoring and found that it complicates things, hurts readability. And when things are not simple, bugs creep in. :)

pkg/cachedinstances/cloudprovider/cached_cloud_provider.go
```diff
@@ -0,0 +1,261 @@
...
+	case <-ctx.Done():
+		// success-ish
+	default:
+		// at least we tried
+	}
+}
+
+func (ch *CachedCloudProvider) emit(statser stats.Statser) {
+	// regular
+	statser.Gauge("cloudprovider.cache_positive", float64(ch.statsCachePositive), nil)
```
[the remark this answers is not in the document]
ash2k: Oh, yeah, I'd like to discuss the metrics we emit as I want' paying too much attention to it while refactoring. Was too much to handle already so I might have disrupted things a bit.
I think not all active components emit stats properly. Some didn't have their `RunMetricsWithContext()`/etc called, some probably never get a non-null statser. An audit should take place. I tried to patch up things a bit, but haven't verified every single component emits stats properly.

pkg/cachedinstances/cloudprovider/cached_cloud_provider.go
tiedotguy: Splitting this in to two aspects, which should be considered individually - returning a slice, and passing a slice to the lookup.

Returning a slice removes the extra state from the structure, simplifying the query of "where is this referenced?". Reducing the scope to a local variable means I don't need to have to think about locking, or what else might touch it, because I can see it's a regular local variable, returned from a function.  The only thing that it simplifies currently is no back pressure. If back pressure were added so we didn't call `doRefresh`, then it wouldn't even add that, we'd just be returning a slice through the struct.

Passing a slice to lookup means the lookup dispatcher has to do a little extra thinking, but I wouldn't say it's much. Without writing code, I'd probably wrap it up in a type which has 3 methods... "push more", "pop maxLookupIps or nil", and "get maxLookupIPs or what you have". No channels, no timers, just a simple data structure (probably just a slice) which is trivial to throw in to a test. The actual consumption of the data structure is also easier to follow at that point.

ash2k:
@tiedotguy thanks for the review! I know I've done a few too many changes here that could all be separate PRs/commits. I know it's a poor excuse, but I don't have much time to rework this just for the sake of nicer history (which I do like to have too!). If nobody feels strongly about it, can we merge this (assuming everything else is fine)? If there are objections, I'll do more work here.

Regarding putting things into the root package - I don't mind having them there but also don't mind moving them somewhere else if that's what you and others prefer. I like to have api types there because a hypothetical `gostatsd.Metric`/etc reads nice. In the particular case of this PR I moved things because there was some dependency cycle. I do not remember now what it was, sorry, this PR is "spread in time" too much and I'm forgetting "whats" and "whys".

If we go ahead with this, I can squash it before merging.

Next step, which I'm planning to work on, is to refactor the k8s cloud provider into the new interface and do some cleanup around it (e.g. no two sets of defaults will be needed).

pkg/cachedinstances/cloudprovider/cached_cloud_provider.go
tiedotguy: Works for me 👍

pkg/cachedinstances/cloudprovider/cached_cloud_provider.go
ash2k: Thanks for elaborating on the idea. I think actually trying it will help to see if it's worth it or not. I think if more logic needs to be packed into how IPs are handled then it may make sense. Another thing I'm thinking of is turning it into a map to dedupe IPs that have been in the queue for too long. And then we can just have a capacity limit on that map. Or maybe we don't even need that capacity limit with deduplication...

pkg/cachedinstances/cloudprovider/cached_cloud_provider.go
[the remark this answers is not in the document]
tiedotguy: I'm sure @adampetrovic or @MovieStoreGuy will let us know if it broke anything :)

In terms of metric changes in this PR, I don't believe there's any loss of fidelity - there's the CP state (cache), and CP actions (actual CP work), and they're just split out, so it's all good.  Maybe for k8s, write the PR without metrics, and then raise an issue to refactor metrics later?

Once we have k8s, we can assess what metrics it can sensible generate, how they overlap with what we already have, and if we need new names / tags / whatever.

tiedotguy:
I did track down the dependency cycle, and I agree it was needed. I don't recall specifically what it was now though :)

My main objection to the root is really the physical file placement. The mixing of source files and project files is a mixing of concerns.  Either way, it doesn't need to be handled in this PR

I think this is clear to merge now.

ash2k:
Cool! I have rebased and squashed, tick it one more time, please.

PR #291, opened 2020-02-25
Release 20.1.0

tiedotguy:
Also includes a minor refactor to regexp handling in the filtering.

### DIFF
--- a/matcher.go
+++ b/matcher.go
@@ -9,39 +9,30 @@ type StringMatch struct {
 	test        string
 	invertMatch bool
 	prefixMatch bool
-	isRegex     bool
 	regex       *regexp.Regexp
 }
 
 type StringMatchList []StringMatch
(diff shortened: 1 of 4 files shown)


matcher.go
```diff
@@ -9,47 +9,38 @@ type StringMatch struct {
...
-	}
-	prefix := false
-	if !regex {
-		if strings.HasSuffix(s, "*") {
-			prefix = true
-			s = s[0 : len(s)-1]
-		}
+	} else if strings.HasSuffix(s, "*") {
+		prefix = true
+		s = s[0 : len(s)-1]
```
ash2k: FYI `0` can be omitted. Also, is this formatted by gofmt?

tiedotguy: It is, which honestly surprised me.

PR #297, opened 2020-03-18
fix `metricMapsReceived` of aggregator

beanliu:
**Problem**

The `aggregator.metricmaps_received` values of different workers are always the same due to a typo in the codes. 

**Fix**

- Correct the variable typo.
- Reset `aggregator.metricmaps_received`  in the same way with `metricsReceived `.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+20.1.2
+------
+- Fixes a bug where `aggregator.metricmaps_received` are the same for for every worker
+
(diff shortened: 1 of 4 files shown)


tiedotguy:
This can be 20.1.2, doesn't need a larger bump than that. Can you add a quick update on `CHANGELOG.md` to note the fix?

beanliu:
Thanks for the quick review. I've updated the metrics doc and change log.

PR #298, opened 2020-03-20
Make `transport.ResponseHeaderTimeout` configurable

beanliu:
When the backend connects to an AWS ELB and one or more instances behind the LB go offline, the total flush delay would hit ~20s, some of the batches might got dropped, and may last for a few minutes (depends on how you configure the ELB) before the LB kicks off the dead node.

Since the TCP connect timeout (10s) is not configurable on AWS side and it is a 7 layer proxy which means the client side has no knowledge of the LB backend side. We need a way in gostatsd to be more sensitive to the backend failure so that the backoff-retry could take effect to avoid data loss.

The `http.transport.ResponseHeaderTimeout` seems a good fit and we've tested it works under normal loads.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+20.2.0
+------
+- Support for `response-header-timeout` for `transport` section
+
 20.1.2
 ------
 - Fixes a bug where `aggregator.metricmaps_received` are the same for for every worker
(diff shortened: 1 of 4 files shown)


tiedotguy:
Looks good generally, needs an update to CHANGELOG.md and TRANSPORT.md, and a minor version bump (20.2.0).

> The http.transport.ResponseHeaderTimeout seems a good fit and we've tested it works under normal loads.

Has this had an opportunity to be tested in the LB node cycling case?

Also for actual releasing (ie, once master has the changelog/documentation updated), it should be tagged to make the actual release...
```
git tag 20.2.0
git push origin 20.2.0
```
... this will kick off the whole dockerhub thing.

beanliu:
Thanks for the quick review and release tip. 👍 

> Has this had an opportunity to be tested in the LB node cycling case?

Yep, a proper timeout value could reduce the chance of data loss greatly (by several orders of magnitude).

tiedotguy:
Good stuff!

PR #302, opened 2020-04-14
fix: use correct timestamps for newrelic backend

Kav91:
The incorrect timestamp is currently being used in the payload, so when this is sent to New Relic, viewing each metric / event does not show the timestamp on flush, but from the individual metric itself. 

For example when a gauge may have not been sent for awhile, it will reuse the last gauges timestamp as the default which will throw off graphing when it should be using the the timestamp during flush.

PR uses the correct timestamp, and supplies the individual metrics attribute as a custom attribute instead.

### DIFF
--- a/pkg/backends/newrelic/flush.go
+++ b/pkg/backends/newrelic/flush.go
@@ -147,7 +147,7 @@ func newMetricSet(n *Client, f *flush, metricName, Type string, Value float64, t
 
 	// GoStatsD provides the timestamp in Nanotime, New Relic requires seconds or milliseconds, see under "Limits and restricted characters"
 	// https://docs.newrelic.com/docs/insights/insights-data-sources/custom-data/send-custom-events-event-api#instrument
-	metricSet["timestamp"] = n.now().Unix()
+	metricSet["timestamp"] = f.timestamp
 	metricSet["interval"] = f.flushIntervalSec
 	metricSet["integration_version"] = integrationVersion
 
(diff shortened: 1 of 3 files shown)


pkg/backends/newrelic/flush.go
```diff
@@ -142,12 +142,12 @@ func (f *flush) finish() {
...
 }
 
-func newMetricSet(n *Client, f *flush, metricName, Type string, Value float64, tags gostatsd.Tags, timestamp gostatsd.Nanotime) map[string]interface{} {
+func newMetricSet(n *Client, f *flush, metricName, Type string, Value float64, tags gostatsd.Tags) map[string]interface{} {
 	metricSet := map[string]interface{}{}
 
 	// GoStatsD provides the timestamp in Nanotime, New Relic requires seconds or milliseconds, see under "Limits and restricted characters"
 	// https://docs.newrelic.com/docs/insights/insights-data-sources/custom-data/send-custom-events-event-api#instrument
-	metricSet["timestamp"] = timestamp / 1e9
+	metricSet["timestamp"] = n.now().Unix()
```
tiedotguy, line 150: This will potentially cause a skew, as the metric data is built up. Because of second precision it's unlikely to be a problem, but I would still suggest avoiding it.  I believe the correct fix is to use `f.timeStamp`, from https://github.com/atlassian/gostatsd/blob/master/pkg/backends/newrelic/newrelic.go#L201

It's probably worth changing the type in the `flush` struct as well.

```suggestion
	metricSet["timestamp"] = int64(f.timestamp / 1e9)
```

(also in `newDimensionalMetricSet`)

Kav91:
@tiedotguy thanks for the suggestion, makes sense.

The only thing I changed was not dividing, since the flush timestamp appears to be in seconds already from `n.now().Unix()`.

tiedotguy:
In my defense, I had just woken up :)

PR #304, opened 2020-04-26
Refactor k8s backend to implement CachedInstances directly

ash2k:
First commit just does the direct conversion and also some minor cleanups.
Second commit adds an internal cache to avoid regexes in the fast path.

Yes, it's funny that I've removed the cache to add it back BUT:
- this cache is invalidated immediately if `Pod` has been updated
- this cache is populated on the first lookup once the `Pod` is in the informer's cache

The above means that cache is never stale or stale negative.

I need to add unit tests and actually test it against real kube.


pkg/cachedinstances/cloudprovider/cached_cloud_provider_test.go
```diff
@@ -53,7 +53,7 @@ func TestCachedCloudProviderExpirationAndRefresh(t *testing.T) {
 	require.Nil(t, instance)
 	require.False(t, exists)
 	ci.IpSink() <- peekIp
-	time.Sleep(70 * time.Millisecond) // Should be refreshed couple of times and evicted.
+	time.Sleep(500 * time.Millisecond) // Should be refreshed couple of times and evicted.
```
ash2k, line 56: This test is flaky. This seem to help.

pkg/cachedinstances/factory.go
```diff
@@ -1,85 +0,0 @@
-package cachedinstances
```
ash2k, line 1: This file is refactored into the one above ^ (`pkg/cachedinstances/cachedinstances.go`)

pkg/cachedinstances/k8s/k8s.go
```diff
@@ -254,7 +285,7 @@ func NewProvider(logger logrus.FieldLogger, clientset kubernetes.Interface, podI
 		if podInfOpts.NodeName == "" {
 			return nil, fmt.Errorf("watch-cluster set to false, and node name not supplied")
 		}
-		fieldSelector := "spec.nodeName=" + podInfOpts.NodeName
+		fieldSelector := fields.OneTermEqualSelector("spec.nodeName", podInfOpts.NodeName).String()
```
ash2k, line 288: Correctly construct it with escaping.

pkg/cachedinstances/k8s/k8s_test.go
```diff
@@ -241,64 +229,24 @@ func TestIPToTags(t *testing.T) {
...
-		require.NoError(t, err)
-
-		instance := instanceData[ipAddr]
+		instance, cacheHit := fixtures.provider.Peek(ipAddr)
+		require.True(t, cacheHit)
 		require.Nil(t, instance)
 	}, viper.New(), nodeName)
 }
 
-func TestWatchNodeOnly(t *testing.T) {
```
ash2k, line 262: Watch/list filtering happens on server so it's not really possible to test it using fakes. See below.
Deleted the test.

pkg/cachedinstances/k8s/k8s_test.go
```diff
@@ -241,64 +229,24 @@ func TestIPToTags(t *testing.T) {
...
-		p.ObjectMeta.Annotations = map[string]string{nodeAnnotationKey: "node2"}
-		p.ObjectMeta.Labels = map[string]string{}
-		fixtures.podsWatch.Add(p)
-
-		waitForFullCache(t, fixtures, 2)
-
-		ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
-		defer cancel()
-
-		instanceData, err := fixtures.cloudProvider.Instance(ctx, ipAddr)
```
ash2k, line 288: So we ask about `ipAddr` only but then check that the returned map does not have `ipAddr2`. Of course it does not. I.e. bug in test. It always fails because filtering is done server-side.

pkg/cachedinstances/cloudprovider/cached_cloud_provider_test.go
tiedotguy: #274 :(

Really need to get a virtual clock plumbed in to fix it properly.

tiedotguy:
Looks ok at a quick glance, I'll try and give it a thorough review tomorrow though!

pkg/cachedinstances/cloudprovider/cached_cloud_provider_test.go
mwhittington21: I also had trouble with that test last time I edited but I wasn't sure if increasing the time bound like that would change some property of the test. If we're sure, I have no problems with this change.

tiedotguy: It used to be that increasing it would change the number of items from 1 to 2 and that would fail the test - that's ultimately what #274 fixed.  If it's too low, it can go from 1 to 0 though, which is what #274 caused...

ash2k: I'm happy to revert this because it's not relevant to this PR. Especially if it's been increased and decreased before.

tiedotguy: Nah, it's fine, better to not be flaky.  If tests run slower, that's just motivation to add the virtual clock.

ash2k:
I think this is ready to be merged but I haven't tested it in a real cluster. I've added two tests for cache invalidation and it seem to work ok. Would be great if someone can test it in a cluster to see if something is not working (including changes from the previous refactoring). :)

tiedotguy:
@ash2k - give it a tag (20.4.0-rc1?) so it gets a docker build, and we can poke @mwhittington21 until he tests it :)

ash2k:
I've pushed one more commit to add negative caching and to remove logging from fast path. `20.4.0-rc4`

tiedotguy:
#328 is basically going to nuke `gostatsd.SourceIP`, so I want to get this in first otherwise there's going to be massive merge conflicts.

ash2k:
Yes, merging this one would be good as I will not have time to fix complex conflicts.

tiedotguy:
I've been chipping away at making sure I understand it, with my current momentum I expect to be done tomorrow.

tiedotguy:
I'm a little nervous at the casts without checks, but I'm assuming you're aware of the invariants in the upstream library and don't need to check them.

Bring on generics!

ash2k:
@tiedotguy Thanks for reviewing. Could you clarify which casts you are talking about? I've just checked the code and I don't understand what you are referring to :)

tiedotguy:
`pod := objs[0].(*core_v1.Pod)` and similar in `cacheInvalidationHandler.OnUpdate`

ash2k:
Yes, this is fine.

tiedotguy:
@mwhittington21 are you able to give this a spin? I've tagged it as 24.0.1-rc1, but it'll be included in a real release when #327 gets out.

mwhittington21:
I'll see if I can give it a spin in a test environment - nowhere is using it with high load or anything yet. Will get back to you.

mwhittington21:
Sorry for only just now getting back to you @tiedotguy, I've finally had a chance to test this. Seems to work fine although we don't have a lot of load going through the system yet as we're only dogfooding a handful of services.

tiedotguy:
Good stuff, thanks!

PR #309, opened 2020-06-10
Adds support for per metric-type expiry intervals.

### LINKED ISSUE #300: Per metric-type expiry interval
How I can send to back the previous value, if the gauge is not updated at the next flush?

tiedotguy:
Closes #300.

### DIFF
--- a/pkg/statsd/aggregator.go
+++ b/pkg/statsd/aggregator.go
@@ -23,21 +23,36 @@ type percentStruct struct {
 
 // MetricAggregator aggregates metrics.
 type MetricAggregator struct {
-	metricsReceived    uint64
-	metricMapsReceived uint64
-	expiryInterval     time.Duration // How often to expire metrics
-	percentThresholds  map[float64]percentStruct
-	now                func() time.Time // Returns current time. Useful for testing.
-	statser            stats.Statser
-	disabledSubtypes   gostatsd.TimerSubtypes
-	histogramLimit     uint32
-	metricMap          *gostatsd.MetricMap
+	metricsReceived       uint64
+	metricMapsReceived    uint64
+	expiryIntervalCounter time.Duration // How often to expire counters
+	expiryIntervalGauge   time.Duration // How often to expire gauges
+	expiryIntervalSet     time.Duration // How often to expire sets
+	expiryIntervalTimer   time.Duration // How often to expire timers
+	percentThresholds     map[float64]percentStruct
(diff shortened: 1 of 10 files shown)


pkg/statsd/aggregator.go
```diff
@@ -187,8 +202,8 @@ func (a *MetricAggregator) Process(f ProcessFunc) {
 	f(a.metricMap)
 }
 
-func (a *MetricAggregator) isExpired(now, ts gostatsd.Nanotime) bool {
-	return a.expiryInterval != 0 && time.Duration(now-ts) > a.expiryInterval
+func (a *MetricAggregator) isExpired(interval time.Duration, now, ts gostatsd.Nanotime) bool {
```
ash2k, line 205: Can be a plain function rather than a method.

pkg/statsd/statsd.go
```diff
@@ -8,14 +8,15 @@ import (
...
-	"github.com/atlassian/gostatsd/pkg/web"
 	"github.com/libp2p/go-reuseport"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/viper"
 	"golang.org/x/time/rate"
+
+	"github.com/atlassian/gostatsd"
+	"github.com/atlassian/gostatsd/pkg/stats"
+	"github.com/atlassian/gostatsd/pkg/transport"
+	"github.com/atlassian/gostatsd/pkg/web"
```
ash2k, line 19: What is happening with formatting? :)

pkg/statsd/aggregator.go
tiedotguy: Good catch, fixed

pkg/statsd/statsd.go
tiedotguy: I think goland did a format, and it should be re-ordered anyway.

MovieStoreGuy:
Do we know what the impact to memory would be adding this in?

tiedotguy:
> Do we know what the impact to memory would be adding this in?

For space, effectively nothing (24 bytes * number of aggregators, which is the number of logical procs by default). `MetricAggregator`s are long lived structures created on startup, not a per-metric or per-aggregated-metric structure.

For bus / cache pressure, also effectively nothing. It's not the most efficient because it's 8->32 bytes per aggregator (it could be a single shared data structure for instance), however it's also only used on flush, which isn't the hot path.

PR #310, opened 2020-06-22
Add support for adding http headers extracted from tags

### LINKED ISSUE #305: Support forwarder adding custom headers dynamically
#261 added support of configuring custom headers to the forwarder. In a multi-tenant environment such as gostatsd deployed as daemonset and shared by multiple pods, we'd like to have support of injecting custom headers from pod metadata.

This could be configured by a pair of flags, similar to the existing `annotation/label-tag-regex` flags of cloud providers.
```
[k8s]
annotation-custom-header-regex = '^gostatsd.atlassian.io/(?P<Service-Name>.*)$'
label-custom-header-regex= ''
```
If matched, forwarder will include the header `Service-Name` with the matched substring in the pod annotation.

Do you think this is a feature fit into gostatsd?

hligit:
Added a new configuration `dynamic-headers` for specifying a list of tags to be matched against for added to headers of http request. Metrics are also regrouped by the matching tags.

Fixes #305.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,8 @@ following configuration options:
 - `log-raw-metric`: logs raw metrics received from the network.  Defaults to `false`.
 - `custom-headers` : a map of strings that are added to each request sent to allow for additional network routing / request inspection.
   Not required, default is empty. Example: `--custom-headers='{"region" : "us-east-1", "service" : "event-producer"}'`
+- `dynamic-headers` : similar with `custom-headers`, but the header values are extracted from metric tags matching the
+  provided list of string. Not required, default is empty. Example: `--dynamic-headers='["region", "service"]'`
 
 
 Metric expiry and persistence
--- a/README.md
+++ b/README.md
@@ -89,7 +89,11 @@ following configuration options:
 - `custom-headers` : a map of strings that are added to each request sent to allow for additional network routing / request inspection.
   Not required, default is empty. Example: `--custom-headers='{"region" : "us-east-1", "service" : "event-producer"}'`
 - `dynamic-headers` : similar with `custom-headers`, but the header values are extracted from metric tags matching the
-  provided list of string. Not required, default is empty. Example: `--dynamic-headers='["region", "service"]'`. If a tag is specified in both `custom-header` and `dynamic-header`, the vaule set by `custom-header` takes precedence.
+  provided list of string. Tag names are canonicalized by first replacing underscores with hyphens, then converting
+  first letter and each letter after a hyphen to uppercase, the rest are converted to lower case. If a tag is specified
+  in both `custom-header` and `dynamic-header`, the vaule set by `custom-header` takes precedence. Not required, default
+  is empty. Example: `--dynamic-headers='["region", "service"]'`.
+  This is an experimental feature and it may be removed or changed in future versions.
 
 
 Metric expiry and persistence
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ following configuration options:
 - `custom-headers` : a map of strings that are added to each request sent to allow for additional network routing / request inspection.
   Not required, default is empty. Example: `--custom-headers='{"region" : "us-east-1", "service" : "event-producer"}'`
 - `dynamic-headers` : similar with `custom-headers`, but the header values are extracted from metric tags matching the
-  provided list of string. Not required, default is empty. Example: `--dynamic-headers='["region", "service"]'`
+  provided list of string. Not required, default is empty. Example: `--dynamic-headers='["region", "service"]'`. If a tag is specified in both `custom-header` and `dynamic-header`, the vaule set by `custom-header` takes precedence.
 
 
 Metric expiry and persistence
--- a/metric_map.go
+++ b/metric_map.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -88,6 +88,8 @@ following configuration options:
 - `log-raw-metric`: logs raw metrics received from the network.  Defaults to `false`.
 - `custom-headers` : a map of strings that are added to each request sent to allow for additional network routing / request inspection.
   Not required, default is empty. Example: `--custom-headers='{"region" : "us-east-1", "service" : "event-producer"}'`
+- `dynamic-headers` : similar with `custom-headers`, but the header values are extracted from metric tags matching the
```
MovieStoreGuy, line 91: Do you expect dynamic-headers to override custom headers?

metric_map.go
```diff
@@ -176,6 +177,80 @@ func (mm *MetricMap) Split(count int) []*MetricMap {
...
+			}
+		}
+	}
+	return strings.Join(res, ",")
+}
+
+func (mm *MetricMap) SplitByTags(tagNames []string) map[string]*MetricMap {
+	maps := make(map[string]*MetricMap)
+	if len(tagNames) == 0 {
+		maps[""] = mm
```
MovieStoreGuy, line 196: Should anything be added here if tagsNames is nil/empty?

metric_map.go
```diff
@@ -176,6 +177,80 @@ func (mm *MetricMap) Split(count int) []*MetricMap {
...
+				res = append(res, tv)
+				break
+			}
+		}
+	}
+	return strings.Join(res, ",")
+}
+
+func (mm *MetricMap) SplitByTags(tagNames []string) map[string]*MetricMap {
+	maps := make(map[string]*MetricMap)
```
MovieStoreGuy, line 194: Is this going to be okay if each of these `Each` functions execute async?

pkg/statsd/handler_http_forwarder_v2.go
```diff
@@ -395,7 +403,7 @@ func (hfh *HttpForwarderHandlerV2) serializeAndCompress(message proto.Message) (
 	return buf.Bytes(), nil
 }
 
-func (hfh *HttpForwarderHandlerV2) constructPost(ctx context.Context, logger logrus.FieldLogger, path string, message proto.Message) (func() error /*doPost*/, error) {
+func (hfh *HttpForwarderHandlerV2) constructPost(ctx context.Context, logger logrus.FieldLogger, path string, message proto.Message, dynHeaderTags string) (func() error /*doPost*/, error) {
```
MovieStoreGuy, line 406: It would be better to avoid splinting the string on each construct post call, do we expected these dynamic header keys to change once the application is started?

pkg/statsd/handler_http_forwarder_v2.go
```diff
@@ -156,6 +160,7 @@ func NewHttpForwarderHandlerV2(
 		consolidatedMetrics:   ch,
 		client:                httpClient.Client,
 		headers:               headers,
+		dynHeaders:            dynHeaders,
```
tiedotguy, line 163: `dynHeaders` should be pre-baked to remove the `tagName+":"` in the loop.

- [ ] this

metric_map.go
tiedotguy: `Each` isn't async

README.md
mwhittington21: This is a good point. Documenting which one supersedes the other in case of a clash is useful for users.

pkg/statsd/handler_http_forwarder_v2.go
```diff
@@ -212,15 +217,18 @@ func (hfh *HttpForwarderHandlerV2) Run(ctx context.Context) {
 		case <-ctx.Done():
 			return
 		case metricMaps := <-hfh.consolidatedMetrics:
-			if !hfh.acquireSem(ctx) {
-				return
+			mergedMetricMap := mergeMaps(metricMaps)
+			mms := mergedMetricMap.SplitByTags(hfh.dynHeaders)
+			for dynHeaderTags, mm := range mms {
+				if !hfh.acquireSem(ctx) {
```
mwhittington21, line 223: I may not understand what's going on here fully but something about acquiring the lock during a loop and releasing it in an async go func below seems wrong to me. Seems like it would do the first one and then if the go func didn't execute fast enough just return and skip the rest of the loop iterations?

tiedotguy: It's a semaphore, to stop a flood of goroutines. In practice, for a forwarder, there's not a high volume of requests anyway. It's more valuable at the aggregator tier, where you might have 500 requests going to a vendor, and want to limit that.

As much as possible, there's explicit back pressure in the system (ultimately ending at not reading from the network stack and dropping udp statsd packets).

mwhittington21: Ah I see, so acquireSem will block and not return false. The not is an error condition, gotcha.

tiedotguy: Correct - the only time it returns is when `ctx` is canceled, which only occurs on shutdown.

pkg/statsd/handler_http_forwarder_v2.go
hligit: `dynHeaderTags` includes both tag names and values. The values cannot be decided by gostatsd itself. In the case of deploying gostatsd as a daemonset, it could receive metrics from multiple pods.

metric_map.go
hligit: This shortcuts the case where dynamic-headers is not used. It makes the logic in `func (hfh *HttpForwarderHandlerV2) Run(ctx context.Context)` simpler. Otherwise, it'll need to duplicate the async post function call a bit, e.g.

```
if hfh.dynHeaders {
  for hdrs, mm := maps {
    // take lock
    // async post(mm, hdrs)
  }
} else {
  // take lock
  // async post(mm)
}
```

README.md
hligit: - [ ] update doc
- [ ] make custom-headers override dynamic-headers

pkg/statsd/handler_http_forwarder_v2.go
hligit: This only includes the header/tag names for extracting values from tags. As explained in another comment, the dynamic header values cannot be pre-baked when gostatsd starts. Or maybe I missed your point?

hligit: Ahh, the `dynName +":"` part can definitely be optimized. 👍

hligit: I've pushed a commit to pre-bake the colon.

hligit:
> Can I ask what problem this addresses or a link to the issue?

Please see previous discussion of #305.

hligit:
>  I'd like to see this labelled as an experimental / temporary feature

Good point! I've updated readme/changelog to mark this as experimental.

pkg/statsd/handler_http_forwarder_v2.go
hligit: We'll probably need to review the value for `max-requests` when deploying forwarder as a daemonset.

README.md
MovieStoreGuy: Is there ways to set defaults if the values are missing?

metric_map.go
MovieStoreGuy: Right.. That make sense :)

README.md
hligit: No. What's the use case in your mind?

I think if there is a sensible default value, it should be set by `custom-headers` already.

MovieStoreGuy: Say I am sending build metrics but my instance is extracting `'service-name'` and I have ALB routing rules based on the existence of that header being there but now can not set a default due to the interaction between custom headers vs dynamic headers.

pkg/statsd/handler_http_forwarder_v2.go
```diff
@@ -139,6 +143,24 @@ func NewHttpForwarderHandlerV2(
 		headers[k] = v
 	}
 
+	dynHeaderNamesWithColon := make([]string, 0)
+	for _, name := range dynHeaderNames {
+		if name == "" {
+			continue
+		}
+		isCustomHeader := false
+		for xk := range xheaders {
```
tiedotguy, line 152: This can be done as...
```
if _, ok := xheaders[name]; !ok {
  dynHeaderNamesWithColon = append(dynHeaderNamesWithColon, name+":")
} else {
  logger.WithField("header-name", name).Warn("static and dynamic header defined, using static")
}
```

README.md
tiedotguy: I think this is in the weeds. Build metrics are low volume, no need to over-complicate it.  Remember, this is a workaround until we can scale properly.

README.md
```diff
@@ -88,6 +88,9 @@ following configuration options:
...
+  provided list of string. Not required, default is empty. Example: `--dynamic-headers='["region", "service"]'`. If a tag is specified in both `custom-header` and `dynamic-header`, the vaule set by `custom-header` takes precedence.
```
tiedotguy, line 92: Mention the `_` to `-` replace as well.

hligit: Fixed

pkg/statsd/handler_http_forwarder_v2.go
hligit: Thanks! Updated as suggested.

pkg/statsd/handler_http_forwarder_v2.go
hligit: @MovieStoreGuy @tiedotguy  `dynHeaderTags` in its string form is used as keys for the map of metricsMap splited by tags. The value is also calculated by `tagMatches` when iterating the metricMap.

Ideally the `tagMatch()` helper function should return `[][2]string` to perserve order of tag names. And `splitByTags()` returns `map[[][2]string]*MetricMap`, but slices/map cannot be used as map keys, at least without some gymnastics types. 

Considering this is forwarder and the code is only executed when `dynamic-headers` is enabled, I'm reluctant to implement such optimization.

hligit:
> I think it looks ok, will need a merge/rebase for the changes Adam made yesterday. Do you want to release a -rc1 tag and give it a real test?

👍 Sure. I'll make a candidate release.

Issue #311, opened 2020-06-23
Coveralls needs to be advisory only


tiedotguy:
At some point coveralls was added in a ratchet configuration which means coverage is only ever allowed to improve, or merges are blocked.  There also appears to be issues where the following situation occurs:

1. commit some code, coverage goes down
2. commit some coverage, coverage goes up
3. squash and merge, coverage goes down, master is "red".  That's the state of 56120725587cf82aac5c7e1b0501a8967af45fb4 (Current master as of raising this issue)

I'm not certain if this is the actual sequence of events, but it looks like it might be based on recent commits.

Coveralls should be changed to be advisory only.

thombergs:
I don't know how coveralls was enabled for this repo, but I disabled it now:

![image](https://user-images.githubusercontent.com/1363578/85358173-6b494400-b556-11ea-8ae2-f6c354912bee.png)

tiedotguy:
Thanks @thombergs !

PR #313, opened 2020-06-25
Refactor pprof endpoints in line with go tool expectations

adampetrovic:
All the references on https://golang.org/pkg/net/http/pprof/ assume a pprof path of `/debug/pprof`, updated to reflect this best practice.


adampetrovic:
Looks like the build is failing due to coveralls:

```
goveralls -coverprofile=coverage.out -service=travis-ci
Bad response status from coveralls: 422
{"message":"Couldn't find a repository matching this job.","error":true}
Makefile:73: recipe for target 'coveralls' failed
make: *** [coveralls] Error 1
The command "make coveralls" exited with 2.
```

pkg/web/httpservers.go
```diff
@@ -88,11 +89,11 @@ func NewHttpServer(
 	}
 
 	if enableProf {
-		profiler := &traceProfiler{}
 		routes = append(routes,
-			route{path: "/memprof", handler: profiler.MemProf, method: "POST", name: "profmem_post"},
-			route{path: "/pprof", handler: profiler.PProf, method: "POST", name: "profpprof_post"},
-			route{path: "/trace", handler: profiler.Trace, method: "POST", name: "proftrace_post"},
+			route{path: "/debug/pprof/", handler: pprof.Index, method: "GET", name: "pprof_index"},
```
tiedotguy, line 93: - [ ] removes trace

```
			route{path: "/debug/pprof/trace", handler: pprof.Trace, method: "GET", name: "pprof_trace"},
```

pkg/web/httpservers.go
```diff
@@ -88,11 +89,11 @@ func NewHttpServer(
...
+			route{path: "/debug/pprof/cmdline", handler: pprof.Cmdline, method: "GET", name: "pprof_cmdline"},
+			route{path: "/debug/pprof/profile", handler: pprof.Profile, method: "GET", name: "pprof_profile"},
+			route{path: "/debug/pprof/symbol", handler: pprof.Symbol, method: "GET", name: "pprof_symbol"},
```
tiedotguy, line 96: - [ ] also needs to support POST.

`pprof.Symbol` has explicit handling for POST and non-POST, the other handlers all don't check what method is used.  I'm not sure what `go tool pprof` actually sends though, so it's probably fine.

I'm also not sure if the router will let you just do...
```
			route{path: "/debug/pprof/symbol", handler: pprof.Symbol, method: "POST", name: "pprof_symbol_post"},
```
... or if it will get confused by the duplicate path.  It should be easy enough to change https://github.com/atlassian/gostatsd/blob/master/pkg/web/httpservers.go#L153 to take an array with `.Methods(route.methods...)` though.

tiedotguy:
- [ ] docs and changelog need updating.

That's awkward about coveralls, we had it turned off in #311 because it turned in to a blocking thing, instead of being advisory. I don't have access to even see settings for that anymore, so I'm not sure what the correct fix is.  I guess we can either nuke coveralls from the build, or if there's a way to change the repo/coveralls config to still run but not enforce it, that would be ideal.

I believe @MovieStoreGuy started with the OSS disturbed for that.

tiedotguy:
I raised #314 for coveralls.

pkg/web/httpservers.go
adampetrovic: ah, sorry. I thought it would be captured by the `pprof.Index` call.

pkg/web/httpservers.go
adampetrovic: tested this approach and both GET and POST work fine on the trace endpoint.

adampetrovic:
> * [ ]  docs and changelog need updating.

major version bump?

tiedotguy:
> major version bump?

Yeah, since we're moving an endpoint.  Technically it's a debug endpoint, but numbers are infinite so why not.

PR #317, opened 2020-06-26
Fix no internal stats regression

### LINKED ISSUE #316: #290 introduces flusher bug that causes internal metrics to not be sent
CC @ash2k 

We just upgraded our production release from 20.3.1 to 22.0.0 and noticed that we stopped sending internal metrics about backend events, specifically: https://github.com/atlassian/gostatsd/blob/979ab8493471cbea7625d07fcf48600209bf45a3/pkg/backends/datadog/datadog.go#L146-L151

A debug session shows that the `flushed` channel is never called on a flush.

I did a git bisect and found the offending commit is from https://github.com/atlassian/gostatsd/pull/290

ash2k:
Fixes #316.

### DIFF
--- a/cmd/gostatsd/main.go
+++ b/cmd/gostatsd/main.go
@@ -13,7 +13,6 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/ash2k/stager"
 	"github.com/atlassian/gostatsd"
 	"github.com/atlassian/gostatsd/pkg/backends"
 	"github.com/atlassian/gostatsd/pkg/cachedinstances"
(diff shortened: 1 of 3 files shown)


cmd/gostatsd/main.go
```diff
@@ -59,23 +58,17 @@ func main() {
...
+	s, err := constructServer(v)
 	if err != nil {
 		return err
 	}
 
-	stgr := stager.New()
-	defer stgr.Shutdown()
-	for _, runnable := range runnables {
-		stgr.NextStage().StartWithContext(runnable)
+	profileAddr := v.GetString(ParamProfile)
```
ash2k, line 67: Moved it here - first try to construct the server and only then start profiler.

tiedotguy:
Pushed a changelog, looks good to me.  Hopefully @hligit's change works, because it's going out :)

tiedotguy:
Sigh. I guess coveralls may need to go.

hligit:
> Pushed a changelog, looks good to me. Hopefully @hligit's change works, because it's going out :)

Confirm the change of #310 works in my lab environment.

PR #319, opened 2020-06-29
Adding channel watcher to metricMap queue

MovieStoreGuy:
We've reached some place were where are blocking and it has to determine the issue while missing continuous profiling in our production environment.

I am adding a channel watcher to metricMap queue that is created be worker.


METRICS.md
```diff
@@ -84,6 +84,7 @@ the samples are reset.
 | Channel name              | Additional tags | Description
 | ------------------------- | --------------- | -----------
 | dispatch_aggregator       | aggregator_id   | Channel to dispatch metrics to a specific aggregator.
+| dispatch_receiver         | aggregator_id   | Channel to dispatch singular metrics to a given aggregator.
```
MovieStoreGuy, line 87: I think I borked up the meaning of this channel, let me go fix it,

pkg/statsd/aggregator.go
```diff
@@ -275,14 +275,16 @@ func (a *MetricAggregator) Reset() {
...
+// Receive takes a batched metrics and will put them on the internal aggregator
+// queue to be processed
 func (a *MetricAggregator) Receive(ms ...*gostatsd.Metric) {
 	a.metricsReceived += uint64(len(ms))
 	for _, m := range ms {
 		a.metricMap.Receive(m)
 	}
 }
 
+// ReceiveMap takes a single Me
```
tiedotguy, line 287: Regular Me or Mini-Me?

METRICS.md
tiedotguy: I'd suggest changing the channel names for both, `dispatch_aggregator` is less accurate now because the term could cover either. `dispatch_aggregator_batch` and `dispatch_aggregator_map` maybe.

MovieStoreGuy:
On a tangent @tiedotguy , looks like the protobuf marshaler and pb definitions are using a deprecated package. (from github) and will have to move to the google provided package.

pkg/statsd/aggregator.go
MovieStoreGuy: The me-me

METRICS.md
MovieStoreGuy: That makes sense to me 👍 

Though, the usage of the two channels is still a mystery to me.

tiedotguy:
I saw the PB thing when I was looking at gRPC. The docs were full of references to "use this package", and then the package was like "no".

I dunno if they've been updated since.

tiedotguy:
I made an issue.

pkg/statsd/worker.go
```diff
@@ -47,13 +47,21 @@ func (w *worker) executeProcess(cmd *processCommand) {
 }
 
 func (w *worker) RunMetrics(ctx context.Context, statser stats.Statser) {
```
ash2k, line 49: This method, like other similar ones that do work concurrently, should block until the goroutine below finishes. `wait.Group` can help here.

METRICS.md
tiedotguy: There's two paths to the aggregators:
- UDP receiver -> []*Metric -> a bunch of PipelineHandler.DispatchMetrics -> Aggregator via dispatch_aggregator_batch
- HTTP receiver -> *MetricMap -> a bunch of PipelineHandler.DispatchMetricMap -> Aggregator via dispatch_aggregator_map

There's also the Consolidator who's job it is to receive []*Metric and *MetricMaps, bundle them all up in to a single *MetricMap, and then send them out somewhere.  Currently it's only used for HTTP forwarding, but it could easily be placed in front of the UDP receiver, which is part 1 of removing DispatchMetrics from the pipeline entirely.

Part 2 is dealing with the cloud provider, although it's basically the same thing, just drop a Consolidator in there.

METRICS.md
```diff
@@ -83,7 +83,8 @@ the samples are reset.
 
 | Channel name              | Additional tags | Description
 | ------------------------- | --------------- | -----------
-| dispatch_aggregator       | aggregator_id   | Channel to dispatch metrics to a specific aggregator.
+| dispatch_aggregator_batch | aggregator_id   | Channel to dispatch metrics to a specific aggregator.
+| dispatch_aggregator_map   | aggregator_id   | Channel to dispatch singular metrics to a given aggregator.
```
tiedotguy, line 87: "metric maps", not "singular metrics"

pkg/statsd/worker.go
MovieStoreGuy: It does @ash2k , it just forces one to run within the background on a different routine and will cancel once context is done.

MovieStoreGuy: Unless I have missed something on how contexts work.

ash2k: What I'm saying is this method will return without waiting for `go stats.NewChannelStatsWatcher().Run(ctx)` to terminate. Other places that start goroutines block until the spawned goroutines finish. This is a very important property that allows to write deterministic tests without data races.

ash2k: https://vorpus.org/blog/notes-on-structured-concurrency-or-go-statement-considered-harmful/

MovieStoreGuy: Would it make sense to put both channels into the `wait.StartWithContext` then wait and the end?
Or just move the singular one there?

tiedotguy: @MovieStoreGuy something like this:
```
func (w *worker) RunMetrics(ctx context.Context, statser stats.Statser) {
	stgr := stager.New()

	csw := stats.NewChannelStatsWatcher(...chan1..., statser)
	stgr.NextStage().StartWithContext(csw.Run)
	
	csw = stats.NewChannelStatsWatcher(...chan2..., statser)
	stgr.NextStage().StartWithContext(csw.Run)
	
	<-ctx.Done()
	stgr.Shutdown()
}
```

When the context.Done is signaled, stgr.Shutdown() will cancel each stage, in reverse order, and wait for them to complete.  There's a wait.Group behind the scenes which ensures ordering.

MovieStoreGuy: I ended going for the former since it make the invocation of the watchers the same and neater control flow.

tiedotguy: Works for me.

pkg/statsd/worker.go
```diff
@@ -47,13 +48,22 @@ func (w *worker) executeProcess(cmd *processCommand) {
...
-	csw := stats.NewChannelStatsWatcher(
+	wg := &wait.Group{}
```
ash2k, line 51: nit: This is fine, but usually it's just `var wg wait.Group`

PR #322, opened 2020-06-30
Add pprof wildcard match to enable the remaining pprof endpoints

adampetrovic:
The current implementation doesn't forward catchall routes to the pprof.Index handler. This means we can't make requests like `/debug/pprof/{heap,goroutine}`etc.

This change binds the `/debug/pprof/.*` route after all of our specific routes to ensure the Index handler is used for any other endpoint.


tiedotguy:
Seems legit.

PR #324, opened 2020-07-02
Inject logger

tiedotguy:
This strips out the rest of `log` aliased to `logrus`, as well as injecting a logger mostly through the system.  `stdout` backend skips it still.


pkg/backends/datadog/datadog.go
```diff
@@ -38,7 +38,7 @@ const (
 	// defaultMetricsPerBatch is the default number of metrics to send in a single batch.
 	defaultMetricsPerBatch = 1000
 	// maxResponseSize is the maximum response size we are willing to read.
-	maxResponseSize     = 10 * 1024
+	maxResponseSize     = 1024
```
tiedotguy, line 41: I reduced this (and in NR) because the response is generally minimal, and we log it.  We shouldn't be logging 10k of data, especially since failures are likely to cluster, so we could end up with hundreds of 10k logs.  There's an argument to be made for "read up to 10k, log up to 1k", just to ensure we drain the connection if possible, without reading to infinity, but we can add that if we need it.

pkg/backends/newrelic/newrelic.go
```diff
@@ -314,7 +315,10 @@ func (n *Client) post(ctx context.Context, buffer *bytes.Buffer, data interface{
 			return fmt.Errorf("[%s] %v", BackendName, err)
 		}
 
-		log.Warnf("[%s] failed to send, sleeping for %s: %v", BackendName, next, err)
+		n.logger.WithFields(logrus.Fields{
```
MovieStoreGuy, line 318: Missing backend name?

tiedotguy: Backend name is now injected higher up in the system, in `InitBackend`.  The `logger` passed already has a name.

When everything is enabled it looks something like this wondrous monstrosity:
```
INFO[0000] Starting server
INFO[0000] No cloud provider specified
INFO[0000] created transport                             dialer-keep-alive=30s dialer-timeout=5s enable-http2=false idle-connection-timeout=1m0s max-idle-connections=50 name=default network=tcp tls-handshake-timeout=3s
INFO[0000] created client                                client-timeout=10s name=default type=http
INFO[0000] created backend                               backend=datadog compress-payload=true max-request-elapsed-time=15s max-requests=24 metrics-per-batch=1000
INFO[0000] Initialised backend                           backend=datadog
INFO[0000] created backend                               address="localhost:2003" backend=graphite counter-namespace=stats.counters dial-timeout=5s gauges-namespace=stats.gauges global-suffix= mode=tags sets-namespace=stats.sets timer-namespace=stats.timers write-timeout=30s
INFO[0000] Initialised backend                           backend=graphite
INFO[0000] Initialised backend                           backend=null
INFO[0000] created backend                               address="127.0.0.1:1234" backend=statsdaemon dial-timeout=5s write-timeout=30s
INFO[0000] Initialised backend                           backend=statsdaemon
INFO[0000] Initialised backend                           backend=stdout
INFO[0000] Initialised backend                           backend=cloudwatch
INFO[0000] internal metrics OFF, to enable set 'statser-type' to 'logging' or 'internal'  backend=newrelic
INFO[0000] flushInterval default (10s)                   backend=newrelic
INFO[0000] created backend                               backend=newrelic flush-interval=10s max-request-elapsed-time=15s max-requests=24 metrics-per-batch=1000
INFO[0000] Initialised backend                           backend=newrelic
INFO[0000] created transport                             dialer-keep-alive=2s dialer-timeout=30s enable-http2=false idle-connection-timeout=4s max-idle-connections=5 name=influxdb network=tcp4 tls-handshake-timeout=6s
INFO[0000] created client                                client-timeout=10s name=influxdb type=http
INFO[0000] created backend                               api-endpoint="http://influxdb-1.8.0.local/" api-version=2 backend=influxdb bucket=derp compress-payload=false credentials="(set)" max-request-elapsed-time=15s max-requests=120 metrics-per-batch=5000 org=org transport=influxdb user-agent=gostatsd
INFO[0000] Initialised backend                           backend=influxdb
```

ash2k:
This is good improvement but have you considered also moving to something like zap and dropping logrus? logrus is in maintenance mode https://github.com/Sirupsen/logrus

why zap
- [zaptest](https://github.com/uber-go/zap/tree/master/zaptest) is very convenient for logging into t.Log() rather than to the console. See e.g. https://github.com/atlassian/smith/blob/98bdc61e18aa963c2b47d16fd84b51e2bf48e0fb/pkg/specchecker/builtin/process_deployment_test.go#L138-L141
  - Can easily tell which test logged what
  - Concurrent tests don't clobber each other's logs
  - Only see logs from failed tests
  - Can capture logs for analyzis (if you really need this...)
- Allows to write type safe field helpers (vs `WithField(string, interface{})` in logrus). E.g. https://github.com/atlassian/smith/blob/98bdc61e18aa963c2b47d16fd84b51e2bf48e0fb/pkg/util/logz/logz.go#L9 or https://github.com/atlassian/ctrl/blob/master/logz/logz.go
  - No need to remember to use a constant for field name as it's encoded in the field helper. This is great for log consistency
- Not important (?), but it's more performant

I'm not familiar with other frameworks, there are likely other good options. Used zap and would use again :)

tiedotguy:
I've briefly looked at zap, but more for greenfield stuff. I agree moving off logrus would be good, but wasn't really in scope for what I was aiming for here (general consistency).  I'll raise an issue.

pkg/backends/newrelic/newrelic.go
MovieStoreGuy: Neat

tiedotguy:
Thought it was sufficient to push the same code to trick coveralls, might need an actual new commit.

MovieStoreGuy:
I agree with moving off logrus to zap, have started adopting it for some of our newer projects and haven't seen any negative impacts.

tiedotguy:
The system works.

PR #326, opened 2020-07-02
Fix the usage of a virtual clock in backoff handler

tiedotguy:
`backoff.NewExponentialBackOff()` initializes an `ExponentialBackOff`, including grabbing
the current time from the system clock.  During tests when we override the clock, this
effectively puts us back in time to 1970.

Eventually all this adhoc retry logic will be centralized when the Transport system
is completed (which handles it properly already)


tiedotguy:
I know, right?

PR #327, opened 2020-07-05
Adds InfluxDB backend

### LINKED ISSUE #14: Add influxdb backend

tiedotguy:
Adds support for sending to InfluxDB as a backend with both pre-1.8 and post-1.8 API.

@adampetrovic - I vaguely recall you've spent a bunch of time with InfluxDB, if you could sanity check the "am I using InfluxDB correctly?" aspects, it would be much appreciated.

Closes #14


BACKENDS.md
```diff
@@ -61,6 +61,122 @@ When `mode` is `legacy`, the graphite backend will emit metrics with the followi
...
+
+Using v1 settings in v2 or v2 settings in v1 will display a warning on startup and ignore the setting.
+
+
+### Tag normalization
+As the dogstatsd protocol doesn't have exact parity with InfluxDBs line protocol, there is some cleanup applied to the
+tags before sending them.  Specifically:
+- tags with only a value are normalized as `unnamed:value`
+- multiple values for duplicate keys are sorted
+- keys with multiple values have their values concatenated with `__`
```
MovieStoreGuy, line 139: Is this something that we would consider making configurable in future?

tiedotguy: If someone comes up with a reason to have multiple implementations, sure.  What I'm hoping is that someone more familiar with InfluxDB raises an issue to say "oh this has a best practice / standard way that everyone knows, check this link", and then we just move to that and tell anyone else they should follow best practices.

Ideally we'll only ever have one way, to keep things simple.

pkg/backends/influxdb/flush.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	sb.WriteByte(' ')
+	return sb.String()
+}
+
+// escape will escape NL (\n), CR (\r), TAB (\t), space, comma, backslash, and equals
+// writing the results to the supplied strings.Builder.
+func escapeTagToBuilder(sb *strings.Builder, s string) {
+	for _, ch := range s {
+		if ch == '\n' {
```
MovieStoreGuy, line 94: Would this be more efficient using a switch instead of if-else chain?

tiedotguy: I'd expect the difference to be down in the noise, if there is one.  Also this is only called on flush, it's not a hot path.

pkg/backends/influxdb/flush.go
```diff
@@ -0,0 +1,256 @@
...
+		}
+		sb.WriteRune(ch)
+	}
+	sb.WriteByte('"')
+}
+
+// writeName writes "measurement,tag1=value1,tag2=value2,... " to the output buffer.
+// Common to all metric types and events.
+func writeName(w io.Writer, name string, tags gostatsd.Tags) {
+	_, _ = w.Write([]byte(formatNameTags(name, tags)))
```
MovieStoreGuy, line 144: Is discarding the write error worth while here? 
I worry this could bite us later.

tiedotguy: Yep, errors are propagated out later when we flush/close.  The gzip writer has a "sticky" error.

https://blog.golang.org/errors-are-values

pkg/backends/influxdb/flush.go
```diff
@@ -0,0 +1,256 @@
...
+	if !f.disabledSubtypes.Sum {
+		sb.WriteString(fmt.Sprintf("sum=%g,", timer.Sum))
+	}
+	if !f.disabledSubtypes.SumSquares {
+		sb.WriteString(fmt.Sprintf("sum_squares=%g,", timer.SumSquares))
+	}
+	for _, pct := range timer.Percentiles {
+		sb.WriteString(fmt.Sprintf("%s=%g,", pct.Str, pct.Float))
+	}
+	if sb.Len() == 0 { // wat
```
MovieStoreGuy, line 200: Nice comment m8

pkg/backends/influxdb/flush.go
```diff
@@ -0,0 +1,256 @@
...
+func (f *flush) flush() {
+	// I don't believe that a gzip.Writer sending to a bytes.Buffer can actually fail, but the
+	// flate.huffmanBitWriter at the core of the object graph *does* have an internal error it
+	// can theoretically produce, so we won't ignore the error.
+	err := f.writer.Close()
+	if err == nil {
+		f.cb(f.buffer, f.metricCount)
+	} else {
+		atomic.AddUint64(f.errorCounter, 1)
+		f.releaseBuffer(f.buffer)
```
MovieStoreGuy, line 252: Is the release buffer a paradigm we have used throughout the project?

pkg/backends/influxdb/influxdb.go
```diff
@@ -0,0 +1,502 @@
...
+	if maxRequests == 0 {
+		return nil, errMaxRequestsIsNotPositive
+	}
+	if maxRequestElapsedTime <= 0 && maxRequestElapsedTime != -1 {
+		return nil, errMaxRequestElapsedTimeInvalid
+	}
+	if metricsPerBatch == 0 {
+		return nil, errMetricsPerBatchIsNotPositive
+	}
+	parsedEndpoint, err := url.Parse(apiEndpoint)
```
MovieStoreGuy, line 150: Is it worthwhile checking the scheme to set with the url?
assuming it has to be `http` or `https`?

tiedotguy: Nah. If you want to post to an ftp site, then good luck!  It's up to the http.Client to cleanup that mess, not us.

pkg/backends/influxdb/flush.go
tiedotguy: Who am I to question if some sysadmin has decided to disable all timers!

pkg/backends/influxdb/flush.go
tiedotguy: Eh... not as explicit as this.  We do have the "a semaphore is a channel of buffers" though.

I'm not a huge fan of this honestly, but it was the cleanest way I found to keep `processMetrics` small-ish.

pkg/backends/influxdb/influxdb.go
tiedotguy: For what it's worth, InfluxDB apparently supports ingestion via http(s) over unix domain socket. I have no idea if this will successfully talk to it, but it probably would, since they're both go, so they should have compatible notions of how to talk over it.

pkg/backends/influxdb/influxdb.go
```diff
@@ -0,0 +1,502 @@
...
+		return err
+	}
+	atomic.AddUint64(&idb.seriesSent, seriesCount)
+	return nil
+}
+
+func writeEvent(w io.Writer, e *gostatsd.Event) {
+	hasHost := false
+	var eventTags []string
+	tags := make(gostatsd.Tags, 0, len(e.Tags)+4)
```
MovieStoreGuy, line 345: Why `+4` on the end here?

pkg/backends/influxdb/influxdb.go
```diff
@@ -0,0 +1,502 @@
...
+}
+
+func (idb *Client) SendEvent(ctx context.Context, e *gostatsd.Event) error {
+	buf, writer := idb.getBuffer(ctx)
+	if buf == nil {
+		return ctx.Err()
+	}
+	defer idb.releaseBuffer(buf)
+
+	writeEvent(writer, e)
```
MovieStoreGuy, line 396: The write error should be captured here

pkg/backends/influxdb/influxdb.go
tiedotguy: 4 extra tags we expect to add - priority, alerttype, etc.  I can add a comment.

pkg/backends/influxdb/influxdb.go
tiedotguy: I have a method for handling errors, which I'm tentatively calling "draeaswh", or "don't return an error, and see what happens".

That's why `writeEvent` has no error return value, but we do grab it from `writer.Close()` on the next line.

pkg/backends/influxdb/influxdb_test.go
```diff
@@ -0,0 +1,525 @@
...
+	return &gostatsd.MetricMap{
+		Timers: gostatsd.Timers{
+			"t1": map[string]gostatsd.Timer{
+				"tag2": {
+					Values:    []float64{0, 1},
+					Timestamp: gostatsd.Nanotime(200),
+					Hostname:  "h2",
+					Tags:      gostatsd.Tags{"tag2", "gsd_histogram:20_30_40_50_60"},
+					Histogram: map[gostatsd.HistogramThreshold]int{
+						20:                                       5,
```
MovieStoreGuy, line 398: That spacing...

MovieStoreGuy:
Everything seems reasonable but I have never worked with influxdb in the past so I am not sure if the client's behaviour is what is expected.

pkg/backends/influxdb/influxdb_test.go
tiedotguy: Bow before `gofmt`.

tiedotguy:
It works at least:
image

(there's also events working, but they're not in view)

I've also got the callout for "please report issues".  In theory it should work with enterprise and all that jazz, but I have no way to confirm that.

adampetrovic:
Sorry, was off sick today. Everything looks good to me from a semantic point of view. It has been a while since I played with InfluxDB :P

tiedotguy:
No worries, thanks for the input. Get better soon!

PR #329, opened 2020-07-06
Aligned time

tiedotguy:
This should be 3 PRs, but that's just a bunch of cycling through PRs for not a lot of point.

Each commit can be assessed completely on its own, and it's the last of what I've got pending. More to come over the coming days though, I'm sure.


tiedotguy:
Not sure why my system has so much trouble with the linter, thought I fixed it after last time.

README.md
```diff
@@ -55,13 +55,63 @@ The server can currently run in two modes: `standalone` and `forwarder`.  It is
...
+- `max-parsers`: the number of workers available to parse metrics.  Defaults to the number of logical cores.
+- `max-workers`: the number of aggregators to process metrics.  Defaults to the number of logical cores.
+- `max-queue-size`: the size of the buffers between parsers and workers.  Defaults to `10000`, monitored via
+  `channel.*` metric, with `dispatch_aggregator_batch` and `dispatch_aggregator_map` channels.
+- `max-concurrent-events`: the maximum number of concurrent events to be dispatching.  Defaults to `1024`, monitored
+  via `channel.*` metric, with `backend_events_sem` channel.
+- `estimated-tags`: provides a hint to the system as to how many tags are expected to be seen on any particular metric,
+  so that memory can be pre-allocated and reducing churn.  Defaults to `4`.  Note: this is only a hint, and it is safe
+  to send more.
+- `log-raw-metric`: logs raw metrics received from the network.  Defaults to `false`.
```
MovieStoreGuy, line 85: Should this be deprecated in favor of the stdout backend?

README.md
```diff
@@ -55,13 +55,63 @@ The server can currently run in two modes: `standalone` and `forwarder`.  It is
...
+- `metrics-addr`: the address to listen to metrics on. Defaults to `:8125`.
```
MovieStoreGuy, line 86: What receiver does this refer to? The statsd endpoint or the internal gostatsd endpoint?

README.md
tiedotguy: Nah, that's post-aggregation, this is specifically for pre-aggregation, I believe it logs the metric the client send after parsing.  I believe @beanliu needed it for something.

cover.sh
```diff
@@ -1,36 +1,3 @@
...
-    tail -n +2 coverage_tmp.out >> coverage.out 2> /dev/null ||:
-done
-
-rm -f coverage_tmp.out
-
-if [ ! -z "$ERROR" ]
-then
-    die "Encountered error, last error was: $ERROR"
-fi
+set -euo pipefail
```
tiedotguy, line 2: Current master build is failing because when I added InfluxDB, I added missing packages to the cover.sh.  Those packages were then removed in the subsequent refactoring, which means the coverage fails to run.

While unplanned, this new cover.sh will fix master when it's merged.

README.md
tiedotguy: UDP endpoint only

README.md
```diff
@@ -55,13 +55,63 @@ The server can currently run in two modes: `standalone` and `forwarder`.  It is
...
+
+
 In `forwarder` mode, raw metrics are collected from a frontend, and instead of being aggregated they are sent via http
 to another gostatsd server after passing through the processing pipeline (cloud provider, static tags, filtering, etc).
 
 A `forwarder` server is intended to run on-host and collect metrics, forwarding them on to a central aggregation
 service.  At present the central aggregation service can only scale vertically, but horizontal scaling through
 clustering is planned.
 
+Aligned flushing is deliberately not supported in `forwarder` mode, as it would impact the central aggregation server
```
MovieStoreGuy, line 111: Also, distributed sync is hard? :badpokerface:

tiedotguy: For forwarder it's more about "if 1000 hosts all send at exactly 8:34:20pm, then the central server blows up" vs "if 100 send at 8:34:30, 100 send at 8:34:31, 100 send at 8:34:32, etc...".  It's not even that bad because it'll be skewed out across milliseconds, not seconds.

But even then that's not the actual point.  The point is that if you have two independent services in us-east-1 and us-west-2, and they're out of phase by 5 seconds, it makes the backend a bit wonky when you query it.

So it's really about the **timestamp** on the data, not the time the data is sent.

internal/util/aligned_ticker_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+	"github.com/atlassian/gostatsd/internal/fixtures"
+)
+
+const ms = int64(time.Millisecond)
+
+func checkTime(t *testing.T, ctx context.Context, ch <-chan time.Time, expected time.Time) {
+	select {
+	case <-ctx.Done():
+		t.Error("timed out")
```
MovieStoreGuy, line 19: This could be moved to `assert.FailNow(t, `timed out`)`

tiedotguy: I though there was a better way, but I'm so used to `require` and `assert` that I couldn't think of it.  I'll fix it.

tiedotguy: Oh, `assert.FailNow`, not `t.FailNow("timed out")`, I see.  Groovy.

internal/util/aligned_ticker.go
```diff
@@ -0,0 +1,99 @@
+package util
+
+import (
+	"context"
+	"time"
+
+	"github.com/tilinna/clock"
+)
+
+// AlignedTicker is a wrapper around time.Ticker will align the time on a specific interval.  The first tick will
```
MovieStoreGuy, line 10: Is the assumption here that the system clock is in sync with an NTP clock?

tiedotguy: Not really, there's nothing high precision here.  It's just about that sweet sweet xx:xx:x0 when you query the data.

I did consider fudging the timestamp (ie, round it up), rather than aligning the time, but that's bad for users, because they don't know which window they landed in, because they don't know the true window.

PR #330, opened 2020-07-07
Refactor source address

tiedotguy:
Writing the description of how `Source` is handled it possibly the most depressing thing I've done in this code base.  I don't even recall why I needed to do this anymore, but I know there was a good reason, which I'll remember tomorrow.

I'm very tempted to go one more level and remove `Source` as a field entirely, replacing it with a `Source()` function which just uses the `host` tag.

Generally the first 3 commits can be ignored, with the actual changes living in e8dedc3 and 06592ea.


pkg/backends/newrelic/flush.go
```diff
@@ -23,7 +23,7 @@ type timeSeries struct {
 }
 
 // addMetric adds a metric to the series.
-func (f *flush) addMetric(n *Client, metricType string, value float64, persecond float64, hostname string, tags gostatsd.Tags, name string) {
+func (f *flush) addMetric(n *Client, metricType string, value float64, persecond float64, tags gostatsd.Tags, name string) {
```
MovieStoreGuy, line 26: We don't need the hostname here? Is it just being pushed into the tags?

tiedotguy: It's never consumed at all, probably copied and pasted from the DD backend which does give it special usage (which is probably why the notion exists at all).

MovieStoreGuy: Sounds good to me.

tiedotguy:
That's just the first handful of commits. Check the two called out specifically for real changes.  There's some logic that's different in there.

MovieStoreGuy:
Double checked the highlighted commits and see nothing that is of concern. Please go ahead and merge

tiedotguy:
Awesome, cheers mate.

Issue #210, opened 2019-01-29
Refactoring aggregators


tiedotguy:
An issue we have is hot aggregators - the same metric name overwhelming a single aggregator.  As it's handled on a single aggregator, it can't scale.

I've come up with an alternate approach, which I believe will enable scaling.  The core data structure (which we'll call Consolidator until a better name comes up) is a slice of channel of MetricMaps and an index:
```
type Consolidator struct {
  next uint64 // atomic
  maps []chan *gostatsd.MetricMap
}
```

Data is submitted to the Consolidator by atomically incrementing the index, reading from the channel with the returned index (modulo the size of the slice), adding any metrics to the map returned, and sending the map back.

When it's time to do actual aggregation, the flusher will receive from each channel, and send a new map.

All the maps can then be combined and aggregated (multiple goroutines optional).  If they're combined in to a master MetricMap, then the Reset will still apply, allowing persistence to still exist.

Properties:
- UDP is non-blocking (except for the swap), even under extreme load with a single metric name, as the number of goroutines is fixed.
- HTTP can potentially block, as the number of goroutines is unbounded.
- No back pressure.  A large number of tag sets would eat memory.  I'm not even sure how it could be added explicitly.

tiedotguy:
Benchmarking indicates it's slightly better to use a regular buffered chan.  I initially avoided that to reduce fan-in pressure, however it appears to be unnecessary as there's enough work done between channel access.  There's practically no code difference anyway, and the intent is clearer, so a buffered channel should be used instead.

There may be a hazard with flushing though, as it needs to drain the entire channel before adding new maps, rather than swapping.

tiedotguy:
The hazard with flushing is good, as it allows us to create back pressure, by not pushing new maps until processing of the old ones is done.

PR #331, opened 2020-07-10
Remove dispatch metrics

tiedotguy:
There's some juggling with the benchmarks, which is primarily because they're now creating a MetricMap in the loop, which is not really valid.  But they're in there so they're staying.

As always, commits are generally isolated.

### DIFF
--- a/pkg/stats/flush_notifier.go
+++ b/pkg/stats/flush_notifier.go
@@ -1,6 +1,7 @@
 package stats
 
 import (
+	"context"
 	"sync"
 	"time"
 )
@@ -35,7 +36,7 @@ func (fn *flushNotifier) RegisterFlush() (ch <-chan time.Duration, unregister fu
 
 // NotifyFlush will notify any registered channels that a flush has completed.
 // Non-blocking, thread-safe.
-func (fn *flushNotifier) NotifyFlush(d time.Duration) {
+func (fn *flushNotifier) NotifyFlush(ctx context.Context, d time.Duration) {
 	fn.lock.RLock()
 	defer fn.lock.RUnlock()
 	for _, hook := range fn.flushTargets {
--- a/pkg/stats/statser_internal.go
+++ b/pkg/stats/statser_internal.go
@@ -2,7 +2,6 @@ package stats
 
 import (
 	"context"
-	"sync/atomic"
 	"time"
 
 	"github.com/atlassian/gostatsd"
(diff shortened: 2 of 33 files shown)


pkg/stats/flush_notifier.go
```diff
@@ -35,7 +36,7 @@ func (fn *flushNotifier) RegisterFlush() (ch <-chan time.Duration, unregister fu
 
 // NotifyFlush will notify any registered channels that a flush has completed.
 // Non-blocking, thread-safe.
-func (fn *flushNotifier) NotifyFlush(d time.Duration) {
+func (fn *flushNotifier) NotifyFlush(ctx context.Context, d time.Duration) {
```
MovieStoreGuy, line 39: `ctx` isn't used within this function

pkg/stats/statser_internal.go
```diff
@@ -37,30 +33,26 @@ func NewInternalStatser(tags gostatsd.Tags, namespace string, hostname gostatsd.
...
-			return
-		case m := <-is.buffer:
-			is.dispatchMetric(ctx, m)
-		case <-flushed:
-			is.Gauge("internal_dropped", float64(atomic.LoadUint64(&is.dropped)), nil)
-		}
+func (is *InternalStatser) NotifyFlush(ctx context.Context, d time.Duration) {
+	mms := is.consolidator.Drain(ctx)
+	if mms == nil {
+		// context is canceled
```
MovieStoreGuy, line 49: You would only know that for sure if you checked `ctx.Error()` right?

pkg/statsd/aggregator_test.go
```diff
@@ -1,13 +1,10 @@
 package statsd
 
 import (
-	"context"
 	"math"
-	"runtime"
 	"testing"
 	"time"
 
-	"github.com/ash2k/stager"
```
MovieStoreGuy, line 10: rip

pkg/statsd/parser.go
```diff
@@ -110,9 +110,13 @@ func (dp *DatagramParser) Run(ctx context.Context) {
 				accumE += eventCount
 				accumB += badLineCount
 			}
+			// TODO: Refactor this to use a MetricConsolidator
```
MovieStoreGuy, line 113: :stare: is there an issue?

pkg/stats/flush_notifier.go
tiedotguy: It was added to the Statser interface so that InternalStatser had a context, which it needs to drain the MetricConsolidator.

pkg/stats/statser_internal.go
tiedotguy: I'm breaking the abstraction and saying "yeah, I know why this failed". It's part of my DRAEASWH system.

pkg/statsd/parser.go
tiedotguy: Yep, #210 

The entirety of this PR was so that when I do #210 it's a small and easy change.

pkg/stats/flush_notifier.go
MovieStoreGuy: Could you not add it to the `select` block done bellow?

tiedotguy: Ahh, I see what you mean.  Technically yes, but there's really no point. The loop is basically non-blocking because of the `default:` clause.

Remember the primary purpose of cancelling a context is to signal a long-running call to finish - this isn't long running.

PR #332, opened 2020-07-13
This adds a new load tester

tiedotguy:
The load tester in the repo currently is not very configurable, yet still overly complicated.  This introduces a completely fresh rewrite, which is both more configurable for testing different things, and simpler.

I'm a bit iffy on using go-flags, but I find it to be infinitely nicer than viper (and infinitely squared nicer than the golang flag library).  Since there's no need for the more complicated options provided by viper, I went with it.

### DIFF
--- a/cmd/loader/args.go
+++ b/cmd/loader/args.go
@@ -1,0 +1,100 @@
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/jessevdk/go-flags"
+)
+
+type commandOptions struct {
+	Target       string `short:"a" long:"address"                 default:"127.0.0.1:8125" description:"Address to send metrics"                 `
+	MetricPrefix string `short:"p" long:"metric-prefix"           default:"loadtest."      description:"Metric name prefix"                      `
+	MetricSuffix string `          long:"metric-suffix"           default:".%d"            description:"Metric suffix with cardinality marker"   `
+	Rate         uint   `short:"r" long:"rate"                    default:"1000"           description:"Target packets per second"               `
+	DatagramSize uint   `          long:"buffer-size"             default:"1500"           description:"Maximum size of datagram to send"        `
+	Workers      uint   `short:"w" long:"workers"                 default:"1"              description:"Number of parallel workers to use"       `
+	Counts       struct {
+		Counter uint64 ` short:"c" long:"counter-count"                                    description:"Number of counters to send"              `
+		Gauge   uint64 ` short:"g" long:"gauge-count"                                      description:"Number of gauges to send"                `
+		Set     uint64 ` short:"s" long:"set-count"                                        description:"Number of sets to send"                  `
+		Timer   uint64 ` short:"t" long:"timer-count"                                      description:"Number of timers to send"                `
+	} `group:"Metric count"`
+	NameCard struct {
+		Counter uint `             long:"counter-cardinality"     default:"1"              description:"Cardinality of counter names"            `
+		Gauge   uint `             long:"gauge-cardinality"       default:"1"              description:"Cardinality of gauges names"             `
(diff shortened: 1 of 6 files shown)


tiedotguy:
image

cmd/loader/args.go
```diff
@@ -0,0 +1,100 @@
...
+		Gauge   uint `             long:"gauge-value-limit"       default:"1"              description:"Maximum value of gauges"                 `
+		Set     uint `             long:"set-value-cardinality"   default:"1"              description:"Maximum number of values to send per set"`
+		Timer   uint `             long:"timer-value-limit"       default:"1"              description:"Maximum value of timers"                 `
+	} `group:"Value range"`
+}
+
+func parseArgs(args []string) commandOptions {
+	var opts commandOptions
+	parser := flags.NewParser(&opts, flags.HelpFlag | flags.PassDoubleDash)
+	parser.LongDescription = "" + // because gofmt
```
MovieStoreGuy, line 46: Why not use `` string? (I can't remember its name)

cmd/loader/generation.go
```diff
@@ -0,0 +1,96 @@
...
+				sb.WriteByte(',')
+			}
+			sb.WriteString(fmt.Sprintf("tag%d:%d", idx, r.Intn(int(c))))
+		}
+	}
+	sb.WriteByte('\n')
+}
+
+func (mg *metricGenerator) nextCounter(sb *strings.Builder) {
+	atomic.AddUint64(&mg.counters.count, ^uint64(0))
```
MovieStoreGuy, line 47: Can I ask why we are using the max unit64 value?
Is that just going to overflow the value?

cmd/loader/main.go
```diff
@@ -0,0 +1,127 @@
...
+)
+
+func main() {
+	opts := parseArgs(os.Args[1:])
+
+	pendingWorkers := make(chan struct{}, opts.Workers)
+	metricGenerators := make([]*metricGenerator, 0, opts.Workers)
+	for i := uint(0); i < opts.Workers; i++ {
+		generator := &metricGenerator{
+			rnd: rand.New(rand.NewSource(rand.Int63())),
```
MovieStoreGuy, line 21: :fry: What? 

Would this be better to seed with `time.Now.UnixNano()`?

cmd/loader/args.go
tiedotguy: Habit mainly.  And it doesn't look good completely left aligned in the source, but if you add spaces then go-flags strips it.  Either way, it's sub-optimal, so I went with the "embrace the suck" style.

cmd/loader/main.go
```diff
@@ -0,0 +1,127 @@
...
+				timers += atomic.LoadUint64(&mg.timers.count)
+			}
+			fmt.Printf("%d counters, %d gauges, %d sets, %d timers\n", counters, gauges, sets, timers)
+		}
+	}
+}
+
+func sendMetricsWorker(
+	address string,
+	bufSize uint,
```
MovieStoreGuy, line 85: Type not needed for this line.

cmd/loader/generation.go
tiedotguy: This is how you do an atomic subtract from an unsigned - `^0` === max unsigned === `-1` signed.

https://golang.org/pkg/sync/atomic/#AddUint64

cmd/loader/main.go
tiedotguy: `time.Now()` makes no guarantee of advancing when called rapidly in succession.  In all fairness, `rand.Int63()` makes no guarantee of not returning the same number twice in a row either.

cmd/loader/main.go
tiedotguy: When every param is on its own line, it's neater.  Also it means if params change then you get cleaner diffs, kind of like having a trailing `,` on the last parameter.

PR #333, opened 2020-08-06
Patch: Discard values that can not be represented by JSON

MovieStoreGuy:
There has been an issue trying to marshal due to the fact that +- Inf
and NaN break the JSON spec and caused data to be dropped.

The alternative I had in mind was that the aggregators cap values to be the max or min value that is represented by a float64 instead of Inf.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+28.1.0
+------
+- The Datadog backend will discard values that result in NaN or ±Inf since they can not be used with our JSON processing.
+
 28.0.0
 ------
 - Internal pipeline refactor removes two internal metrics
--- a/pkg/backends/datadog/flush.go
+++ b/pkg/backends/datadog/flush.go
@@ -2,6 +2,7 @@ package datadog
 
 import (
 	"fmt"
+	"math"
 
 	"github.com/atlassian/gostatsd"
 )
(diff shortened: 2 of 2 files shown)


MovieStoreGuy:
Related issue within the json library used here: https://github.com/json-iterator/go/issues/365

tiedotguy:
Seems legit.

tiedotguy:
I'd consider a log (with rate limiting) too, to find out where it's coming from. I don't **think** we can do a divide by zero.  If it is coming from a client, ideally it would be picked up on parse and rejected as an invalid value, that way (in a forwarder based deployment) the "failed parse: infinity" log would be on the local forwarder (where a service owner would see it), and not make it to the aggregators logs.

On the flip side, maybe other backends support infinities, so it's perfectly valid to allow it to propagate. I'm not really sure.

adampetrovic:
Agree on the addition of logging.

My gut feel is that this is coming as a result of aggregation rather than +Inf coming from the client. If you add two sufficiently large values together as a result of an aggregated sum or even sum of squares it's somewhat trivial to reproduce the issue.

tiedotguy:
Good point that they could do it, however I believe in Atlassian sum of squares (and maybe sum?) are disabled. So while they're calculated, they wouldn't be emitted.  Looking in lexer.go, we use [strconv.ParseFloat](https://golang.org/pkg/strconv/#ParseFloat), and discard NaNs, but accept +/-inf.  My instinct when I see a +inf is divide by zero, and we don't have that in the aggregator, hence my assumption of client.

Summing or counting enough values to get over 1.8e308 also suggests that something has gone wrong in a client somewhere, so regardless, as good stewards of metrics we should let them know.

I guess ultimately this is why we need the hard data of logs, rather than guessing and speculation :)

PR #334, opened 2020-08-07
Introduces sparse internal metrics

tiedotguy:
These are gauges which are only sent when the value changes. They're sent for several flush intervals, with the expectation that they are attached to error counters, so may need to survive a bursty failures.

Overall this will reduce the amount of data sent in the happy path when deployed in a forwarder configuration (2 metrics per host).


METRICS.md
```diff
@@ -52,7 +52,6 @@ Metrics:
 | cloudprovider.cache_refresh_positive        | gauge (cumulative)  |                              | The cumulative number of positive refreshes
 | cloudprovider.cache_refresh_negative        | gauge (cumulative)  |                              | The cumulative number of refreshes which had an error refreshing and used old data
 | cloudprovider.cache_hit                     | gauge (cumulative)  |                              | The cumulative number of cache hits (host was in the cache)
-|                                             |                     |                              | in progress which completed)
```
tiedotguy, line 55: :hawkward:

pkg/backends/datadog/datadog.go
```diff
@@ -56,11 +56,11 @@ var (
...
-	batchesCreated uint64 // Accumulated number of batches created
-	batchesRetried uint64 // Accumulated number of batches retried (first send is not a retry)
-	batchesDropped uint64 // Accumulated number of batches aborted (data loss)
-	batchesSent    uint64 // Accumulated number of batches successfully sent
-	seriesSent     uint64 // Accumulated number of series successfully sent
+	batchesCreated uint64            // Accumulated number of batches created
+	batchesDropped uint64            // Accumulated number of batches aborted (data loss)
+	batchesSent    uint64            // Accumulated number of batches successfully sent
+	seriesSent     uint64            // Accumulated number of series successfully sent
+	batchesRetried stats.ChangeGauge // Accumulated number of batches retried (first send is not a retry)
```
tiedotguy, line 63: I re-ordered these to make it explicit that the uint64s are at the top (required for atomic alignment).  Technically not required because stats.ChangeGauge is all uint64s as well, but I don't want to bank on that.

tiedotguy:
Looking at it again, I realize it's not as good as it sounds - the sparse backend metric will only be on the aggregation server, not the forwarder.  But still, it's something.  I've spoken to @MovieStoreGuy a bit about updating `http.forwarder.*`, but they're counters, so it's likely to break dashboards if they're changed to sparse gauges.

PR #336, opened 2020-08-12
Document how the project uses semver

### LINKED ISSUE #335: Document version system used
I found that this project already [supports Go modules](https://github.com/atlassian/gostatsd/blob/master/go.mod). But sadly, the [tags](https://github.com/atlassian/gostatsd/tags) doesn't follow [Semantic Versioning](https://semver.org), which means that all tags of this project will be ignored by Go modules and replaced by pseudo-versions, go get acts weirdly when tags are not in that form. It would be great to have the tagged release be named in the format vX.X.X format so that go mod can read it. 
```
	github.com/atlassian/gostatsd v0.0.0-20200807124358-416562efa755
```
Else the mod file shows something like `github.com/atlassian/gostatsd v0.0.0-20200807124358-416562efa755` which is not very readable and difficult to upgrade. It’s hard to verify which version is in use. This is not conducive to version control.

So, I propose this project to follow Semantic Versioning in future versions. For example, `v1.0.1`, `v2.0.0`, `v3.1.0-alpha`, `v3.1.0-beta.2`etc.

tiedotguy:
I'm going to skip cutting another release of 28.2.0, because we have the docker tag v28.2.0 already, and I don't want to cause confusion of which one should be consumed.  So 28.2.1 will be the "real" 28.2.0.

Closes #335

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+28.2.1
+------
+- Document how semver is used in the codebase.
+
 28.2.0
 ------
 - Changes how `parser.bad_lines_seen` and `backend.retried` metrics are sent.  They are now only sent on change.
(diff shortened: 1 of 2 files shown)


tiedotguy:
I like how the coverage went down even though only `.md` files were changed, because the exit condition on some of the clean shutdowns is racy :sigh:

PR #340, opened 2020-08-13
Fixes 4 known flaky coverage reports

### LINKED ISSUE #338: Variable test coverage in TestTagStripMerges* due to map iteration ordering
Due to maps having a random iteration order (by design), the `TestTagStripMerges*` tests can execute in two different orderings:
- key:value is processed, key:value,key2:value2 is processed
- key:value,key2:value2 is processed, key:value is processed

The outcome will always be the same (if the test passes) because there's order created by the timestamp, however the test coverage varies depending on which path is executed.

This makes coveralls sad, which makes me sad.

### LINKED ISSUE #339: Variable test coverage in BackendHandler.{DispatchMetricMap|Process} and Sender.cleanup
The shutdown path isn't explicitly tested, so sometimes it gets a chance to run, sometimes it doesn't.

tiedotguy:
I'm sure there's more.

Fixes #338 
Fixes #339

### DIFF
--- a/pkg/backends/sender/sender_test.go
+++ b/pkg/backends/sender/sender_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/ash2k/stager/wait"
 	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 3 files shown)


tiedotguy:
No changelog/release for this.

PR #341, opened 2020-08-13
Update CloudHandler.DispatchMetricMap to use MetricMap all the way through

tiedotguy:
This is the current "big" thing that I'm aware of blocking clustering, 'cause it interferes with loop detection.

As usual, commits are relatively stand-alone.

@mwhittington21 - if you could get a chance to test this at some point (I'll release it as 28.3.0-rc1), that would be much appreciated.  I generally soak changes locally, but I can only validate this one with tests.

### DIFF
--- a/CLOUDPROVIDERS.md
+++ b/CLOUDPROVIDERS.md
@@ -12,9 +12,6 @@ There are currently two supported cloud providers:
 
 All configuration is in a stanza named after the backend, and takes simple key value pairs.
 
-**Cloud providers should be disabled on the aggregation server when using http forwarding, as the source IP isn't
-propagated, and that information should be collected on the ingestion server.**
-
 aws
 ---
 ### TODO
--- a/counters.go
+++ b/counters.go
@@ -14,6 +14,11 @@ func NewCounter(timestamp Nanotime, value int64, source Source, tags Tags) Count
 	return Counter{Value: value, Timestamp: timestamp, Source: source, Tags: tags.Copy()}
 }
 
+func (c *Counter) AddTagsSetSource(additionalTags Tags, newSource Source) {
+	c.Tags = c.Tags.Concat(additionalTags)
+	c.Source = newSource
+}
+
 // Counters stores a map of counters by tags.
 type Counters map[string]map[string]Counter
 
--- a/metric_map.go
+++ b/metric_map.go
@@ -54,87 +54,93 @@ func MergeMaps(mms []*MetricMap) *MetricMap {
 }
 
 func (mm *MetricMap) Merge(mmFrom *MetricMap) {
-	mmFrom.Counters.Each(func(metricName string, tagsKey string, counterFrom Counter) {
-		v, ok := mm.Counters[metricName]
-		if ok {
-			counterInto, ok := v[tagsKey]
-			if ok {
-				if counterInto.Timestamp < counterFrom.Timestamp {
-					counterInto.Timestamp = counterFrom.Timestamp
-				}
-				counterInto.Value += counterFrom.Value
-			} else {
-				counterInto = counterFrom
-			}
-			v[tagsKey] = counterInto
-		} else {
-			mm.Counters[metricName] = map[string]Counter{
-				tagsKey: counterFrom,
-			}
-		}
-	})
+	mmFrom.Counters.Each(mm.MergeCounter)
+	mmFrom.Gauges.Each(mm.MergeGauge)
+	mmFrom.Sets.Each(mm.MergeSet)
+	mmFrom.Timers.Each(mm.MergeTimer)
+}
 
-	mmFrom.Gauges.Each(func(metricName string, tagsKey string, gaugeFrom Gauge) {
-		v, ok := mm.Gauges[metricName]
+func (mm *MetricMap) MergeCounter(metricName string, tagsKey string, counterFrom Counter) {
+	v, ok := mm.Counters[metricName]
+	if ok {
+		counterInto, ok := v[tagsKey]
 		if ok {
-			gaugeInto, ok := v[tagsKey]
-			if ok {
-				if gaugeInto.Timestamp < gaugeFrom.Timestamp {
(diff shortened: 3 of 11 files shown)


CLOUDPROVIDERS.md
```diff
@@ -12,9 +12,6 @@ There are currently two supported cloud providers:
 
 All configuration is in a stanza named after the backend, and takes simple key value pairs.
 
-**Cloud providers should be disabled on the aggregation server when using http forwarding, as the source IP isn't
```
tiedotguy, line 15: I believe technically this became safe with #330

metric_map.go
```diff
@@ -367,6 +373,7 @@ func (mm *MetricMap) String() string {
 }
 
 // AsMetrics will synthesize Metrics from the MetricMap and return them as a slice
+// TODO: Remove this function, it only supports tests now
```
tiedotguy, line 376: I'll raise an issue after this is merged.

pkg/statsd/handler_cloud.go
```diff
@@ -192,12 +203,11 @@ func (ch *CloudHandler) Run(ctx context.Context) {
 }
 
 func (ch *CloudHandler) handleInstanceInfo(ctx context.Context, info gostatsd.InstanceInfo) {
-	metrics := ch.awaitingMetrics[info.IP]
-	if len(metrics) > 0 {
+	mm := ch.awaitingMetrics[info.IP]
```
tiedotguy, line 206: Someone plz double check my logic here:
- Metric comes in, goes to `handleIncomingMetrics`
- statsMetricHostsQueued is incremented because the host isn't in `awaitingMetrics` or `awaitingEvents`
- Event comes in, goes to `handleIncomingEvent`
- statsEventHostsQueued is **not** incremented because the host is in `awaitingMetrics`
- Lookup completes
- `handleInstanceInfo` sees the IP in `awaitingMetrics`, decrements `statsMetricHostsQueued`
- `handleInstanceInfo` sees the IP in `awaitingEvents`, decrements `statsEventHostsQueued `, it's now `-1`

I think the situation occurred in both the old logic, and this new code?  I'm inclined to think we should just drop both and use a single `hostsQueued` (or even just a `len(toLookupIPs)`)?

pkg/statsd/handler_cloud.go
```diff
@@ -57,37 +56,50 @@ func (ch *CloudHandler) EstimatedTags() int {
...
 		select {
 		case <-ctx.Done():
-		case ch.incomingMetrics <- toHandle:
+		case ch.incomingMetrics <- mmToHandle:
 		}
 	}
 }
 
-// DispatchMetricMap re-dispatches a MetricMap through CloudHandler.processMetrics
-// TODO: This is inefficient, and should be handled first class, however that is a major re-factor of
```
tiedotguy, line 84: > that is a major re-factor

so that was a lie

mwhittington21:
@tiedotguy I'll try and find some time to upgrade the clusters. A bit swamped right now, but I'll get to it asap.

counters.go
```diff
@@ -14,6 +14,11 @@ func NewCounter(timestamp Nanotime, value int64, source Source, tags Tags) Count
 	return Counter{Value: value, Timestamp: timestamp, Source: source, Tags: tags.Copy()}
 }
 
+func (c *Counter) AddTagsSetSource(additionalTags Tags, newSource Source) {
```
MovieStoreGuy, line 17: Does this operation need to be threadsafe?

tiedotguy: Nope

pkg/statsd/handler_cloud.go
tiedotguy: This potentially problematic logic can be ignored, there's a refactor queued up which addresses it.

tiedotguy:
:bueller:

pkg/statsd/handler_cloud.go
```diff
@@ -57,37 +56,50 @@ func (ch *CloudHandler) EstimatedTags() int {
...
 
-func (ch *CloudHandler) processMetrics(ctx context.Context, metrics []*gostatsd.Metric) {
+func (ch *CloudHandler) DispatchMetricMap(ctx context.Context, mm *gostatsd.MetricMap) {
 	mmToDispatch := gostatsd.NewMetricMap()
-	var toHandle []*gostatsd.Metric
-	for _, m := range metrics {
-		if ch.updateTagsAndHostname(m, m.Source) {
-			mmToDispatch.Receive(m)
+	mmToHandle := gostatsd.NewMetricMap()
+	mm.Counters.Each(func(metricName string, tagsKey string, c gostatsd.Counter) {
```
MovieStoreGuy, line 62: Should the pointer of the values be passed in here?

pkg/statsd/handler_cloud.go
MovieStoreGuy: I look forward to the update.

MovieStoreGuy (approved):

metric_map.go
tiedotguy: Sure.  Does your IDE use `deprecated` or `Deprecated`? Goland seems to want the D or it won't highlight it.

pkg/statsd/handler_cloud.go
tiedotguy: I look forward to merging my third tweaking of this branch in to this once this branch is merged to master :okay:

pkg/statsd/handler_cloud.go
tiedotguy: You mean like:
```
	mm.Counters.Each(func(metricName string, tagsKey string, c *gostatsd.Counter) {
```
...?

If so, no - a MetricMap is all structs, not pointers to structs.

metric_map.go
tiedotguy: Goland has a "deprecate harder", for when the symbol is actually intended to be removed, but damned if I can find the comment syntax for that.

MovieStoreGuy (approved):

Issue #342, opened 2020-09-07
Address how aggregations deals with large numbers


MovieStoreGuy:
As part of https://github.com/atlassian/gostatsd/pull/333, it was identified that it is possible for aggregations to produce large numbers that can not be expressed as within json and for the time being causing dataloss for the datadog backend. 

I'd like to suggest that we look at clipping values to closest representative value within the aggregator:
- Inf+ --> float.Max
- Inf- --> float.Min
- NaN --> 0 

Or at least, allow this to be a configuration that would allow for the aggregator to drop or clip since.

On the same vein, users shouldn't be sending values that would result in one of these however, it has been seen and with larger aggregation periods, I suspect we may actually see this happen more often.

Curious to know your thoughts @tiedotguy

tiedotguy:
Is this done now with the change from @digitalpoetry , @MovieStoreGuy ?

MovieStoreGuy:
Yeah, I think so :)

PR #343, opened 2020-09-09
Coerce non-numeric values to numeric in the datadog backend

digitalpoetry:
In order to address the case where data is dropped from the datadog backend if the aggregation result is a non-numeric value (NaN, +Inf, -Inf), this PR modifies the component's behaviour to coerce such values into a numeric. This serves as a basic fix to #342. 

* Inf+ --> float.Max
* Inf- --> float.Min
* NaN --> -1

NaN's conversion to -1 was chosen because negative values are (anecdotally) less frequent compared to zero counters.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+28.3.0
+------
+- Rather than dropping data points, the Datadog backend will coerce non-numeric values resulting from aggregation to numeric.
+`NaN` is converted to -1, `+Inf` to float64 maximum, and `-Inf` to float64 minimum.
+
 28.2.1
 ------
 - Document how semver is used in the codebase.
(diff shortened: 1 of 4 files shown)


tiedotguy:
lgtm

tiedotguy (approved):

MovieStoreGuy:
Should we not do this for all agregations produced instead of just for the datadog backend?

MovieStoreGuy (approved):

tiedotguy:
We can pull it out if needed.  I think InfluxDB allows infinity for instance (the query language can check for it, but that might only be usable if you **produce** an infinity, not **query** an infinity).

digitalpoetry:
Thanks team!

PR #344, opened 2020-09-30
Improve test names

digitalpoetry:
Fixing a mistake in the testcase name I made in a previous PR


tiedotguy (approved):
lgtm conceptually, although looking at the larger picture, they should all have had different names from the start :pokerface:

digitalpoetry:
Thanks for the feedback @tiedotguy. I've clarified the intent of each test in the latest commit.

tiedotguy (approved):

PR #353, opened 2020-10-28
Refactor lexer

tiedotguy:


internal/lexer/lexer_test.go
```diff
@@ -164,46 +164,50 @@ func compareMetric(t *testing.T, tests map[string]gostatsd.Metric, namespace str
...
+	}
+
 	b.ReportAllocs()
 	b.ResetTimer()
 	for n := 0; n < b.N; n++ {
-		r, _, _ = dp.parseLine(slice)
+		r, _, _ = l.Run(slice, ns)
 		r.Done()
 	}
 	parselineBlackhole = r
```
tiedotguy, line 181: One day I'll go back and figure out why I needed this line and comment it :coneofshame:

Something something optimizer.

tiedotguy: Oh wow, looks like maybe I didn't add it, it's some real dawn of time stuff.  I'm pretty sure that since `r.Done()` was added, it's no longer necessary.

internal/lexer/lexer.go
```diff
@@ -63,7 +64,24 @@ func (l *lexer) next() byte {
 	return b
 }
 
-func (l *lexer) run(input []byte, namespace string) (*gostatsd.Metric, *gostatsd.Event, error) {
+func (l *Lexer) reset() {
```
MovieStoreGuy, line 67: Does this need to be thread safe?

tiedotguy: Explicitly not - hence why every DatagramParser.Run creates its own.

MovieStoreGuy (approved):

Issue #284, opened 2020-02-11
Implement Prometheus statser


tiedotguy:
This would allow for internal metrics to be monitored from prom.

This could be useful when there's a large number of ephemeral hosts, and other backends may not scale as well with the high cardinality of internal metrics across many hosts.

This is mostly stand-alone and doesn't interact with the more complicated aspects of how gostatsd handles things, as it's mostly just adding an http server + prom, so it's probably a good issue for someone looking to get across the code base.

This is explicitly not a prom backend, this is only for internal metrics.

0AlexZhong0:
> This would allow for internal metrics to be monitored from prom.
> 
> This could be useful when there's a large number of ephemeral hosts, and other backends may not scale as well with the high cardinality of internal metrics across many hosts.
> 
> This is mostly stand-alone and doesn't interact with the more complicated aspects of how gostatsd handles things, as it's mostly just adding an http server + prom, so it's probably a good issue for someone looking to get across the code base.
> 
> This is explicitly not a prom backend, this is only for internal metrics, see #285 for that.

Hey, I am keen to give this issue a go. All good if I take it from here?

tiedotguy:
That'd be great!

0AlexZhong0:
Would `pkg/web` be the right place to make the changes at,, or is there a more suitable directory where I can add the HTTP server with Prometheus to?

tiedotguy:
I'm not familiar with the Prom SDK in go, but my approach would be to create an appropriate PromStatser (which also includes the necessary Handler) in `createStatser`, then propagate that through to `NewHttpServer`, then if it's enabled (for that particular http server), attempt a type assertion to convert the interface to a PromStatser, and then attach the Handler if it's the right type.

If it's the wrong type, return an error - fast fail on startup is encouraged.

0AlexZhong0:
What do you mean by propagating the PromStatser to `NewHttpServer`? I am not able to find the related code.

And when you said attaching the Handler at the end, that is the Handler for the `NewHttpServer`, right? And specifically, what would this handler do?

tiedotguy:
You don't need to worry about Metric/EventHandler, that would be for making a prom backend (#285), this is purely for internal metrics.

[...]

PR #355, opened 2020-10-31
WIP: Implement Prometheus statser

0AlexZhong0:
Currently working on #284.


pkg/web/prometheus.go
```diff
@@ -0,0 +1,66 @@
...
+	"net/http"
+
+	"log"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+// TODO: now I have setup the Prometheus metric objects, how do I integrate with gostatsd internal metrics measuring
+// tools? How do I run this file, and start writing some tests for this. Need to adhere to the project testing styles.
```
tiedotguy, line 14: The expected interface is a `stats.Statser` (https://github.com/atlassian/gostatsd/blob/master/pkg/stats/statser.go), and you can embed a `stats.flushNotifier` to take care of the `RegisterFlush` and `NotifyFlush` calls.

You'll likely need to lazily create metrics on-demand and store them in a map - gostatsd doesn't have any arbitrary metrics, so you don't need to worry about removing things from the map.  The map can be protected with an RWMutex, since after warmup, it will be very rare (if ever), that additional metrics are created.

tiedotguy: Also, there's basically no testing of the actual Statsers.  There is a bit for `pkg/web`, but it's acceptable to just make sure you get a 200 return code when it's expected.

pkg/stats/statser_prometheus.go
```diff
@@ -0,0 +1,78 @@
...
+// NewPrometheusStatser creates a new Statser which
+// sends internal metrics to prometheus
+func NewPrometheusStatser(gaugeVec prometheus.GaugeVec, counterVec prometheus.CounterVec) *PrometheusStatser {
+	return &PrometheusStatser{
+		gaugeVec:   gaugeVec,
+		counterVec: counterVec,
+	}
+}
+
+func (ps *PrometheusStatser) NotifyFlush(ctx context.Context, d time.Duration) {
```
tiedotguy, line 39: You don't need to implement `NotifyFlush` and `RegisterFlush`, it's picked up implicitly from the embedded `flushNotifier`

pkg/web/httpservers.go
```diff
@@ -88,6 +93,10 @@ func NewHttpServer(
 		address: address,
 	}
 
+	if _, enablePromStatser := statser.(*stats.PrometheusStatser); enablePromStatser {
```
tiedotguy, line 96: There should also be an explicit `enablePromStatser` argument, otherwise the prom statser will be exposed to all http servers, which may not be what is desired.

```
if enablePromStatser {
  if _, ok := statser.(*stats.PrometheusStatser); ok {
    // TODO: append the route
  } else {
    return errNotPromStatser
  }
}
```

pkg/stats/statser_prometheus.go
```diff
@@ -0,0 +1,78 @@
...
+// this depends on the metrics being collected, if they are the same thing
+// but can be partitioned into different types/groups, e.g. http requests partitioned
+// by user age and demographics, then using a collector makes sense. (is this correct?)
+
+// PrometheusStatser is a Statser that monitors gostasd's internal metrics from
+// Prometheus, it is useful when there is a large number of ephemeral hosts.
+type PrometheusStatser struct {
+	flushNotifier
+
+	// collector of gauges that stores the internal gauge metrics of gostatsd
```
tiedotguy, line 24: Each `GaugeVec` and `CounterVec` is a single metric name, they need to be lazily created.

pkg/stats/statser_prometheus.go
```diff
@@ -0,0 +1,78 @@
...
+	ps.flushNotifier.NotifyFlush(ctx, d)
+}
+
+func (ps *PrometheusStatser) RegisterFlush() (<-chan time.Duration, func()) {
+	return ps.flushNotifier.RegisterFlush()
+}
+
+// TODO: how do I use tags here, same for the Count and the TimingMS methods
+func (ps *PrometheusStatser) Gauge(name string, value float64, tags gostatsd.Tags) {
+	ps.gaugeVec.WithLabelValues(name).Add(value)
```
tiedotguy, line 49: The tags map to labels

tiedotguy:
Just as an aside, I believe prom uses `snake_case` to name its metric, not `dotted.case` - you should have a place where there's a basic `strings.Replace` to change . to _

pkg/stats/statser_prometheus.go
0AlexZhong0: How come the `TaggedStatser` implements those two functions or why does `TaggedStatser` not use a `flushNotifier` as one of its attributes?

squizzling: The `TaggedStatser` is a wrapper that explicitly calls to the wrapped `Statser`.

It shouldn't use a `flushNotifier` directly, because then the `TaggedStatser` and the underlying `Statser` would have different notification targets (and ultimately `MetricFlusher.Run` would only call `NotifyFlush` on the underlying `Statser`)

That being said, it could and probably should use an embedded `Statser` and remove the explicit implementation.

tiedotguy: My bad, that comments from my personal account.

The `TaggedStatser` is a wrapper that explicitly calls to the wrapped `Statser`.

It shouldn't use a `flushNotifier` directly, because then the `TaggedStatser` and the underlying `Statser` would have different notification targets (and ultimately `MetricFlusher.Run` would only call `NotifyFlush` on the underlying `Statser`)

That being said, it could and probably should use an embedded `Statser` and remove the explicit implementation.

0AlexZhong0: Oh, I see that now, what was the design process behind, is it because that they have different notification targets, so one such as `TaggedStatser` is a wrapper of  `Statser` but `InternalStatser` is not a wrapper because they have the same notification targets?

I am using the GoLand IDE, and I am trying to find the usage of `NotifyFlush`, for some reason I cannot see any call to this method outside of the `statsers` and the `flush_notifer`.

tiedotguy: There was no special design considerations behind `TaggedStatser` not embedding the `Statser`, it's possible at the time that I thought you could only embed a struct in a struct, and an interface in an interface, but not an interface in a struct, I don't recall for sure though.

`InternalStatser` does explicitly implement `NotifyFlush` however, as it needs to take action on the flush **before** returning.  Flush notification is an asynchronous, best effort action, and any metrics triggered by it are processed in the next flush cycle.

The `InternalStatser` however explicitly needs to process the metrics in the current flush cycle, so it takes action (dispatching internal metrics) before returning.

I'm not sure why GoLand doesn't detect usage - it does in my setup.  You could try rebuilding the index through File -> Invalidatae Caches / Restart, but I don't know if it will help.

The only place it's called is at https://github.com/atlassian/gostatsd/blob/master/pkg/statsd/flusher.go#L69

0AlexZhong0: My problem was I was not aware that the `Statser` interface also has a `NotifyFlush` method, I was only trying to find the usage of `NotifyFlush` from the `flsuhNotifier` struct.

pkg/stats/statser_prometheus.go
0AlexZhong0: Does it make sense to use the `Gauges` and `Counter` structs `gostatsd` already provides or would using the structs provided from the `prometheus` package be better?

pkg/stats/statser_prometheus.go
tiedotguy: Ahhh, that would do it :)

pkg/stats/statser_prometheus.go
tiedotguy: You'd need to use the ones from the prometheus packages, because they're ultimately added to the prometheus registry.

PR #358, opened 2020-11-20
Build and publish Docker images for arm64

hligit:
A test build can be found [here](https://travis-ci.org/github/hligit/gostatsd/jobs/744840123), with temporary changes for pushing to my own docker hub repository.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,9 @@ go:
   - "1.13.6"  # Keep in sync with Makefile and README.md
 
 arch:
- - amd64
- - arm64
+  - amd64
+  - arm64
+
 
 dist: bionic
 os: linux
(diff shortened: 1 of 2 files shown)


tiedotguy:
lgtm.  Do the builds run in parallel?

tiedotguy (approved):

hligit:
Yes, they should run in parallel.

MovieStoreGuy (dismissed):

hligit:
The manifest creation and pushing need to be on the same host to avoid overriding. I moved manifest job into a separate stage. Tested with a rc tag. You can see the multi-arch image pushed in [Docker hub](https://hub.docker.com/layers/atlassianlabs/gostatsd/28.3.1-rc8/images/sha256-171a87859827fa6df680ae1fad42bfab456671db212920dd0d1f7f91812e6625?context=explore). Please review again.

.travis.yml
```diff
@@ -8,8 +6,33 @@ services:
...
+   - os: linux
+     arch: amd64
+   - os: linux
+     arch: arm64
+   - stage: Push Docker manifest
+     env:
+       - DOCKER_CLI_EXPERIMENTAL=enabled
+     script:
+       - curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
+       - sudo add-apt-repository "deb [arch=$TRAVIS_CPU_ARCH] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
```
MovieStoreGuy, line 28: What is the reason we need to add in our docker client?

hligit: Default Docker client version is very old, 18.06, there has been many bug fixes around buildkit since. I also changed Ubuntu version by `dist: bionic`, to get the newer Docker version.

I didn't hit particular issue for this change, but I did attempt to use `buildx`, which is not supported in v18.06. In general, I think it's better to use a newer version of Docker.

tiedotguy (approved):
I'm far from an expert in the build system or docker, but I see nothing that jumps out as a problem.

tiedotguy:
Don't do a release after merging this - there's been a couple of minor unreleased changes, so I'll submit a PR with an updated change log to roll them all up after the merge.

Issue #280, opened 2020-01-04
Remove / refactor textContext


tiedotguy:
https://github.com/atlassian/gostatsd/pull/279/files#r363030348

tl;dr - I've been using a test context with a 1100ms timeout to timeout tests, but the fail occurs on another goroutine, which is a disallowed / racey.

PR #383, opened 2021-08-24
Enable builds on all branches

Excaleo:
I keep forgetting release builds have their own separate branch, previous attempt at limiting builds ended up disabling release builds. 

I have turned off the build on pull request option, since building on every commit to a pushed branch would replace it. This prevents the build from building twice per PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,10 +13,6 @@ arch:
 dist: bionic
 os: linux
 
-branches:
-  only:
-    - master
-
 env:
   global:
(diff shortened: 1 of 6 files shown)


chenzhihao (approved):

tiedotguy:
Closes #280

Excaleo:
@tiedotguy I've also increased the http forwarder test timeout duration to 1.5 seconds as the ARM build sometimes times out in Travis CI, is this an acceptable test degredation for the AMD64 build?

1.5 seconds was derived from the previous failing travis build which indicated a duration of 1.3s

tiedotguy:
If we've got tests timing out, that's an indication of a problem somewhere - the purpose of the timeout is to say "hey, this test is slow, if you need to perform a sleep in order to avoid 'raceyness', then the test should be refactored"

It's fine to bump it if it's necessary, but we should give a failing test a sniff test to see if it's got bad behavior, and raise an issue to address it if necessary.

tiedotguy (approved):

tiedotguy:
You can also merge without coveralls passing, but give it a quick look and see if there's an actual flakeyness problem, and raise an issue if so - in this case, the report doesn't say anything went down, so I think it's fine. Not certain why it's shown as a decrease in GH.

Excaleo:
> If we've got tests timing out, that's an indication of a problem somewhere - the purpose of the timeout is to say "hey, this test is slow, if you need to perform a sleep in order to avoid 'raceyness', then the test should be refactored"

From this build result, it seems this is a degradation within the ARM compiled binary. https://app.travis-ci.com/github/atlassian/gostatsd/jobs/533596613#L843

`BenchmarkReceive-80                                   	       1	1228779055 ns/op	262749480 B/op	   12341 allocs/op`

The same benchmark in the AMD build takes a much lower time.

`BenchmarkReceive-2                                  	  105783	     11717 ns/op	   18962 B/op	       4 allocs/op`

Might be worth raising this as a separate issue, I think the test case itself is fine.

tiedotguy:
That's a lot of nanoseconds.  Please raise an issue, it should be looked in to.

Excaleo:
Issue #384 has been raised, going ahead with this merge to patch the security vulnerability.

PR #388, opened 2021-08-30
Adds Prometheus Remote Write

tiedotguy:
Probably easiest to just look at 6ca4ea3, it's based off the Datadog backend, with a lot of reused stuff.


pkg/backends/promremotewriter/promremotewriter.go
```diff
@@ -0,0 +1,381 @@
...
+	"github.com/atlassian/gostatsd/internal/util"
+	"github.com/atlassian/gostatsd/pb"
+	"github.com/atlassian/gostatsd/pkg/stats"
+	"github.com/atlassian/gostatsd/pkg/transport"
+)
+
+const (
+	// BackendName is the name of this backend.
+	BackendName                  = "promremotewriter"
+	defaultUserAgent             = "gostatsd" // TODO: Add version
```
MovieStoreGuy, line 33: Worth raising an issue for?

pkg/backends/promremotewriter/promremotewriter.go
```diff
@@ -0,0 +1,381 @@
...
+)
+
+var (
+	// defaultMaxRequests is the number of parallel outgoing requests to prometheus.  As this mixes
+	// both CPU (TLS) and network bound operations, balancing may require some experimentation.
+	defaultMaxRequests = uint(10 * runtime.NumCPU())
+)
+
+// Client represents a Prometheus Remote Writer client.
+type Client struct {
```
MovieStoreGuy, line 48: Is there an interface that this implements?

pkg/backends/promremotewriter/promremotewriter.go
```diff
@@ -0,0 +1,381 @@
...
+
+// SendMetricsAsync flushes the metrics to Prometheus, preparing payload synchronously but doing the send asynchronously.
+func (prw *Client) SendMetricsAsync(ctx context.Context, metrics *gostatsd.MetricMap, cb gostatsd.SendCallback) {
+	counter := 0
+	results := make(chan error)
+
+	now := clock.FromContext(ctx).Now().UnixNano() / 1_000_000
+	prw.processMetrics(now, metrics, func(writeReq *pb.PromWriteRequest) {
+		atomic.AddUint64(&prw.batchesCreated, 1)
+		go func() {
```
MovieStoreGuy, line 74: Any concern if this is called mutliple times?

pkg/backends/promremotewriter/promremotewriter.go
```diff
@@ -0,0 +1,381 @@
...
+	raw, err := proto.Marshal(writeReq)
+	if err != nil {
+		return nil, err
+	}
+
+	bufBorrowed := (*buffer).Bytes()
+	bufCompressed := snappy.Encode(bufBorrowed, raw)
+	if cap(bufBorrowed) < cap(bufCompressed) {
+		// it overflowed the buffer and had to allocate, so we'll keep the larger buffer for next time
+		*buffer = bytes.NewBuffer(bufCompressed)
```
MovieStoreGuy, line 271: is it worth looking at https://pkg.go.dev/sync#Pool for this functionality instead of doing the memory management for go?

MovieStoreGuy (commented):
Just some questions on things, I would like to enforce compile time interface checks where possible.

pb/remote_write.pb.go
```diff
@@ -0,0 +1,229 @@
+// Code generated by protoc-gen-go. DO NOT EDIT.
```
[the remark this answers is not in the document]
tiedotguy: Yep, we do it already with the other PB stuff.  It simplifies the need to deal with protoc.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
+29.1.0
+------
+- Initial Prometheus Remote Write support added
+
 29.0.2
 ------
+- New Relic backend handles infinities and NaN better
```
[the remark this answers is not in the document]
tiedotguy: No, it was an "unreleased" change in 29.0.2, this just documents when the behavior went in.

pkg/backends/promremotewriter/flush.go
```diff
@@ -0,0 +1,159 @@
...
+	"sort"
+	"strings"
+	"unicode"
+
+	"github.com/atlassian/gostatsd"
+	"github.com/atlassian/gostatsd/pb"
+)
+
+// flush represents a send operation.
+type flush struct {
```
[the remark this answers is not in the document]
tiedotguy: Nope

pkg/backends/promremotewriter/flush.go
```diff
@@ -0,0 +1,159 @@
...
+		labels = append(labels, &pb.PromLabel{
+			Name:  key,
+			Value: strings.Join(newTags[key], "__"),
+		})
+	}
+	return labels
+}
+
+func sanitizeLabelName(labelName string) string {
+	// Label names MUST adhere to the regex `[a-zA-Z_]([a-zA-Z0-9_])*`
```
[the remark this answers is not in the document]
tiedotguy: I generally don't favor regex, and don't use it for trivial things.  We do validate against the regex in tests though.

pkg/backends/promremotewriter/flush.go
```diff
@@ -0,0 +1,159 @@
...
+	} else if math.IsInf(v, 1) {
+		return math.MaxFloat64
+	} else if math.IsInf(v, -1) {
+		return -math.MaxFloat64
+	}
+	return v
+}
+
+func (f *flush) maybeFlush() {
+	if uint(len(f.writeRequest.Timeseries))+20 >= f.metricsPerBatch { // flush before it reaches max size and grows the slice
```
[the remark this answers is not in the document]
tiedotguy: It's inherited from datadog, but it's meant to be the maximum cardinality of a single metric.  That is, a timer is going to have like 10 metrics it emits.  20 just gives it a buffer.

The intent is `len(...) < metricsPerBatch`, rather than `len(...) > metricsPerBatch+epsilon`, because if there's a limit on the backend, then it might be a hard limit, so going 1 over is a hard fail.

pkg/backends/promremotewriter/promremotewriter.go
tiedotguy: Yup, `gostatsd.SendMetricsAsync`.  It's not formally checked, but the code won't compile if it doesn't implement it (`NewClientFromViper` will fail its return type)

pkg/backends/promremotewriter/promremotewriter.go
tiedotguy: `snappy` is creating/growing the byte array if needed, the best we can do is detect if it needed to grow, and re-use it.  The `bytes.Buffer` is used in the semaphore, so we'll reach a local maxima and then stop allocating more data.  There's not a large amount of byte slices allocated.

pb/remote_write.proto
```diff
@@ -0,0 +1,29 @@
+syntax = "proto3";
```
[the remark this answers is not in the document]
tiedotguy: Yeah, came from the prom remote write spec, I can add a ref to it.

- [ ] this

CHANGELOG.md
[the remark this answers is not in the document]
MovieStoreGuy: No wokkas

pb/remote_write.proto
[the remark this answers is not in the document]
MovieStoreGuy: Yes please, just in case we need to chase it up.

pkg/backends/promremotewriter/flush.go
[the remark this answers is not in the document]
MovieStoreGuy: Could I ask you to make the number a constant with the file to make it easier in future to understand the code a bit more?

pkg/backends/promremotewriter/promremotewriter.go
MovieStoreGuy: Not worth doing the compiler time interface check within the file?

pkg/backends/promremotewriter/promremotewriter.go
MovieStoreGuy: Sounds good.

MovieStoreGuy (approved):

pkg/backends/promremotewriter/promremotewriter.go
tiedotguy: Nah, pretty sure it's a theme.

pkg/backends/promremotewriter/promremotewriter.go
tiedotguy: We do, `NewClientFromViper` will fail if it doesn't match.

tiedotguy:
Once again we need to ignore coveralls, I think it's just sad because there's new code without sufficient coverage to actively increase the %.

PR #389, opened 2021-09-08
update alpine 3.13.1 -> 3.13.2

jamesmoessis:
Update alpine version, mainly to solve openssl vulns.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+29.0.3
+------
+- Update to Alpine 3.14.2
+
 29.0.2
 ------
 - Update to Alpine 3.14.1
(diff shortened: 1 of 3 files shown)


MovieStoreGuy:
I wonder if it is worthwhile moving away from alpine and using a distroless / scratch container for GoStatsd? 

There is nothing within alpine itself that it really needs.

Can I also ask that you raise an issue for the docker image as well?

jamesmoessis:
I've raised the issue: https://github.com/atlassian/gostatsd/issues/390

MovieStoreGuy:
I am going to reject this PR in favour of the issue created since it will reduce the number of times it is needed to bump alpine to the latest fixed version.

Issue #390, opened 2021-09-08
Alpine libraries causing unnecessary bloat


jamesmoessis:
As mentioned over in #389, the gostatsd image doesn't really need to be based on Alpine. It could be done with distroless or scratch image. This has the benefit of less dependencies and less maintenance, particularly keeping security vulnerabilities up to date.

PR #391, opened 2021-09-08
change to distroless image 

jamesmoessis:
as a result of https://github.com/atlassian/gostatsd/issues/390

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,3 @@
-29.0.3
-------
-- Update to Alpine 3.14.2
-
 29.0.2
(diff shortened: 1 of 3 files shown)


jamesmoessis:
@MovieStoreGuy I've run `make test-all-full` locally and it passes. Anything else we need to verify to merge this?

MovieStoreGuy (approved):

MovieStoreGuy:
@jamesmoessis , just make sure the resulting docker container starts without issue. I imagine there is nothing wrong but better to be safe than sorry.

PR #392, opened 2021-09-09
change back to alpine base image

jamesmoessis:
we are sorry we ever left, take us back alpine


MovieStoreGuy (approved):

tiedotguy:
Feels like there's a story here.

jamesmoessis:
@tiedotguy I realised that you can't run scripts in distroless images because they don't have a shell. Something that many dependencies would likely want to do, including our own.

Issue #393, opened 2021-10-14
Open, fixable vulnerabilities in imported packages


flo-everett-xero:
Running [trivy](https://github.com/aquasecurity/trivy) on the gostatsd package after building from source reveals three fixable security vulnerabilities. trivy output:
```
usr/bin/gostatsd (gobinary)
===========================
Total: 3 (HIGH: 3, CRITICAL: 0)

+--------------------------+------------------+----------+--------------------------------------+------------------------------------+
|         LIBRARY          | VULNERABILITY ID | SEVERITY |          INSTALLED VERSION           |           FIXED VERSION            |
+--------------------------+------------------+----------+--------------------------------------+------------------------------------+
| github.com/gogo/protobuf | CVE-2021-3121    | HIGH     | v1.2.2-0.20190723190241-65acae22fc9d | v1.3.2                             |
+--------------------------+------------------+          +--------------------------------------+------------------------------------+
| golang.org/x/crypto      | CVE-2020-29652   |          | v0.0.0-20191011191535-87dc89f01550   | v0.0.0-20201216223049-8b5274cf687f |
+                          +------------------+          +                                      +------------------------------------+
|                          | CVE-2020-9283    |          |                                      | v0.0.0-20200220183623-bac4c82f6975 |
+--------------------------+------------------+----------+--------------------------------------+------------------------------------+
```

PR #394, opened 2021-10-15
Moving gostatsd to use go1.17 and updating deps

MovieStoreGuy:
Looking to resolve https://github.com/atlassian/gostatsd/issues/393

I am going to create a follow up PR that adds dependabot and tests on push using github actions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ server based on load.
 
 Building the server
 -------------------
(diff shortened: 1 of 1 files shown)


MovieStoreGuy:
Looks bumping this means that arm is busted?

tiedotguy:
Re-running can be dicey I've found. Something about "but I've already got a result (failure), so I'm going to fail this run as well".

Issue #320, opened 2020-06-29
Upgrade protobuf


tiedotguy:
github.com/golang/protobuf is superseded by google.golang.org/protobuf

MovieStoreGuy:
Not sure on the impact if you upgrade the either the client or server.

This will also require rebuilding the protobuf bindings with protoc

tiedotguy:
The protocols the same, just a different code interface right?

MovieStoreGuy:
I believe so, haven't looked into enough to fully understand the implications.

PR #397, opened 2021-10-19
Updating protobuf library used for gostatsd

MovieStoreGuy:
There is an open issue for this that I will link in a moment.


MovieStoreGuy:
This is the linked issue https://github.com/atlassian/gostatsd/issues/320

MovieStoreGuy:
Is there a version you have in mind that it needs to support @tiedotguy ?

PR #422, opened 2021-11-20
Ensure forwarder flushes all queued data

MovieStoreGuy:
Does what it says on the tin :) 

There is a breaking change that it now enforces that a forwarder is single use since there is no means of unclosing a channel.

### DIFF
--- a/pkg/statsd/handler_http_forwarder_v2.go
+++ b/pkg/statsd/handler_http_forwarder_v2.go
@@ -233,12 +233,10 @@ func (hfh *HttpForwarderHandlerV2) Run(ctx context.Context) {
 			mergedMetricMap := mergeMaps(metricMaps)
 			mms := mergedMetricMap.SplitByTags(hfh.dynHeaderNames)
 			for dynHeaderTags, mm := range mms {
-				if !hfh.acquireSem(ctx) {
-					return
-				}
+				hfh.acquireSem()
 				postId := atomic.AddUint64(&hfh.postId, 1) - 1
 				go func(postId uint64, metricMap *gostatsd.MetricMap, dynHeaderTags string) {
-					hfh.postMetrics(ctx, metricMap, dynHeaderTags, postId)
+					hfh.postMetrics(context.Background(), metricMap, dynHeaderTags, postId)
 					hfh.releaseSem()
 				}(postId, mm, dynHeaderTags)
 			}
@@ -265,13 +263,8 @@ func mergeMaps(maps []*gostatsd.MetricMap) *gostatsd.MetricMap {
 	return mm
 }
 
-func (hfh *HttpForwarderHandlerV2) acquireSem(ctx context.Context) bool {
-	select {
-	case <-ctx.Done():
-		return false
-	case <-hfh.metricsSem:
-		return true
-	}
+func (hfh *HttpForwarderHandlerV2) acquireSem() {
+	<-hfh.metricsSem // will potentially block
 }
 
 func (hfh *HttpForwarderHandlerV2) releaseSem() {
--- a/pkg/statsd/handler_http_forwarder_v2.go
+++ b/pkg/statsd/handler_http_forwarder_v2.go
@@ -50,12 +50,14 @@ type HttpForwarderHandlerV2 struct {
 	maxRequestElapsedTime time.Duration
 	metricsSem            chan struct{}
 	client                *http.Client
-	consolidator          *gostatsd.MetricConsolidator
-	consolidatedMetrics   <-chan []*gostatsd.MetricMap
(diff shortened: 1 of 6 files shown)


pkg/statsd/handler_http_forwarder_v2.go
```diff
@@ -182,7 +185,14 @@ func (hfh *HttpForwarderHandlerV2) EstimatedTags() int {
 
 // DispatchMetricMap dispatches a metric map to the MetricConsolidator
 func (hfh *HttpForwarderHandlerV2) DispatchMetricMap(ctx context.Context, mm *gostatsd.MetricMap) {
-	hfh.consolidator.ReceiveMetricMap(mm)
```
tiedotguy, line 185: :interesting: I wasn't checking `ctx.Done()`

pkg/statsd/handler_http_forwarder_v2.go
```diff
@@ -217,14 +227,9 @@ func (hfh *HttpForwarderHandlerV2) emitMetrics(statser stats.Statser) {
...
-	for {
-		select {
-		case <-ctx.Done():
-			return
-		case metricMaps := <-hfh.consolidatedMetrics:
+	wg.Start(func() {
+		for metricMaps := range hfh.consolidatedMetrics {
 			mergedMetricMap := mergeMaps(metricMaps)
 			mms := mergedMetricMap.SplitByTags(hfh.dynHeaderNames)
 			for dynHeaderTags, mm := range mms {
```
tiedotguy, line 235: If `ctx` is cancelled, `acquireSem` will fail to get the sem, which means anything pending in the `mm`, `mms`, and `hfh.consolidatedMetrics` won't be processed.

Same with the `hfh.postMetrics` context.

I think it could be handled by removing `ctx` from `acquireSem` and `postMetrics`, then after the `range` has completed, acquiring the semaphore until you own everything (you can probably do a `for len(sem) > 0 { acquire }`.  That will ensure that:
- the `range` has finished, so nothing new will go on
- everything that had a semaphore has released it, so there's nothing pending.

tiedotguy: - [ ] don't depend on canceled ctx

tiedotguy (changes requested):

tiedotguy:
I'm not sure if it's possible, but my approach would be:
- make a channel
- stuff it full of known data
- close the channel
- start the runner
- wait for completion
- valid things are processed

The problem is you're not testing and validating the channel contents got processed with known and exact data, you're testing the channel is empty.

- [ ] this? maybe?

MovieStoreGuy:
Hey @tiedotguy , let me know what you think of the changes.

pkg/statsd/handler_http_forwarder_v2.go
```diff
@@ -182,7 +185,14 @@ func (hfh *HttpForwarderHandlerV2) EstimatedTags() int {
 
 // DispatchMetricMap dispatches a metric map to the MetricConsolidator
 func (hfh *HttpForwarderHandlerV2) DispatchMetricMap(ctx context.Context, mm *gostatsd.MetricMap) {
-	hfh.consolidator.ReceiveMetricMap(mm)
+	select {
+	case <-ctx.Done():
```
MovieStoreGuy, line 189: There is a point that both context and done are ready so it can randomly return either, does that matter here?

pkg/statsd/handler_http_forwarder_v2.go
```diff
@@ -182,7 +185,14 @@ func (hfh *HttpForwarderHandlerV2) EstimatedTags() int {
...
+		hfh.logger.WithError(ctx.Err()).Debug("Context is cancelled")
+	case <-hfh.done:
+		hfh.logger.Debug("Forwarder has shutdown and unable to accept metrics")
+	default:
+		hfh.consolidator.ReceiveMetricMap(mm)
```
MovieStoreGuy, line 194: Is it also worth guarding against nil MetricMap?

pkg/statsd/handler_http_forwarder_v2.go
tiedotguy: Nah, the only variability is in the logs, there's no meaningful observable difference - also they're both debug lines.

pkg/statsd/handler_http_forwarder_v2.go
tiedotguy: Nah. Would be nice if there was like.. non-nil annotations or something, so we can say "yes it's a pointer, but we know it's never nil".  But this is the world we live in.

tiedotguy (approved):

PR #439, opened 2022-01-09
Release Candidate 34.0.0

MovieStoreGuy:
## Breaking Change

There is a behavioural change to how the flush system works, it will now hold up any `context.Done` events until it has finished all of the required work. This should not impact statistically aggregation negatively but rather enough that there is data for those time intervals that would have been missed during a system's shutdown event.

## Minor Change

There is some additional changes for the dependancies and fixes to some CVEs found in the docker image.

# Change List

Please see https://github.com/atlassian/gostatsd/compare/33.0.2...master for the list of changes being made between versions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
-unreleased
+34.0.0
 ------
-- Forces http metrics forwarder to flush all data queued on shutdown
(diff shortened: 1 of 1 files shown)


jamesmoessis (approved):
LGTM

tiedotguy (approved):

PR #454, opened 2022-02-01
New Relic backend: Added support for throttling ("Retry-After" header)

malderete:
## Description
Under heavy loads using New Relic backend I've observed GoStatsD is not handling throttling correctly. In cases when New Relic reply with `StatusCode = 429`, we should read a header, sleep (potentially) and then retry.

### New Relic documentation
**API returns a 429 response for the remainder of the minute. The response will include a Retry-After header indicating how long to wait in seconds before resubmitting or sending new data.**

## Rationale
With this PR the code will try to use "Retry-After" (if available) header to sleep instead of backoff algorithm. In case of invalid or missing header it will fallback to backoff algorithm.
This change is intended to be backward-compatible.

# Changes
- [New Relic backend] added support to handle rate-limiting from the API by handling `Retry-After` response's header.
- [New Relic backend] added new Error for rate-limit case.
- [New Relic backend] added unit-test for new feature.

### DIFF
--- a/pkg/backends/newrelic/newrelic.go
+++ b/pkg/backends/newrelic/newrelic.go
@@ -5,6 +5,7 @@ import (
 	"compress/gzip"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -129,7 +130,6 @@ type NRMetric struct {
 
 // SendMetricsAsync flushes the metrics to New Relic, preparing payload synchronously but doing the send asynchronously.
 func (n *Client) SendMetricsAsync(ctx context.Context, metrics *gostatsd.MetricMap, cb gostatsd.SendCallback) {
-
 	counter := 0
 	results := make(chan error)
 	now := float64(clock.FromContext(ctx).Now().Unix())
@@ -295,6 +295,26 @@ func (n *Client) Name() string {
 	return BackendName
 }
 
+// RetryAfterError indicates if we should wait before retrying
+type RetryAfterError struct {
+	Response *http.Response
+}
+
+// Error returns the error message
+func (r RetryAfterError) Error() string {
+	return fmt.Sprintf("Too many request retry-after=%+v", r.RetryAfter())
+}
+
+// RetryAfter returns a duration to wait before based on response header Retry-After
+// If the header is not present or invalid returns 0
+func (r RetryAfterError) RetryAfter() time.Duration {
+	retryAfterSecs, err := strconv.Atoi(r.Response.Header.Get("Retry-After"))
+	if err != nil || retryAfterSecs <= 0 {
+		return 0
+	}
+	return time.Duration(retryAfterSecs) * time.Second
+}
+
 func (n *Client) post(ctx context.Context, buffer *bytes.Buffer, data interface{}) error {
 	post, err := n.constructPost(ctx, buffer, data)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


malderete:
@tiedotguy, @MovieStoreGuy please can I get a review on this one?
Thanks in advance!

pkg/backends/newrelic/newrelic.go
```diff
@@ -295,6 +295,26 @@ func (n *Client) Name() string {
 	return BackendName
 }
 
+// RetryAfterError indicates if we should wait before retrying
+type RetryAfterError struct {
+	Response *http.Response
```
tiedotguy, line 300: This mixes concerns - the error is about "retry the operation after a provided interval", but the implementation is "here's an http response that failed, figure out the rest yourself".

The parsing of the header should be done immediately, and `RetryAfterError` should just contain a `time.Duration`.  It also simplifies the error handling, because if you can't parse it, or the header is missing, you can fall back to the default error path.

pkg/backends/newrelic/newrelic.go
```diff
@@ -307,13 +327,31 @@ func (n *Client) post(ctx context.Context, buffer *bytes.Buffer, data interface{
...
+	var retryAfterErr *RetryAfterError
+	canRetryRateLimit := true
 	for {
 		if err = post(); err == nil {
 			atomic.AddUint64(&n.batchesSent, 1)
 			return nil
 		}
 
-		next := b.NextBackOff()
+		// If New Relic limitted us we should retry after
```
tiedotguy, line 340: This will retry only one (if we're in the 429 case), which may be intended, but it also means we're ignoring the configured `n.maxRequestElapsedTime`.

There's two concerns:
- Should we keep retrying until we reach the configured maximum, or only retry once?
- We should never retry for longer than the configured `maxRequestElapsedTime`, even if NR tells us to wait that long.

I would suggest using a pattern more like:
```
next := b.NextBackOff()
if next != backoff.Stop && errors.As(err, &retryAfterError) {
  next = max(next, retryAfterError.RetryAfter())
  if n.maxRequestElapsedTime > 0 {
    // this is not ideal, because it means we could potentially sleep for 2*n.maxRequestElapsedTime,
    // but it does put an upper bound on the duration.  The proper solution would be to track the
    // start time of the overall request, and make sure we don't sleep past that, but this is good enough.
    next = min(next, n.maxRequestElapsedTime)
  }
}
```

pkg/backends/newrelic/newrelic.go
```diff
@@ -407,13 +445,20 @@ func (n *Client) postWrapper(ctx context.Context, json []byte, dataType string)
 		if err != nil {
 			return fmt.Errorf("error POSTing: %s", err.Error())
 		}
+
+		if resp.StatusCode == http.StatusTooManyRequests {
+			return &RetryAfterError{Response: resp}
```
tiedotguy, line 450: This block should be moved in to the `if` block below, so that the `defer` and `ReadAll` are both handled - if you don't properly close a response body (ie, read all the data and call Close), then you don't get connection reuse.

You could put this check between the `ReadAll` and `logger.Info`, that way we do the read, but don't log when we know it's not a problem.

- [ ] properly close response body

tiedotguy (changes requested):
Requested changes inline

pkg/backends/newrelic/newrelic.go
malderete: Thanks for your review.

Indeed, I was not sure about putting a limit on the number of retries. I think, the approach you mentioned covers more and use `maxRequestElapsedTime`, I will use that.

pkg/backends/newrelic/newrelic.go
malderete: I've addressed this in last commit.

pkg/backends/newrelic/newrelic.go
malderete: Implemented in last commit

pkg/backends/newrelic/newrelic.go
malderete: I've addressed this in last commit.

pkg/backends/newrelic/newrelic.go
```diff
@@ -307,13 +317,24 @@ func (n *Client) post(ctx context.Context, buffer *bytes.Buffer, data interface{
 	b.Clock = clck
 	b.Reset()
 	b.MaxElapsedTime = n.maxRequestElapsedTime
+
+	var next time.Duration
+	var retryAfterErr *RetryAfterError
```
tiedotguy, line 322: Purely as a stylistic thing, `next` can be a `:=`, and `retryAfterErr` should be declared where it's consumed.  No need to change it though, just a consideration.

tiedotguy (approved):
lgtm!

tiedotguy:
I'll get a release cut shortly.

PR #455, opened 2022-02-03
Add rewrite rule for go-header as repo has moved

aelse:
Fix account takeover vuln. go-header has moved and there's a redirect in place on github. An attacker could register the old name and inject malicious code.

This change does a rewrite to the new repo location. The dependency is pulled in by golangci-lint, which i'll put in a separate change for.


aelse:
Opened https://github.com/golangci/golangci-lint/pull/2549 to correct this upstream

MovieStoreGuy (approved):

PR #459, opened 2022-02-07
Track forwarded metrics in preparation for clustering

tiedotguy:
This is a very noisy change, so it lives on its own.

Note: there is currently no source of forwarded metrics, it's purely foundational.


metric_map.go
```diff
@@ -46,7 +50,10 @@ func (mm *MetricMap) Receive(m *Metric) {
 }
 
 func MergeMaps(mms []*MetricMap) *MetricMap {
-	mm := NewMetricMap()
+	if len(mms) == 0 { // garbage in, garbage out
+		return nil
+	}
+	mm := NewMetricMap(mms[0].Forwarded)
```
MovieStoreGuy, line 56: Would we only want the first element here?

MovieStoreGuy (commented):
Just a query for you, but seems okay otherwise.

metric_map.go
tiedotguy: Not 100% sure what you mean.

We assume that all maps have the same forwarded value.  If they don't then it's a logic bug somewhere else.  I don't want to deal with returning an error from this, so we just assume the first thing is correct.

MovieStoreGuy: Is the slice ever going to container more than one entry in it?

tiedotguy: Yep - `MetricsConsolidator` spits out a slice on Drain/Flush

MovieStoreGuy (approved):

metric_consolidator.go
```diff
@@ -27,14 +27,19 @@ type MetricConsolidator struct {
 	maps          chan *MetricMap
 	sink          chan<- []*MetricMap
 	flushInterval time.Duration
+	forwarded     bool
```
MovieStoreGuy, line 30: What does forwarded do here?

metric_map.go
```diff
@@ -15,10 +15,14 @@ type MetricMap struct {
 	Timers   Timers
 	Gauges   Gauges
 	Sets     Sets
+
+	Forwarded bool
```
MovieStoreGuy, line 19: Should this be public and not part of the above group?

metric_map.go
MovieStoreGuy, line 56: This is an assumption that all metric maps are forwarded, is that safe to do so?

pkg/statsd/parser.go
```diff
@@ -116,7 +116,7 @@ func (dp *DatagramParser) Run(ctx context.Context) {
 				accumB += badLineCount
 			}
 			// TODO: Refactor this to use a MetricConsolidator
```
MovieStoreGuy, line 118: Casual TODO above your changes? :P

MovieStoreGuy (commented):
I have forgotten a lot of this PR, please remind me

MovieStoreGuy (approved):

metric_consolidator.go
tiedotguy: A consolidator will emit all forwarded or all non-forwarded.  It ignores the flag on the map, and maps it output will be set to this.

metric_map.go
tiedotguy: It's an assumption that everything in the slice has the same Forwarded flag, which they should.  We don't do any enforcement of this, because that means we need to propagate an error out, and it's not worth it.

metric_map.go
tiedotguy: I believe it will eventually be consumed outside the package, it's been a while since I wrote the code that comes next, but I think that's the case.

pkg/statsd/parser.go
tiedotguy: It'd be cool to do the todo. One day.

metric_map.go
tiedotguy: Technically we would panic, since an invariant has been broken.

PR #470, opened 2022-03-10
Now Publishes to Atlassian Public docker

ppapas123:
Now Publishes to Atlassian Public docker


MovieStoreGuy (commented):
I am fine for this to go in once we confirm that the token has been uploaded to Travis.

I'll approve the PR once that happens.

PR #472, opened 2022-03-18
Allow using Unix Domain Socket instead of UDP

rubenruizdegauna:
Allow using Unix Domain Sockets instead of UDP for the server.
If the `metrics-addr` is not empty and not `[host]:port` combination it will be considered a Unix Socket path and it will create and listen in that socket instead of UDP.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,9 +47,9 @@ to run `gostatsd` with a graphite backend and a grafana dashboard.
 While not generally tested on Windows, it should work.  Maximum throughput is likely to be better on
 a linux system, however.
 
-The server listens for UDP packets by default. Unix Domain Sockets (UDS) can be used specifying a file path instead 
-of `address:port` with the `metrics-addr` configuration option. This only works on linux and will ignore `conn-per-reader` 
-configuration option.
+The server listens for UDP packets by default. You can use unix sockets providing an absolute path to the socket 
+in the `metrics-addr` configuration option. The socket mode used in this case is SOCK_DGRAM.
+Note that using unix sockets will only work on linux and that it will ignore `conn-per-reader` configuration option.
 
 Configuring the server mode
 ---------------------------
@@ -98,8 +98,8 @@ This configuration mode allows the following configuration options:
   Defaults to `false`.
 - `receive-batch-size`: the number of datagrams to attempt to read.  It is more CPU efficient to read multiple, however
   it takes extra memory.  See [Memory allocation for read buffers] section below for details.  Defaults to 50.
-- `conn-per-reader`: attempts to create a connection for every UDP receiver.  Not supported by all OS versions. Will be
-  ignored if UDS is used instead of UDP.
+- `conn-per-reader`: attempts to create a connection for every UDP receiver.  Not supported by all OS versions. It will 
+  be ignored when unix sockets are used for the connection.
   Defaults to `false`.
 - `bad-lines-per-minute`: the number of metrics which fail to parse to log per minute.  This is used to prevent a bad
   client spamming malformed statsd data, while still logging some information to enable troubleshooting.  Defaults to `0`.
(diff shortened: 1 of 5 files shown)


aelse (commented):
Thanks @rubenruizdegauna . I would prefer not to make this the default behaviour applied to a metrics-addr string. We could use either a uri form unix:///path/to/socket (as some other systems do) or an abs path /path/to/socket to indicate unix domain sockets. I'd be happy with the abs path and we can identify a unix socket by looking for prefix '/'.

README.md
```diff
@@ -47,6 +47,10 @@ to run `gostatsd` with a graphite backend and a grafana dashboard.
 While not generally tested on Windows, it should work.  Maximum throughput is likely to be better on
 a linux system, however.
 
+The server listens for UDP packets by default. Unix Domain Sockets (UDS) can be used specifying a file path instead
```
aelse, line 50: Rather than abbreviating to UDS (which afaik is not generally used) we could use a shorthand of 'unix sockets' or 'IPC sockets'. I don't want to nitpick but we should use the accepted terminology - if i'm wrong and UDS is a commonly used acronym for this then that's fine.

README.md
```diff
@@ -47,6 +47,10 @@ to run `gostatsd` with a graphite backend and a grafana dashboard.
 While not generally tested on Windows, it should work.  Maximum throughput is likely to be better on
 a linux system, however.
 
+The server listens for UDP packets by default. Unix Domain Sockets (UDS) can be used specifying a file path instead 
+of `address:port` with the `metrics-addr` configuration option. This only works on linux and will ignore `conn-per-reader` 
+configuration option.
```
aelse, line 52: Somewhere the docs should mention the socket mode is SOCK_DGRAM not SOCK_STREAM

aelse (changes requested):
Please change this to treat an absolute path as a socket and avoid the UDS acronym (unless that genuinely is the widely recognised acronym)

pkg/statsd/receiver_test.go
```diff
@@ -2,12 +2,15 @@ package statsd
...
 import (
 	"context"
+	"net"
+	"os"
 	"runtime"
 	"sync"
 	"testing"
 	"time"
 
 	"github.com/magiconair/properties/assert"
```
tiedotguy, line 12: I didn't realize we had other assert libs. Raised #480 to remove others.

pkg/statsd/statsd_test.go
```diff
@@ -2,8 +2,10 @@ package statsd
 
 import (
 	"context"
+	"github.com/stretchr/testify/assert"
```
tiedotguy, line 5: There should be a check somewhere for this, not sure why the build didn't hit it.  Import ordering is:

- stdlib
- 3rd party
- local

tiedotguy (changes requested):

README.md
rubenruizdegauna: 👍  I've used `unix sockets` and  rephrased it a bit in https://github.com/atlassian/gostatsd/pull/472/commits/11d015b2bd6251deb18b8db3ae611374181465aa

README.md
rubenruizdegauna: I've added it to this paragraph in https://github.com/atlassian/gostatsd/pull/472/commits/11d015b2bd6251deb18b8db3ae611374181465aa. I wasn't sure if it was too much info about the socket in this section and I thought about adding another section just for this. Let me know what do you think and I can quickly change it.

rubenruizdegauna:
Thanks @aelse @tiedotguy for the quick review. I have made the requested changes in https://github.com/atlassian/gostatsd/pull/472/commits/11d015b2bd6251deb18b8db3ae611374181465aa

aelse (approved):

aelse:
lgtm. If @tiedotguy is happy we can merge.

rubenruizdegauna:
> lgtm. If @tiedotguy is happy we can merge.

Hi @tiedotguy , do you think this can be merged? Thanks!

tiedotguy (approved):

Issue #481, opened 2022-03-27
golangci-lint times out on arm making the build flaky


tiedotguy:
We should just skip it on arm, it's ran on x64 and doesn't time out.

PR #492, opened 2022-04-05
Sfarsaci/update alpine

sfarsaci:


.github/workflows/lint.yml
```diff
@@ -16,5 +16,5 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v2
         with:
-          version: v1.33
-          args: --timeout=5m
\ No newline at end of file
+          version: v1.45.2
```
tiedotguy, line 19: Does golangci-lint as an action mean we don't need to run golangci-lint in the build?  (#481 relates)

I'm about to throw in a PR to disable bench-race and golangci-lint for arm64, but if I can remove golangci-lint from both, that's even better.

tiedotguy (approved):

.github/workflows/lint.yml
MovieStoreGuy: Yeah, it means it can be canned.

MovieStoreGuy (approved):

PR #497, opened 2022-04-18
New relic statsd source

sdewitt-newrelic:
## Change summary

The purpose of this change is to allow NRQL queries to `FACET` by the source of a given `Metric` coming from a host through `gostatsd`.

* Added `maybeAddSource` to add the `statsdSource` tag to metrics in the New Relic backend.

### DIFF
--- a/BACKENDS.md
+++ b/BACKENDS.md
@@ -285,3 +285,18 @@ Additional options are available to rename attributes if required.
 	timer-sum = "samples_sum"
 	timer-sumsquare = "samples_sum_squares"
 ```
+
+### Host Tag
+When `--ignore-host` is not set, the New Relic backend will add the tag `statsdSource` to
+all metrics. The value of this tag will be the `gostatsd.Source` string recorded
+with each metric. So as not to collide with the existing `host` attribute that
+is added to many infrastructure samples, the name `statsdSource` was chosen
+instead of `host`.
+
+*NOTE:* The addition of the `statsdSource` tag can potentially cause an
+explosion in [cardinality](https://docs.newrelic.com/docs/data-apis/ingest-apis/metric-api/NRQL-high-cardinality-metrics/).
+To avoid hitting cardinality limits, the `--ignore-host` option is the default
+when using the New Relic backend. To enable this functionality, supply
+`--ignore-host false` and be sure to watch your cardinality closely with a query
+like `SELECT cardinality() FROM Metric WHERE integration.name='GoStatsD'` or
+`SELECT cardinality() FROM Metric WHERE integration.name='GoStatsD' FACET metricName`.
--- a/BACKENDS.md
+++ b/BACKENDS.md
@@ -287,8 +287,8 @@ Additional options are available to rename attributes if required.
 ```
 
 ### Host Tag
-When `--ignore-host` is not set, the New Relic backend will add the tag `statsdSource` to
-all metrics. The value of this tag will be the `gostatsd.Source` string recorded
+When `--ignore-host` is not set, the New Relic backend will add the tag `statsdSource`
+to all metrics. The value of this tag will be the `gostatsd.Source` string recorded
 with each metric. So as not to collide with the existing `host` attribute that
 is added to many infrastructure samples, the name `statsdSource` was chosen
 instead of `host`.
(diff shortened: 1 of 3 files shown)


tiedotguy:
Hi,

Looks good generally, although this would typically be the `host` tag.  If `statsdSource` is the official NR thing though, then that's fine though.

Please also update BACKENDS.md and CHANGELOG.md (major bump, because this is a change in behavior), calling out that admins may want to use the `--ignore-host` flag.

There is a potential hazard here where there's suddenly a large cardinality explosion as this starts being added, although I don't know what the charging model is for NR.  I wouldn't want customers to experience bill shock as a result of this.

You may want to consider setting the default for `ignore-host` to `true` in [NewClientFromViper](https://github.com/atlassian/gostatsd/blob/master/pkg/backends/newrelic/newrelic.go#L493) to pre-emptively avoid it.

sdewitt-newrelic:
Thanks @tiedotguy.  I used `statsdSource` because if the New Relic backend can pump it's data through our infrastructure agent and I believe that already has a `host` attribute but it is the `host` of the node the infrastructure agent is running on which may not match the node the statsd data is coming from.  @Kav91 can you confirm if this is true?  I have never pushed data via the internal infrastructure http agent.

I will update the files you call out.  Thanks.  When I looked at the `--ignore-host` flag, it seemed to me like this only applied to the internal metrics used if you are using the internal `statser`.  Is this true?  If so, the New Relic backend sets the `statser` to `null` anyway.  If this is not true, when using this flag, does it _already_ add `host` as a dimension on every metric?

Thanks you for the call out on cardinality.  I hadn't considered that.  We don't charge based on that but we do have limits on cardinality so I really appreciate the call out and I will need to add something about this to the `BACKENDS.md`.  I wonder if I should take into account the `--ignore-host` setting or will the `Source` already be `null` if this setting is `true` (in which case my code already handles it).

Kav91:
@sdewitt-newrelic yes sometimes host, hostname, fullHostname are attributes (among others) that can be potentially stamped automatically when going through the agent.

Also for this PR can you increment the integrationVersion say to 2.4.0 in newrelic.go, then we can track this newer version easily.

sdewitt-newrelic:
@tiedotguy added section to `BACKENDS.md`, bumped backend version, and set `--ignore-host` to `true` by default for the `newrelic` backend.  When you mentioned `CHANGELOG.md (major bump, because this is a change in behavior)`, did you want it bumped to `35.0.0` then?  Also it looks like the version in the binary is sucked in at build time from the git tag.  I presume you or someone else actually creates the tag?

gorakong:
pushed a dummy commit to re-trigger the CLA check. Feel free to revert it 🙂

sdewitt-newrelic:
@tiedotguy can you comment on my last comment?

sdewitt-newrelic:
@MovieStoreGuy @tiedotguy Any chance I could get a review on my last comments?  Would like to get this merged if possible. Thanks!

tiedotguy:
Hi, taking another look now!

BACKENDS.md
```diff
@@ -285,3 +285,18 @@ Additional options are available to rename attributes if required.
...
+to all metrics. The value of this tag will be the `gostatsd.Source` string recorded
+with each metric. So as not to collide with the existing `host` attribute that
+is added to many infrastructure samples, the name `statsdSource` was chosen
+instead of `host`.
+
+*NOTE:* The addition of the `statsdSource` tag can potentially cause an
+explosion in [cardinality](https://docs.newrelic.com/docs/data-apis/ingest-apis/metric-api/NRQL-high-cardinality-metrics/).
+To avoid hitting cardinality limits, the `--ignore-host` option is the default
+when using the New Relic backend. To enable this functionality, supply
+`--ignore-host false` and be sure to watch your cardinality closely with a query
```
tiedotguy, line 300: TIL you can use `--flag false` with a boolean flag.

tiedotguy (approved):

tiedotguy:
Happy to merge, I'll write a brief changelog for it!

sdewitt-newrelic:
Thanks so much @tiedotguy for the comments and help and for the merge!

Issue #524, opened 2022-06-23
Parse dogstatsd container id


tiedotguy:
https://github.com/DataDog/datadog-go/pull/250

https://docs.datadoghq.com/developers/dogstatsd/datagram_shell/?tab=metrics&tabs=metrics#dogstatsd-protocol-v12

tl;dr datadog added an extra field on their line protocol, which their client adds unless you actively stop it /sigh

We need to parse it in to the source field, at which point `--ignore-host` should handle it as necessary.

tiedotguy:
Done in #530

PR #530, opened 2022-07-05
Lbest/container bugfix

lisanmason:
This PR addresses the issue [https://github.com/atlassian/gostatsd/issues/524]. A change to dogstatsd client to add a container attribute causes the last tag to have |c:{container_id} appended when parsed by gostatsd parser. 

Currently tags are assumed to be the last attribute for both events and metrics. To address the issue, the lexer is being updated to:

- accept optional attributes in any order including the tags attribute
- ignore any attribute that is unrecognised, including container

This change has been applied consistently to parsing of both the events and metrics.

Implementation notes:
- the initial implementation of this fix attempted to use lexAssert/lexUntil functions in the attribute parsing for unknown attributes. However, this seemed to increase the benchmark time to process per metric due to the return of anonymous functions.
- to maintain performance, lexUntil function was modified and renamed to seekUntil.
- to improve readability seekUntil was used to replace existing code for parsing sample rate
- a new function seekDelimited was created to replace the lextTags. There was some performance improvement due to inlining the lexTags implementation.


tiedotguy:
lgtm, but because of it's criticality, we should do an rc build, and get it in front of some services.  You can push a tag like `35.1.0-rc1` which will trigger the build (+docker build), then I believe we can consume that internally somehow.

internal/lexer/lexer.go
```diff
@@ -320,19 +339,35 @@ func lexAssert(nextByte byte, next stateFn) stateFn {
...
-		start := l.pos
-		p := bytes.IndexByte(l.input[l.pos:], stop)
-		switch p {
-		case -1:
-			l.pos = l.len
-		default:
-			l.pos += uint32(p)
+// seekUntil Seeks the next stop or eof byte and then return the byte slice between
+// the start position and the end byte.
+func seekUntil(l *Lexer, stop byte) []byte {
```
aelse, line 344: seekUntil and seekDelimited could both use their own unit tests in addition to the coverage offered by TestMetricsLexer

aelse:
lgtm but @tiedotguy can gatekeep this release

internal/lexer/lexer.go
lisanmason: Done

aelse (approved):
lgtm

tiedotguy (approved):

PR #533, opened 2022-07-15
Lbest/usegithub actions

lisanmason:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,87 @@
...
+    # minimal permissions for Get Artifact Publish Token
+    permissions:
+      contents: read
+      id-token: write
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.19.7
```
tiedotguy, line 42: We have a couple places where the go version is specified, they should be updated to refer to this, and this should be updated to refer to them, specifically:
- readme
- makefile
- dockerfiles

build/Dockerfile-multiarch
```diff
@@ -0,0 +1,29 @@
+FROM golang:1.17.2-alpine AS build
```
tiedotguy, line 1: 1.19.7 + previous comment

build/Dockerfile-multiarch-glibc
```diff
@@ -0,0 +1,31 @@
+FROM golang:1.17.2-alpine AS build
```
tiedotguy, line 1: Also 1.19.7

.travis.yml
```diff
@@ -6,60 +6,6 @@ services:
 go:
   - "1.19"  # Keep in sync with Makefile and README.md
```
tiedotguy, line 7: Do we need this file anymore at all?

tiedotguy (changes requested):
I don't know a lot about github actions, but it seems legit.  Couple actions:

- [ ] Use 1.19.7
- [ ] Reference golang versions at each usage site
- [ ] Maybe remove .travis.yml?

Also a couple questions around creds and such:
- Where do creds live, for dockerhub and slack?
- Are creds accessible to forks?

.github/workflows/ci.yml
jcmackie: Okay... Will comment

.github/workflows/ci.yml
```diff
@@ -0,0 +1,87 @@
...
+  release:
+    needs:
+      - build
+      - check
+    if: github.ref_type == 'tag' && (needs.check.outputs.branch == 'master' || needs.check.outputs.branch == 'main' || needs.check.outputs.branch == 'release')
+    runs-on: ubuntu-latest
+    # minimal permissions for Get Artifact Publish Token
+    permissions:
+      contents: read
+      id-token: write
```
jcmackie, line 69: @tiedotguy I'm looking into the way the Github does secrets, etc because you were asking about that before but the comment seems to have disappeared.

From my own usage of Github actions I know that you can set repository level secrets, but it seems like we don't have that kind of access, so let's assume we won't be using those.

These two permissions are for giving the Action the access it need to do its job...
`contents: read` are used to checkout the repo into the build environment
`id-token:write` are used to generate a JWT token, presumably unique to this repo.
See here: https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#adding-permissions-settings

So, I'm not sure how travisci worked (will have to look at the config) but it seems like we might need to set up access to push the image to `Dockerhub.com/atlassianlabs/` via JWT auth. It might the the case that any JWT from this org is permitted?

You were also asking about fork access to secrets: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token
> Maximum access for pull requests from public forked repositories [1]

This is `read` for `id-token`, which paradoxically means they don't get access to it.
The [1] there means:
> Private repositories can control if pull requests from forks can run workflows, and configure the permissions assigned to GITHUB_TOKEN. For more information, see "[Managing GitHub Actions settings for a repository](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#enabling-workflows-for-forks-of-private-repositories)."

From the documentation is looks like you need to go in and explicitly allow fork pull requests permission to run actions, via the settings page.

tiedotguy (approved):

PR #566, opened 2022-08-28
Rewrite node tracking

tiedotguy:
This should be considered as a from scratch rewrite of the node tracking stuff, and not a change.  Nothing uses it, and the whole approach has changed.


tiedotguy:
Pants. There's a race condition.

go.mod
```diff
@@ -28,8 +31,12 @@ require (
 	k8s.io/api v0.24.3
 	k8s.io/apimachinery v0.24.3
 	k8s.io/client-go v0.24.3
+	k8s.io/utils v0.0.0-20220210201930-3a6ce19ff2f9 // indirect
+	stathat.com/c/consistent v1.0.0
 )
 
+require github.com/alicebob/miniredis/v2 v2.23.0
```
MovieStoreGuy, line 38: Odd that this made its own group? 

`go mod tidy -compat=1.17` should resolve this into two groups.

internal/cluster/nodes/picker_consistent.go
```diff
@@ -0,0 +1,46 @@
+package nodes
+
+import (
+	"context"
+
+	"stathat.com/c/consistent"
```
MovieStoreGuy, line 6: Any reason to not use `hash/fnv`? Or does this do something extra I am missing?

MovieStoreGuy (dismissed):
So far all of these changes makes sense to me

go.mod
tiedotguy: I'm doing a big version bump + cleanup (I have protobuf problems on my wip branch, I think it's just because the branch is super old though), and it cleans this up.

internal/cluster/nodes/nodes.go
```diff
@@ -0,0 +1,59 @@
...
+	// Add will add the node to the list of nodes tracked.  Thread safe.
+	Add(node string)
+
+	// Remove will remove the node from the list of nodes tracked. Thread safe.
+	Remove(node string)
+}
+
+// NodeTracker will track a list of nodes, it is responsible for tracking life
+// cycle updates, and will typically pass them on to a NodePicker.
+type NodeTracker interface {
```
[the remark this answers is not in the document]
tiedotguy: That's an explicit non-goal (I think the original version did that)

A picker should do consistent hashing across some data source, the data source could be redis, ddb, static, k8s, etc.

There might be alternate hashing things to use, if this one doesn't work as well as we'd like.

internal/cluster/nodes/picker_consistent.go
tiedotguy: It does consistent hashing, so ideally when a node comes up or goes down, most things stay on the same node, and only a small amount of data moves around.  During a re-convergence, we want as much stuff to remain on the same host as possible, because we only get one forward.

Anything that's already been forwarded, and ends up on the wrong host due to the re-convergence, gets dropped.

tiedotguy:
It wants go 1.19.  I have a bunch of changes for that, I'll throw it on this PR too.  I don't like overloading it, but it is what it is.

Makefile
```diff
@@ -12,35 +12,31 @@ CPU_ARCH ?= amd64
 MANIFEST_NAME := atlassianlabs/$(BINARY_NAME)
 IMAGE_NAME := $(MANIFEST_NAME)-$(CPU_ARCH)
 ARCH ?= $$(uname -s | tr A-Z a-z)
-GOVERSION := 1.17.2  # Keep in sync with .travis.yml and README.md
+GOVERSION := 1.19  # Keep in sync with .travis.yml and README.md
 GP := /gopath
 MAIN_PKG := github.com/atlassian/gostatsd/cmd/gostatsd
 CLUSTER_PKG := github.com/atlassian/gostatsd/cmd/cluster
-PROTOBUF_VERSION ?= 3.6.1
```
MovieStoreGuy, line 19: That is an insane jump...

Makefile
```diff
@@ -12,35 +12,31 @@ CPU_ARCH ?= amd64
...
 setup-ci: tools/bin/protoc
 	@#go install github.com/golangci/golangci-lint/cmd/golangci-lint
 
 build-cluster:
 	go build -i -v -o build/bin/$(ARCH)/cluster $(GOBUILD_VERSION_ARGS) $(CLUSTER_PKG)
 
-pb/gostatsd.pb.go: pb/gostatsd.proto
-	GOPATH="tools/bin/protoc:${GOPATH}" protoc --go_out=.\
-		--go_opt=paths=source_relative \
-		--go-grpc_out=. \
```
tiedotguy, line 41: We don't use grpc, not sure how this snuck in, but it's ok to remove.

internal/cluster/nodes/tracker_static.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"context"
+
+	"github.com/atlassian/gostatsd"
+)
+
+// staticNodeTracker is a tracker with a static list of hosts
+type staticNodeTracker struct{}
+
+func NewStaticNodeTracker(picker consistentNodePicker, host []string) gostatsd.Runnable {
```
tiedotguy, line 12: Technically this should be returning a `NodeTracker`, but it's unused at present.  Awkward.

cmd/cluster/main.go
```diff
@@ -23,5 +27,16 @@ func main() {
...
+
+	ctx, cancel := context.WithCancel(context.Background())
+	chCancel := make(chan os.Signal)
+	signal.Notify(chCancel, syscall.SIGINT, syscall.SIGTERM)
+	go func() {
+		fmt.Printf("Waiting\n")
+		<-chCancel
+		fmt.Printf("cancelled\n")
+		cancel()
+	}()
```
MovieStoreGuy, line 39:
```suggestion
    ctx, done := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
```

Reduce the amount of cruft you need?

cmd/cluster/server.go
```diff
@@ -21,44 +24,56 @@ type Cluster struct {
 
 // newCluster will create a new Cluster with default values.
 func newCluster() *Cluster {
+	// 1.1.1.1 is used for route lookup, to determine the local interface that would
+	// be used to reach this address.  It doesn't actually talk to that address.
 	local, err := nodes.LocalAddress("1.1.1.1:1")
```
MovieStoreGuy, line 29: Does it need to be an outbound addr in order to get the interface addr that is used to connect to the outside world?

ie. does trying to dial `127.0.0.1` give you an interface that isn't compatible?

cmd/cluster/server.go
```diff
@@ -21,44 +24,56 @@ type Cluster struct {
...
-	go rnt.Run(ctx)
+func (c *Cluster) Run(ctx context.Context) {
+	picker := nodes.NewConsistentNodePicker(c.Target, 20)
+
+	options := &redis.Options{
+		Addr: c.RedisAddr,
+		DB:   0,
+	}
+
+	redisClient := redis.NewClient(options)
```
MovieStoreGuy, line 61: I can't see it here, but it is worth using a universal client here from `go-redis` so that you can use single host or clustered redis?

go.mod
```diff
@@ -19,15 +21,19 @@ require (
...
 	golang.org/x/time v0.0.0-20220210224613-90d013bbcef8
 	golang.org/x/tools v0.1.12
 	google.golang.org/protobuf v1.28.1
-	k8s.io/api v0.24.3
-	k8s.io/apimachinery v0.24.3
-	k8s.io/client-go v0.24.3
+	k8s.io/api v0.25.0
+	k8s.io/apimachinery v0.25.0
+	k8s.io/client-go v0.25.0
+	k8s.io/utils v0.0.0-20220728103510-ee6ede2d64ed // indirect
```
MovieStoreGuy, line 35: I wish these could go

internal/cluster/nodes/tracker_redis.go
```diff
@@ -0,0 +1,202 @@
...
+			rnt.handleMessage(ctx, msg.Payload, clck.Now())
+		}
+	}
+}
+
+func (rnt *redisNodeTracker) handleMessage(ctx context.Context, message string, now time.Time) {
+	if message == "" {
+		return
+	}
+	node := message[1:]
```
MovieStoreGuy, line 112: Does this return bytes or runes here?

internal/cluster/nodes/tracker_redis.go
```diff
@@ -0,0 +1,202 @@
...
+		rnt.refreshNode(node, now)
+		if node != rnt.nodeId {
+			// It's not us, and it wants to know about us, send a broadcast out to let it know we exist.
+			if err := rnt.sendHeartbeat(ctx); err != nil {
+				rnt.logger.WithError(err).WithField("newNode", node).Warning("Failed to send introduction reply")
+			}
+		}
+	case '+':
+		rnt.refreshNode(node, now)
+	}
```
MovieStoreGuy, line 126: No default?

MovieStoreGuy (approved):

Makefile
tiedotguy: Check the gostatsd.pb.go generated file, it's not what it looks like.

cmd/cluster/server.go
tiedotguy: Yeah, it would return 127.0.0.1:
```
1.1.1.1:1111 -> 10.0.2.15 (default)
127.0.0.1:1234 -> 127.0.0.1
192.168.56.1:1234 -> 192.168.56.101 (virtualbox internal network)
```

go.mod
tiedotguy: Yeah, we pull in a ton of crap for k8s support :(

cmd/cluster/server.go
tiedotguy: /nods

I'll look in to it.  For now I just want to get this merged.

cmd/cluster/main.go
tiedotguy: I'll include it in the fix for the gofmt /tableflip

internal/cluster/nodes/tracker_redis.go
tiedotguy: No idea. This isn't a py3 port, we're dealing with hostnames.  Pretty sure they need to be ascii.

internal/cluster/nodes/tracker_redis.go
tiedotguy: Garbage in, garbage out.

internal/cluster/nodes/tracker_redis.go
MovieStoreGuy: Spent to much time watching string fuzzing, questioning everything!

internal/cluster/nodes/tracker_static.go
MovieStoreGuy: Do you have a fix for this in future?

tiedotguy: Not originally, but I will.  In the original draft I just returned a runnable, but since realized I'm probably going to extend the interface, so it's better to give myself the option now.

PR #590, opened 2023-01-31
Allow parallel metrics merging with HTTP forwarder mode

beanliu:
In the forwarder mode, metrics merging is a blocking operation which may introduce significant back pressure to the UDP consumption if there is large load. This patch adds a configuration `concurrent-merge` to allow for parallel processing of the merging operation. The default value is set to 1 to maintain backward-compatibility since it aggravates the out-of-orderness of the forwarding metrics.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+35.1.3
+------
+- HTTP forwarder updated with a new option `concurrent-merge` to support parallel processing, the
+  default value is backward-compatible.
+
 35.1.2
 ------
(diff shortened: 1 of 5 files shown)


tiedotguy:
lgtm - can you add a CHANGELOG? I think because it's backwards compatible, we can just have a patch bump.

beanliu:
> lgtm - can you add a CHANGELOG? I think because it's backwards compatible, we can just have a patch bump.

updated the CHANGELOG with `35.1.3`.

tiedotguy (approved):
lgtm

PR #598, opened 2023-03-16
Make Release a manual action

jcmackie:
Due to our release workflow:

1. Create PR
2. Merge PR
3. Tag master
4. Push tag to master

And the restriction on being able to get an Artifactory deployment token only on the master branch we have to do releases slightly differently.

This change moves the 'release' job into it's own workflow, which is manual and can only be run in the master branch.
The steps ensure that the latest tag and master are pointing at the same commit.

For now the actual docker push steps are commented out until we can be sure that the rest of the logic works (including getting a token).


tiedotguy (approved):

.github/workflows/manual-release.yml
```diff
@@ -0,0 +1,44 @@
...
+        uses: docker/setup-buildx-action@master
+      # - name: Release
+      #   run: |
+      #     echo "Executing release on tag build $GITHUB_REF_NAME"
+      #     ARCH=linux CPU_ARCH=amd64 make release-ci
+      #     ARCH=linux CPU_ARCH=arm64 make release-ci
+      # - name: Push Docker Manifest
+      #   run: |
+      #     echo "Executing release-manifest on tag build $GITHUB_REF_NAME"
+      #     make release-manifest-ci
```
MovieStoreGuy, line 44: Are these still required or could that be nuked?

tiedotguy: I believe we still want them, it's just commented out for testing purposes, then it'll be enabled if the builds all work.

jcmackie: yes, exactly that.

PR #629, opened 2023-07-26
Feature: Improve internal observability when running in forwarder mode

MovieStoreGuy:
## Context

When running a large deployment, it is possible that the internal metrics can contribute a significant amount of traffic just at idle.

## Changes

- Added a feature to disable internal events
- Changed a subset of internal metrics to be non cumulative counters


jordan-simonovski (dismissed):

pkg/statsd/parser.go
```diff
@@ -85,8 +85,8 @@ func (dp *DatagramParser) RunMetricsContext(ctx context.Context) {
 		case <-ctx.Done():
 			return
 		case <-flushed:
-			statser.Gauge("parser.metrics_received", float64(atomic.LoadUint64(&dp.metricsReceived)), nil)
-			statser.Gauge("parser.events_received", float64(atomic.LoadUint64(&dp.eventsReceived)), nil)
+			statser.Report("parser.metrics_received", float64(atomic.LoadUint64(&dp.metricsReceived)), nil)
```
tiedotguy, line 88: These will still accumulate.  I think a better approach is to pass a pointer, and then have `Report` do the `LoadUint64` vs `SwapUint64(0)`.

Nothing actually needs it to be a float64, and if we define it as a "counter type thing", nothing needs to.

pkg/stats/heartbeat.go
```diff
@@ -37,5 +37,5 @@ func (hb *HeartBeater) Run(ctx context.Context) {
 }
 
 func (hb *HeartBeater) emit(statser Statser) {
-	statser.Gauge(hb.metricName, 1, nil)
+	statser.Report(hb.metricName, 1, nil)
```
tiedotguy, line 40: It's a bit messy, but a
```
one := uint64(1)
report(&one)
```

would work with the below comment.

pkg/statsd/parser.go
tiedotguy: - [ ] Pass pointer, and have atomic ops in the statser

METRICS.md
```diff
@@ -11,58 +11,59 @@ Metric types:
...
+| aggregator.aggregation_time                 | gauge (time)          | aggregator_id                | The time taken (in ms) to aggregate all counter and timer
+|                                             |                       |                              | datapoints in this flush interval
+| aggregator.process_time                     | gauge (time)          | aggregator_id                | The time taken to process all synchronous flush actions
+| aggregator.reset_time                       | gauge (time)          | aggregator_id                | The time taken to reset the aggregator after flush
+| parser.bad_lines_seen                       | gauge (sparse)        |                              | The number of unparseable lines
+| parser.events_received                      | reported              |                              | The number of events parsed
+| parser.metrics_received                     | reported              |                              | The number of metrics parsed
+| receiver.datagrams_received                 | reported (cumulative) |                              | The number of datagrams received
```
tiedotguy, line 29: - [ ] Remove "cumulative"

pkg/statsd/receiver.go
```diff
@@ -68,8 +68,8 @@ func (dr *DatagramReceiver) RunMetricsContext(ctx context.Context) {
 			} else {
 				avgDatagramsInBatch = float64(datagramsReceived) / float64(batchesRead)
 			}
-			statser.Gauge("receiver.datagrams_received", float64(dr.cumulDatagramsReceived), nil)
-			statser.Gauge("receiver.avg_datagrams_in_batch", avgDatagramsInBatch, nil)
+			statser.Report("receiver.datagrams_received", float64(dr.cumulDatagramsReceived), nil)
+			statser.Report("receiver.avg_datagrams_in_batch", avgDatagramsInBatch, nil)
```
tiedotguy, line 72: - [ ] Replace with `batchesRead`

tiedotguy (changes requested):
sdfgsdgdf

pkg/statsd/parser.go
MovieStoreGuy: So keep the same logic that Report can either be a Gauge or Count but have it accept a pointer to a value that it can reset on each report?

pkg/statsd/receiver.go
MovieStoreGuy: I made the name `batches_read` to keep in line with other metric names.

tiedotguy (approved):

PR #633, opened 2023-10-12
[chore] Update Go Version to 1.21.3

MovieStoreGuy:
## Context

The latest version of go fixes a published [VULN](https://pkg.go.dev/vuln/GO-2023-2102) , this is to address that potential concern. It should be limited to HTTP/2 connections, however, it is in our best interest to update the version to ensure that patch is applied.

## Changes.

- Updated go version to 1.21.3


jamesmoessis (approved):

build/Dockerfile-multiarch
```diff
@@ -1,5 +1,5 @@
 # Go version needs to be the same in: CI config, README, Dockerfiles, and Makefile
-FROM golang:1.19.7-alpine AS build
+FROM golang:1.21.3 AS build
```
jamesmoessis, line 2: getting rid of alpine? Is this a breaking change?

jamesmoessis: Oh never mind, just the build step

Excaleo (approved):

PR #652, opened 2024-01-09
Add Lambda extension telemetry server

Excaleo:
Currently the lambda extension binary will flush periodically based on the flush interval. Because the gostatsd runtime is suspended once the invocation is finished, the metric may:
- Not flush at all
- Flush at the next invocation

To address the flakiness of the flush when the lambda is not constantly invoked, we can subscribe to the telemetry API that AWS provides and listen for the `platform.runtimeDone` event to flush on within the extension.

This PR sets up the initial step to start a server and listen to lambda telemetry events, with a hook function defined on the above event.

A subsequent PR will add the logic to trigger a flush using the hook function.

### DIFF
--- a/internal/awslambda/extension/manager.go
+++ b/internal/awslambda/extension/manager.go
@@ -66,7 +66,7 @@ func NewManager(lambdaDomain string, lambdaFileName string, log logrus.FieldLogg
 		name:   lambdaFileName,
 	}
 
-	m.telemetryServer = telemetry.NewServer(log, func() {})
+	m.telemetryServer = telemetry.NewServer(log, telemetry.NoopHook())
 
 	return m
 }
@@ -242,6 +242,9 @@ func (m *manager) subscribeToTelemetry(ctx context.Context) error {
 		return err
 	}
 	req, err := http.NewRequestWithContext(ctx, http.MethodPut, telemetry.SubscribeEndpoint.GetUrl(m.domain), bytes.NewReader(b))
+	if err != nil {
+		return err
+	}
 	req.Header.Set(api.LambdaExtensionIdentifierHeaderKey, m.registeredID)
 	req.Header.Set("Content-Type", "application/json")
 
@@ -256,11 +259,14 @@ func (m *manager) subscribeToTelemetry(ctx context.Context) error {
 
 	if resp.StatusCode != http.StatusOK {
 		body, _ := io.ReadAll(resp.Body)
-		m.log.WithError(err).Errorf("received %d response code from telemetry subscription with body %s", resp.StatusCode, string(body))
+		m.log.WithError(err).WithFields(map[string]interface{}{
+			"statusCode": resp.StatusCode,
+			"msg":        string(body),
+		}).Error("Error subscribing to telemetry API")
 		return ErrFailedTelemetrySubscription
 	}
 
-	m.log.Info("successfully subscribed to telemetry API")
+	m.log.Info("Successfully subscribed to telemetry API")
 
 	return nil
 }
--- a/internal/awslambda/extension/manager.go
+++ b/internal/awslambda/extension/manager.go
@@ -66,7 +66,7 @@ func NewManager(lambdaDomain string, lambdaFileName string, log logrus.FieldLogg
 		name:   lambdaFileName,
 	}
 
-	m.telemetryServer = telemetry.NewServer(log, telemetry.NoopHook())
+	m.telemetryServer = telemetry.NewServer(telemetry.LambdaRuntimeAvailableAddr, log, telemetry.NoopHook())
 
 	return m
 }
--- a/internal/awslambda/extension/manager.go
+++ b/internal/awslambda/extension/manager.go
@@ -18,6 +18,7 @@ import (
 	"go.uber.org/multierr"
 
 	"github.com/atlassian/gostatsd/internal/awslambda/extension/api"
+	"github.com/atlassian/gostatsd/internal/awslambda/extension/telemetry"
 )
 
 const (
@@ -26,9 +27,10 @@ const (
 )
 
 var (
-	ErrFailedRegistration = errors.New("extension failed to register with lambda")
-	ErrIssueProgress      = errors.New("unable to continue execution")
-	ErrServerEarlyExit    = errors.New("server has shutdown early without cause")
+	ErrFailedRegistration          = errors.New("extension failed to register with lambda")
+	ErrIssueProgress               = errors.New("unable continue execution")
+	ErrServerEarlyExit             = errors.New("server has shutdown early without cause")
+	ErrFailedTelemetrySubscription = errors.New("extension failed to subscribe to lambda telemetry API")
 )
 
 // Manager ensures that the lifetime management of the lambda
@@ -50,6 +52,8 @@ type manager struct {
 	domain       string
 	name         string
 	registeredID string
+
+	telemetryServer *telemetry.Server
 }
 
 var _ Manager = (*manager)(nil)
@@ -62,21 +66,33 @@ func NewManager(lambdaDomain string, lambdaFileName string, log logrus.FieldLogg
 		name:   lambdaFileName,
 	}
 
+	m.telemetryServer = telemetry.NewServer(log, func() {})
+
 	return m
 }
 
 func (m *manager) Run(parent context.Context, server Server) error {
+	var wg wait.Group
+	var chErrs = make(chan error, 3)
+
 	// Init Phase of the lambda
 	if err := m.register(parent); err != nil {
 		return err
 	}
 
+	// Start telemetry server
 	ctx, cancel := context.WithCancel(parent)
-	// Start gostatsd server
-	var wg wait.Group
-	var chErrs = make(chan error, 2)
 	defer cancel()
+	wg.StartWithContext(ctx, func(c context.Context) {
+		chErrs <- m.telemetryServer.Start(c)
+	})
 
+	if err := m.subscribeToTelemetry(ctx); err != nil {
+		m.log.WithError(err).Error("error subscribing to lambda telemetry endpoint")
+		return err
+	}
+
+	// Start gostatsd server
 	wg.StartWithContext(ctx, func(c context.Context) {
 		err := server.Run(c)
 		// Shutting down due to context is an acceptable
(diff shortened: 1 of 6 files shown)


internal/awslambda/extension/manager.go
```diff
@@ -214,6 +230,41 @@ func (m *manager) register(ctx context.Context) error {
...
+		SchemaVersion: telemetry.ApiVersion,
+		Types:         []string{telemetry.PlatformSubscriptionType},
+		Destination:   &telemetry.SubscriptionDestination{URI: m.telemetryServer.Endpoint(), Protocol: telemetry.ProtocolHTTP},
+		// we buffer for the minimum time on the lambda side to avoid adding additional billedDuration to the lambda
+		Buffering: &telemetry.SubscriptionBufferingConfig{TimeoutMs: telemetry.MinBufferingTimeoutMs},
+	})
+	if err != nil {
+		return err
+	}
+	req, err := http.NewRequestWithContext(ctx, http.MethodPut, telemetry.SubscribeEndpoint.GetUrl(m.domain), bytes.NewReader(b))
```
MovieStoreGuy, line 244:
```suggestion
	req, err := http.NewRequestWithContext(
		ctx,
		http.MethodPut,
		telemetry.SubscribeEndpoint.GetUrl(m.domain),
		bytes.NewReader(b),
	)
	if err != nil {
		return err
	}
```
Missing check for the returned error

internal/awslambda/extension/manager.go
```diff
@@ -214,6 +230,41 @@ func (m *manager) register(ctx context.Context) error {
...
+		return err
+	}
+	defer func() {
+		_, _ = io.Copy(io.Discard, resp.Body)
+		_ = resp.Body.Close()
+	}()
+
+	if resp.StatusCode != http.StatusOK {
+		body, _ := io.ReadAll(resp.Body)
+		m.log.WithError(err).Errorf("received %d response code from telemetry subscription with body %s", resp.StatusCode, string(body))
```
MovieStoreGuy, line 259: The values of status code and body should be fields, the log line should also be capitalised.

internal/awslambda/extension/manager.go
```diff
@@ -214,6 +230,41 @@ func (m *manager) register(ctx context.Context) error {
...
+		return ErrFailedTelemetrySubscription
+	}
+
+	m.log.Info("successfully subscribed to telemetry API")
```
MovieStoreGuy, line 263:
```suggestion
	m.log.Info("Successfully subscribed to telemetry API")
```

Ensuring that log lines begin with an Upper case.

internal/awslambda/extension/manager_test.go
```diff
@@ -162,6 +162,21 @@ func EventNextHandler(tb testing.TB, ctx context.Context, statusCode, shutdownAf
...
+		require.Equal(tb, tb.Name(), r.Header.Get(api.LambdaExtensionIdentifierHeaderKey))
+		w.WriteHeader(statusCode)
+	})
+}
+
+func availableAddr() string {
+	l, _ := net.Listen("tcp", ":0")
+	defer l.Close()
+	return l.Addr().String()
+}
```
MovieStoreGuy, line 178: Please use `httptest.NewServer` instead of trying to figure out which port is free.

internal/awslambda/extension/telemetry/server.go
```diff
@@ -0,0 +1,97 @@
...
+
+	server := &http.Server{Addr: s.addr, Handler: mx}
+
+	go func() {
+		<-ctx.Done()
+		c, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+		defer cancel()
+		err := server.Shutdown(c)
+		if err != nil {
+			s.log.WithError(err).Info("did not shutdown gracefully")
```
MovieStoreGuy, line 62: Ensure the log line starts with an upper case.

internal/awslambda/extension/telemetry/server.go
```diff
@@ -0,0 +1,97 @@
...
+	return nil
+}
+
+func (s *Server) eventHandler(_ http.ResponseWriter, r *http.Request) {
+	b, err := io.ReadAll(r.Body)
+	if err != nil {
+		s.log.WithError(err).Error("error reading body")
+	}
+
+	var telePayload []Event
```
MovieStoreGuy, line 82: Should there be things being done with `telePayload`?

internal/awslambda/extension/telemetry/server.go
```diff
@@ -0,0 +1,97 @@
...
+type RuntimeDoneHook func()
+
+func NoopHook() RuntimeDoneHook {
+	return func() {}
+}
+
+type Server struct {
+	addr string
+	log  logrus.FieldLogger
+	f    RuntimeDoneHook
```
MovieStoreGuy, line 25: `context.CancelFunc` is reasonable to do here.

internal/awslambda/extension/telemetry/server.go
```diff
@@ -0,0 +1,97 @@
...
+)
+
+type RuntimeDoneHook func()
+
+func NoopHook() RuntimeDoneHook {
+	return func() {}
+}
+
+type Server struct {
+	addr string
```
MovieStoreGuy, line 23: Instead of caching the addr here, I think it might make more sense to store the server so you can get its listen addr once it has been started for testing.

internal/awslambda/extension/telemetry/server_test.go
```diff
@@ -0,0 +1,111 @@
...
+	assert.NotNil(t, NewServer(logrus.New(), NoopHook()))
+}
+
+func TestServerStart(t *testing.T) {
+	t.Parallel()
+
+	s := NewServer(logrus.New(), NoopHook(), WithCustomAddr("127.0.0.1:8083"))
+	ctx, cancel := context.WithCancel(context.Background())
+	go func() {
+		<-time.NewTimer(1 * time.Second).C
```
MovieStoreGuy, line 29: Use `time.After` instead.

internal/awslambda/extension/telemetry/server_test.go
```diff
@@ -0,0 +1,111 @@
...
+func TestNewServer(t *testing.T) {
+	t.Parallel()
+
+	assert.NotNil(t, NewServer(logrus.New(), NoopHook()))
+}
+
+func TestServerStart(t *testing.T) {
+	t.Parallel()
+
+	s := NewServer(logrus.New(), NoopHook(), WithCustomAddr("127.0.0.1:8083"))
```
MovieStoreGuy, line 26: Avoid using dedicated ports since if other tests or services on the machine running could clash causing a flake.

internal/awslambda/extension/telemetry/api.go
```diff
@@ -0,0 +1,40 @@
...
+
+	RuntimeDone EventType = "platform.runtimeDone"
+	// the sandbox hostname must be used in the lambda runtime, the port however can be configured if required
+	LambdaRuntimeAvailableAddr = "sandbox:8083"
+)
+
+type EventType string
+
+type Event struct {
+	// type is the only field we deserialise as it is used for flushing metrics
```
MovieStoreGuy, line 20: Out of curiosity, is there any other values we'd care about?

internal/awslambda/extension/manager.go
```diff
@@ -62,21 +66,33 @@ func NewManager(lambdaDomain string, lambdaFileName string, log logrus.FieldLogg
 		name:   lambdaFileName,
 	}
 
+	m.telemetryServer = telemetry.NewServer(log, func() {})
```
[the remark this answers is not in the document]
Excaleo: No, it's the setup for triggering a flush later, the server isn't that useful without some sort of hook we call on certain events. I'm starting with a noop hook.

internal/awslambda/extension/telemetry/api.go
Excaleo: The full catalogue of events and their schemas can be found here https://docs.aws.amazon.com/lambda/latest/dg/telemetry-schema-reference.html#platform-initStart

We don't care about anything else apart from when the runtime is done to trigger the flush, that only requires the type field

internal/awslambda/extension/telemetry/server.go
Excaleo: It's iterated through on line 88

internal/awslambda/extension/manager_test.go
Excaleo: Might be related to [this](https://github.com/atlassian/gostatsd/pull/652#discussion_r1446874217) , but I use this for the telemetry server since it takes an address and it's in a separate package so we don't have access to the underlying server

How should I structure it here?

internal/awslambda/extension/telemetry/server.go
Excaleo: I planned on having the whole lifecycle controlled via `Run(context Context)` , what's the additional value of storing the cancelFunc?

internal/awslambda/extension/telemetry/server.go
Excaleo: Something like ?
```
type Server struct {
    s *http.Server
}

func (s *Server) Addr() {
    return s.Addr()
}
```

I don't see much value in storing the server since nothing is called on it, but it could make it a bit more testable.

internal/awslambda/extension/telemetry/api.go
MovieStoreGuy: Easy, these might be useful in future to capture by default but we can save that for a future PR.

Excaleo: I can link it the schemas here for reference in case they we ever need more.

internal/awslambda/extension/manager_test.go
MovieStoreGuy: Yes, you're correct. let me follow up on that comment there to help with context.

internal/awslambda/extension/telemetry/server.go
MovieStoreGuy: If you wanted to test a full end to end, that storing the server provides the value that you can call the `server.Addr` and it will give you the bound address instead of the address it started with (ie `localhost:0`).

The other way to test this is expose the middleware or Mux and pass that to `httptest.NewServer(...)` and do the full end to end tests there.

Excaleo: Ah yep, I've changed it to server already, that makes sense, I've just been using the specific address and didn't know you could start with localhost:0

internal/awslambda/extension/manager_test.go
Excaleo: I've done the refactor for server but I still need an addr lol

I'll save the full end to end test for the next PR when flushes are added in.

internal/awslambda/extension/telemetry/server.go
```diff
@@ -0,0 +1,85 @@
...
+
+	for _, p := range telePayload {
+		if p.Type == RuntimeDone {
+			s.f()
+		}
+	}
+}
+
+func (s *Server) Endpoint() string {
+	return fmt.Sprintf("http://%s/telemetry", s.httpServer.Addr)
```
MovieStoreGuy, line 84: This will be wrong if you ever bound to `:0`, you may need to use `net.Listener` instead?

Perhaps wait til we get to the full End to End

MovieStoreGuy (approved):

PR #653, opened 2024-01-10
Backend: Adding initial OTLP types for new OTLP Backend

MovieStoreGuy:
## Context

We are seeing a growing increase to have data flowing through as a standardise set, so this work will enable a new backend to be written that will allow exporting of OTLP data. This work at the moment is allow helper functionality on the OTLP prototype values while keeping it performant.

There is a high level of usage of pointers that is borrowed from the collector's package, `pdata`. This is to allow a small struct to be assigned on the stack and copied instead of copying the entire the entire protobuf or having to deal with, "Should this be a pointer".

## Changes

- Adding in new backend directory named `otlp`
- Added new packages for OTLP
- Helper functionality for otlp package.


pkg/backends/otlp/internal/data/scope_metrics_test.go
```diff
@@ -0,0 +1,32 @@
...
+			},
+		},
+		empty,
+	)
+
+	sm := NewScopeMetrics(
+		NewInstrumentationScope("gostatsd/aggregation", "test"),
+		NewMetric("my-awesome-metric"),
+	)
+	assert.Len(t, sm.raw.Metrics, 1, "Must have one metric set")
```
Excaleo, line 31: Should we also assert its equality to the metric passed in on [L29](https://github.com/atlassian/gostatsd/pull/653/files#diff-5a74a3cd8cea5f055070e57ff0e99de842608140d97ce9477771fbc623d3964aR29)

pkg/backends/otlp/internal/data/histograms.go
```diff
@@ -0,0 +1,91 @@
...
+		ht.raw.DataPoints = append(
+			ht.raw.DataPoints,
+			datapoints[i].raw,
+		)
+	}
+
+	return ht
+}
+
+func WithHistogramDataPointTimeStamp(ts int64) func(HistogramDataPoint) {
```
Excaleo, line 37: Are timestamps optional for metrics? If not then we should force it in the constructor

pkg/backends/otlp/internal/data/metric_response_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+	for _, tc := range []struct {
+		name    string
+		body    io.Reader
+		dropped int64
+		errVal  string
+	}{
+		{
+			name:   "Valid response",
+			body:   bytes.NewBuffer(nil),
```
Excaleo, line 26: Let's make it explicit that we're dropping 0 here even though that is the 0 value. It makes it easier to read at a glance.

pkg/backends/otlp/internal/data/histograms.go
```diff
@@ -0,0 +1,91 @@
...
+func WithHistogramDataPointBucketValues[Buckets ~map[float64]uint64](buckets Buckets) func(HistogramDataPoint) {
+	return func(hdp HistogramDataPoint) {
+		bounds := maps.Keys(buckets)
+		slices.Sort(bounds)
+
+		hdp.raw.BucketCounts = make([]uint64, len(buckets))
+		hdp.raw.ExplicitBounds = make([]float64, len(buckets)-1)
+
+		for i, bound := range bounds {
+			hdp.raw.BucketCounts[i] = buckets[bound]
```
Excaleo, line 73: We will have a duplicate /redundant bucket in bucketcounts if `math.Inf` is in buckets, is this intended?

Excaleo (changes requested):

pkg/backends/otlp/internal/data/metric_request_test.go
```diff
@@ -0,0 +1,20 @@
...
+func TestMetricsRequest(t *testing.T) {
+	t.Parallel()
+
+	req, err := NewMetricsRequest(
+		context.Background(),
+		"not-a-valid-url",
+		NewResourceMetrics(NewResource()),
+	)
+	assert.NoError(t, err, "Must not error creating request")
+	assert.NotNil(t, req, "Must have a valid request")
```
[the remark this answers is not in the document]
MovieStoreGuy: What is your expectations with regards to checking the body? 

I wanted to avoid checking the body contents since protobuf adds extra metadata into unexposed values so it makes `assert.Equal` a pain to use.

pkg/backends/otlp/internal/data/metric_response.go
```diff
@@ -0,0 +1,33 @@
...
+	"io"
+	"net/http"
+
+	v1export "go.opentelemetry.io/proto/otlp/collector/metrics/v1"
+	"go.uber.org/multierr"
+	"google.golang.org/protobuf/proto"
+)
+
+func ProcessMetricResponse(resp *http.Response) (dropped int64, errs error) {
+	buf, err := io.ReadAll(resp.Body)
```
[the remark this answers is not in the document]
MovieStoreGuy: I've left it for now, since the response message unable to be passed would fail, and any errors would be pushed through proto field. 

Once I start the client and confer with the spec of the collector, I can update it then.

pkg/backends/otlp/internal/data/scope_metrics_test.go
MovieStoreGuy: I didn't have it in me to write out with the raw proto value that the content matched.

pkg/backends/otlp/internal/data/number_datapoint.go
```diff
@@ -0,0 +1,49 @@
...
+
+func WithNumberDataPointDoubleValue(value float64) func(NumberDataPoint) {
+	return func(ndp NumberDataPoint) {
+		ndp.raw.Value = &v1metrics.NumberDataPoint_AsDouble{
+			AsDouble: value,
+		}
+	}
+}
+
+func NewNumberDataPoint(opts ...func(NumberDataPoint)) NumberDataPoint {
```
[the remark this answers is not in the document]
MovieStoreGuy: I originally considered this, however, due to proto using private interfaces for `oneof` values, it meant that I had to some really ugly hacks that wouldn't have been performant. In this respect, last write wins is fine with me. 

Perhaps it should be documented?

pkg/backends/otlp/internal/data/metric_response_test.go
MovieStoreGuy: Easy

pkg/backends/otlp/internal/data/number_datapoint.go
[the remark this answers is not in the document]
MovieStoreGuy: This also happens with `Metric` btw, same case again with the `oneof` private interface.

pkg/backends/otlp/internal/data/metric_request_test.go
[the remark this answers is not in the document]
Excaleo: Just checking that the metrics that are attached to the post request as expected, at a minimum I reckon we should check the body has something

Excaleo (approved):
LGTM

PR #655, opened 2024-01-10
Backend: Adding OTLP Configuration

MovieStoreGuy:
## Context

This is the expected configuration of the OTLP backend once releases.

## Changes

- Added OTLP backend configuration

### DIFF
--- a/pkg/backends/otlp/config.go
+++ b/pkg/backends/otlp/config.go
@@ -1,0 +1,89 @@
+package otlp
+
+import (
+	"errors"
+	"runtime"
+
+	"github.com/spf13/viper"
+	"go.uber.org/multierr"
+)
+
+const (
+	ConversionAsGauge     = "AsGauge"
+	ConversionAsHistogram = "AsHistogram"
+)
+
+type Config struct {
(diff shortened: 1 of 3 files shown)


pkg/backends/otlp/config.go
```diff
@@ -0,0 +1,89 @@
...
+	// user agent header when making requests.
+	UserAgent string `mapstructure:"user_agent"`
+}
+
+func newDefaultConfig() Config {
+	return Config{
+		Transport:           "default",
+		MaxRequests:         runtime.NumCPU() * 2,
+		TimerConversion:     "AsGauge",
+		HistogramConversion: "AsGauge",
```
Excaleo, line 48: Should we type def AsGauge and AsHistograms since they're fixed constants?

Excaleo (approved):
Just a question on whether we should typedef, but not a blocker

pkg/backends/otlp/config.go
MovieStoreGuy: They will change into single value within the next PR, so I am not too worried

PR #656, opened 2024-01-11
Backend OTLP: Adding OTLP as a configurable backend for GoStatsD

MovieStoreGuy:
## Context

This adds in support for OTLP as a supported backend for Gostatsd however it does come with some limitations:

- Events are no (currently) supported
- Slight processing overhead converting from internal metrics values to OTLP values

Events will need to be exported as logs which currently isn't written yet and ignored for this PR. 

## Migration 

To ensure the highest amount of compatibility with other configurable backends, timers can either exported as Gauges with the calculated value added as suffix to the metric. This should ensure that existing metrics being emitted to your configured backend should maintain the same semantic value when visualising / evaluating within metrics platform.

## Outstanding

- [x] Adding tests for Backend 
- [x] Limiting the amount of concurrent requests being made
- [x] Respecting batching limits
- [x] Cleanup unused helper functionality


pkg/backends/otlp/internal/process/splitby_test.go
MovieStoreGuy, line 1: This will move to the top level folder for OTLP, not really needed being its own package

pkg/backends/otlp/backend.go
```diff
@@ -0,0 +1,237 @@
...
+		cb(multierr.Errors(err))
+	}
+}
+
+func (c *Backend) postMetrics(ctx context.Context, resourceMetrics []data.ResourceMetrics) error {
+	req, err := data.NewMetricsRequest(ctx, c.endpoint)
+	if err != nil {
+		return err
+	}
+	resp, err := c.client.Do(req)
```
Excaleo, line 230: IIRC we don't handle the status codes in ProcessMetricResponse so we should handle it here

pkg/backends/otlp/group.go
```diff
@@ -0,0 +1,34 @@
...
+		entry = data.NewResourceMetrics(
+			data.NewResource(
+				data.WithResourceMap(resources),
+			),
+		)
+	}
+	for _, m := range metrics {
+		entry.AppendMetric(is, m)
+	}
+	(*g)[key] = entry
```
Excaleo, line 33: Do we need a dereference here?

Excaleo (changes requested):

pkg/backends/otlp/internal/data/map.go
```diff
@@ -51,9 +51,18 @@ func (m Map) Equal(mm Map) bool {
...
+		sb.WriteString((*m.raw)[i].Key)
+		sb.WriteString((*m.raw)[i].Value.String())
+	}
+	return sb.String()
+}
+
 func (m Map) Insert(key string, value string) {
 	index := slices.IndexFunc(*m.raw, func(kv *v1common.KeyValue) bool {
-		return kv.Key == key
+		return kv.Key <= key
```
Excaleo, line 65: Are we guaranteeing that this is sorted with this logic? If so we should change the naming of the data struct to include that info.

pkg/backends/otlp/internal/data/map.go
```diff
@@ -70,22 +79,40 @@ func (m Map) Insert(key string, value string) {
...
+				return
+			}
+			t.ArrayValue.Values = append(t.ArrayValue.Values, &v1common.AnyValue{
+				Value: &v1common.AnyValue_StringValue{
+					StringValue: value,
+				},
+			})
 		}
-		v.ArrayValue.Values = append(v.ArrayValue.Values, val)
+		slices.Insert((*m.raw), index, &v1common.KeyValue{
```
Excaleo, line 112: I think we're inserting even if the value is appended to or set, I think this will lead to deplicate resources

pkg/backends/otlp/group.go
MovieStoreGuy: Yeah, I wanted to ensure we are referencing the same values and you can't index a pointer to a map so you need to dereference it.

MovieStoreGuy:
I've parked the additional actions, I would like to follow up on those to be done within the transport instead of within the encoding step.

pkg/backends/otlp/config.go
```diff
@@ -40,28 +42,30 @@ type Config struct {
...
-	return Config{
-		Transport:           "default",
-		MaxRequests:         runtime.NumCPU() * 2,
-		TimerConversion:     "AsGauge",
-		HistogramConversion: "AsGauge",
-		UserAgent:           "gostatsd",
+func newDefaultConfig() *Config {
+	return &Config{
+		Transport:   "default",
+		MaxRequests: runtime.NumCPU() * 2,
```
hstan, line 48: Is this too conservative?

MovieStoreGuy: So it really depends on the backend supporting out of order writes, if it does, you can open this up wide, otherwise, restricting it is a good idea.

hstan (approved):

PR #657, opened 2024-01-11
Lambda extension - support manual flush

Excaleo:
This PR adds manual flushes at the end of the execution of a lambda runtime. By default this is disabled, in the gostatsd and a `flush.Coordinator` has to be explicitly injected for it to be enabled. This maintains backwards compatibility with the existing `main.go` for gostatsd.

### Changes
- Added and integrated `flush.Coordinator` which can be used to trigger flushes to the gostatsd server in forwarder mode, disabled by default.
- Add full end to end integration test
### Remaining tasks
- [x] update documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,6 +161,29 @@ The following settings from the previous section are also supported:
 - `hostname`
 - `log-raw-metric`
 
+Running as a Lambda Extension
+-----------------------------
+Gostatsd can be run as a lambda extension in forwarder mode. The metrics are flushed at the end of each lambda invocation
+by default. The flush interval is ignored for your custom metrics, internal metrics are still flushed on a best effort
+basis using the configured flush interval.
+
+To support flushes based on the runtime function, a lambda telemetry server is started at the reserved lambda hostname
+`sandbox` on port 8083. This can be configured by setting the `lambda-extension-telemetry-address` configuration parameter.
+This will need to be done if port 8083 is not available within the lambda runtime.
+
+The flush per invocation setting can be disabled by setting `lambda-extension-manual-flush` to `false`, however this
+is not recommended unless the lambda is constantly invoked. Since the extensions are suspended once the user lambda
+functions return, this may lead to metric loss (for inflight requests) or metric delay until the next invocation in
+lambdas that are sparsely invoked.
+
+Configurable options:
+- `lambda-extension-telemetry-address`: address that the extension telemetry server should listen on
+- `lambda-extension-manual-flush`: boolean indicating whether the lambda should flush per invocation and disregard the
+   the flush interval
+
+All options for specified in the previous section for the forwarder are also configurable with the following caveats:
+- `dynamic-headers` are not supported
+- `flush-interval` will not be respected when `lambda-extension-manual-flush` is set to true
 
 Metric expiry and persistence
 -----------------------------
--- a/README.md
+++ b/README.md
@@ -161,7 +161,7 @@ The following settings from the previous section are also supported:
 - `hostname`
 - `log-raw-metric`
 
-Running as a Lambda Extension
+Running as a Lambda Extension (experimental feature)
 -----------------------------
 Gostatsd can be run as a lambda extension in forwarder mode. The metrics are flushed at the end of each lambda invocation
 by default. The flush interval is ignored for your custom metrics, internal metrics are still flushed on a best effort
--- a/cmd/lambda-extension/main.go
+++ b/cmd/lambda-extension/main.go
@@ -14,6 +14,7 @@ import (
 	"github.com/atlassian/gostatsd"
 	"github.com/atlassian/gostatsd/internal/awslambda/extension"
 	"github.com/atlassian/gostatsd/internal/awslambda/extension/api"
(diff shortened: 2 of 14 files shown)


README.md
```diff
@@ -161,6 +161,29 @@ The following settings from the previous section are also supported:
 - `hostname`
 - `log-raw-metric`
 
+Running as a Lambda Extension
+-----------------------------
+Gostatsd can be run as a lambda extension in forwarder mode. The metrics are flushed at the end of each lambda invocation
```
MovieStoreGuy, line 166: I would highlight that this is still considered experimental / alpha and will be GA in an upcoming TBD future releas.es

cmd/lambda-extension/main.go
```diff
@@ -119,20 +129,27 @@ func main() {
...
+	var opts = make([]extension.ManagerOpt, 0)
+	if conf.GetBool(gostatsd.ParamLambdaExtensionManualFlush) {
+		opts = append(opts, extension.WithManualFlushEnabled(server.ForwarderFlushCoordinator))
 	}
 
+	if addr := conf.GetString(gostatsd.ParamLambdaExtensionTelemetryAddress); addr != "" {
+		opts = append(opts, extension.WithCustomTelemetryServerAddr(addr))
+	}
+
+	manager := extension.NewManager(os.Getenv(api.EnvLambdaAPIHostname),
```
MovieStoreGuy, line 148: Move the `os.Getenv` to a new line to keep invocations the same

internal/flush/flush_coordinator.go
```diff
@@ -0,0 +1,55 @@
+package flush
+
+type Coordinator interface {
```
MovieStoreGuy, line 3: I would rename this to `FlushNotifier` instead of `Coordinator` since it does extend `Flushable`

internal/awslambda/extension/manager.go
```diff
@@ -46,31 +47,57 @@ type Server interface {
...
 
+func WithManualFlushEnabled(fc flush.Coordinator) ManagerOpt {
+	return func(m *manager) {
+		if fc == nil {
+			return
+		}
+
+		if m.telemetryServer == nil {
+			m.telemetryServer = telemetry.NewServer(telemetry.LambdaRuntimeAvailableAddr, m.log, fc.Flush)
+		}
```
MovieStoreGuy, line 89: This should only set the Coordinator not also set the Telemetry Server.

MovieStoreGuy: If these are strongly coupled, then I suggest setting the two together.

internal/awslambda/extension/manager_integration_test.go
```diff
@@ -0,0 +1,177 @@
...
+				cancel()
+			}
+		}
+	}))
+	apiMux.Handle("/v2/event", http.HandlerFunc(func(_ http.ResponseWriter, _ *http.Request) {}))
+
+	gsdServer := httptest.NewServer(apiMux)
+	t.Cleanup(gsdServer.Close)
+
+	log := logrus.New()
```
MovieStoreGuy, line 70: This is going to be rather noisey, is there way to provide it a test logger?

internal/awslambda/extension/manager_integration_test.go
```diff
@@ -0,0 +1,177 @@
...
+	res, err := sendRuntimeDoneMsg(teleAddr)
+	require.NoError(t, err)
+	require.Equal(t, http.StatusOK, res.StatusCode)
+
+	// wait for shutdown
+	wg.Wait()
+	assert.EqualValues(t, forwarderCallReceived, 2)
+}
+
+func createStatsDServer(apiAddr string,
```
MovieStoreGuy, line 107:
```suggestion
func createStatsDServer(
	apiAddr string,
```

MovieStoreGuy (commented):
Here is some initial feedback, I will review it in more indepth later today.

internal/flush/flush_coordinator.go
Excaleo: The only problem is there's already a something called `flushNotifier` in the repo, I made it a coordinator because you can also invoke a flush manually

The other flushNotifier interface is strictly for notification of flushes

internal/awslambda/extension/manager_integration_test.go
Excaleo: Setting the output to `io.Discard`

MovieStoreGuy (dismissed):

cmd/lambda-extension/main.go
```diff
@@ -119,20 +130,25 @@ func main() {
...
 
-	// TODO: Configure log group
-
 	log.Info("Starting extension runtime")
 
 	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt)
 	defer cancel()
 
-	manager := extension.NewManager(os.Getenv(api.EnvLambdaAPIHostname), conf.GetString(ParamLambdaFileName), log)
+	server := CreateServer(conf, log)
```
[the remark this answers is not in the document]
Excaleo: Fixed

MovieStoreGuy (approved):

PR #662, opened 2024-02-15
Add FIPS only

irisgve:


MovieStoreGuy (approved):

MovieStoreGuy:
When you got to make the release, would also be good to call out that we've defaulted to FIPs only support

PR #680, opened 2024-05-21
Fix viper environment variable reading

tiedotguy:
Some sort of breaking change from viper 1.15 to 1.16, because viper isn't semver or something I don't know.


hligit (approved):
Thanks!

MovieStoreGuy (approved):

PR #681, opened 2024-05-22
Improving Lambda Extension API

MovieStoreGuy:
## Context

This removes inverse dep injection on start to make it more a concrete API that aligns with the existing server interface.

This will help when creating a "runtime" that can be included into a different project.

## Changes

- Simplified Lambda Extension API
- Updated tests with impacted changes.


irisgve (approved):

MovieStoreGuy:
Skipping changelog due to changes being internal and non user facing.

PR #682, opened 2024-05-22
Feature: Expose Lambda Extension package

MovieStoreGuy:
## Context

In order to allow the lambda extension work to be included in other external projects, it isn't completely clean (but gives a minimal required code changes). 

## Changes

I wanted to get some feedback before I commit to finishing it, the checked parts are done, the unchecked are what I'd like to do if we are okay with the direction.

- [x] Adds new package `pkg/lambda` that exposes a small subset of functionality
- [x] Added tests
- [ ] Replaces invocation in `cmd/lambda-extension`

### DIFF
--- a/pkg/lambda/options.go
+++ b/pkg/lambda/options.go
@@ -1,0 +1,91 @@
+package lambda
+
+import (
+	"errors"
+	"os"
+	"path"
+
+	"github.com/spf13/pflag"
+	"go.uber.org/multierr"
+
+	"github.com/atlassian/gostatsd/internal/awslambda/extension/api"
+)
+
+const (
+	ParamRuntimeAPI        = `lambda-runtime-api`
+	ParamEntryPointName    = `lamda-entrypoint-name`
+	ParamEnableManualFlush = `enable-manual-flush`
+	ParamTelemetryAddr     = `telemetry-addr`
+)
+
+// Options is used to provide overrides when calling `NewExtension`
+type Options struct {
+	// RuntimeAPI is AWS service that is responsible for orchestrating
+	// the extension with the other Lambda's that are part of the deployment.
+	// This will set to `env:AWS_LAMBDA_RUNTIME_API` by default as per the docs,
+	// but can be overriden for testing purposes and validation.
+	RuntimeAPI string
+	// ExecutableName is the full file name of the lambda extension that is
+	// used to validate the bootstrap process.
+	ExecutableName string
+	// EnableManualFlush allows for Lambda's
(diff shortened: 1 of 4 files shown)


hstan (approved):

pkg/lambda/options.go
```diff
@@ -0,0 +1,105 @@
...
+	ParamRuntimeAPI        = `lambda-runtime-api`
+	ParamEntryPointName    = `lamda-entrypoint-name`
+	ParamEnableManualFlush = `enable-manual-flush`
+	ParamTelemetryAddr     = `telemetry-addr`
+)
+
+// Options is used to provide overrides when calling `NewExtension`
+type Options struct {
+	// Server is extended by the extension
+	// This value is required, no defaults are provided.
```
tiedotguy, line 26: If things are required, they should be passed as parameters.  The struct should be optional stuff only, if it's intention is to only be overridden sometimes.

pkg/lambda/options.go
```diff
@@ -0,0 +1,105 @@
...
+	ExecutableName string
+	// EnableManualFlush allows for Lambda's
+	// to control when the forwarder would send the accumlated metrics.
+	EnableManualFlush bool
+	// TelemetryAddr is used by the AWS runtime to publish
+	// events back to service.
+	TelemetryAddr string
+}
+
+func NewDefaultOptions() Options {
```
tiedotguy, line 47: Can we call this `NewOptionsFromEnvironment`?  It makes it more explicit this isn't purely defaults, and that it's not a pure function generally.

pkg/lambda/options.go
```diff
@@ -0,0 +1,105 @@
...
+	return Options{
+		RuntimeAPI:        os.Getenv(api.EnvLambdaAPIHostname),
+		ExecutableName:    path.Base(os.Args[0]),
+		EnableManualFlush: false,
+		TelemetryAddr:     ":8083",
```
tiedotguy, line 52: Is this normal to be on `:8083`? Are we going to have any conflicts with expvar?

tiedotguy (changes requested):

pkg/lambda/options.go
MovieStoreGuy: That is a good question, I am actually not sure what it is expected to be by the AWS library.

tiedotguy (dismissed):
lgtm

MovieStoreGuy:
I'll add the replace in `cmd/lambda-extension` in a future PR

tiedotguy (approved):

PR #694, opened 2024-05-30
Support Events in OTLP backend

hstan:
This PR adds support for events in the OTLP backend. 

There is no native OTLP events support, so it is done via converting events to log data then sending to an OTLP receiver.


pkg/backends/otlp/internal/data/event.go
```diff
@@ -0,0 +1,125 @@
...
+func WithPropertiesAttrKey(key string) func(*OtlpEvent) {
+	return func(e *OtlpEvent) {
+		e.propertiesAttrKey = key
+	}
+}
+
+func (s *OtlpEvent) TransformToLog() *v1log.LogRecord {
+	attrs := eventLogAttributes(make([]*eventLogAttribute, 0))
+	e := s.raw
+	dimensions := e.Tags.ToMap(string(e.Source))
```
irisgve, line 47: Hmm why don't we just straight up put `source` in attrs rather than putting it to `tags` then `dimensions` then `attrs`?

hstan: good question, the logic was the source is converted into `host` in `toMap`.  It should be keep as it is and be converted as needed in the downstream services.

pkg/backends/otlp/internal/data/event.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	return &OtlpEvent{raw: e}, nil
+}
+
+func (s *OtlpEvent) TransformToLog() *v1log.LogRecord {
+	attrs := make(map[string]string)
+	e := s.raw
+
+	dimensions := e.Tags.ToMap()
+	for key, value := range dimensions {
```
irisgve, line 29: Should these still be inside a map called tags i.e. so the gateway processor can distinguish these dimensions from those below as a tags?

pkg/backends/otlp/internal/data/event.go
```diff
@@ -0,0 +1,64 @@
...
+		attrs[key] = value
+	}
+
+	attrs["title"] = e.Title
+	attrs["text"] = e.Text
+	attrs["source"] = string(e.Source)
+	attrs["priority"] = e.Priority.String()
+	attrs["alert_type"] = e.AlertType.String()
```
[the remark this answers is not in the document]
hstan: yeah they should be included too, to keep it as it is.
They were not added initially as our particular downstream doesn't need them..

pkg/backends/otlp/internal/data/event.go
hstan: Currently I'm flattening it and override the tags with fields on the events. As tested here https://github.com/atlassian/gostatsd/pull/694/files#diff-8f2c078c61ccc41bce7f801d20bd30203625ca465ca38f0ea9c1afed0d3df7b0R66

I think you're right, we should let downstream decide how to deal with it.

pkg/backends/otlp/internal/data/event_response.go
```diff
@@ -0,0 +1,34 @@
...
+	}
+
+	var response v1export.ExportLogsServiceResponse
+	if err := proto.Unmarshal(buf, &response); err != nil {
+		errs = multierr.Append(errs, fmt.Errorf("failed to unmarshal response: %w", err))
+		return errs
+	}
+
+	if ps := response.PartialSuccess; ps != nil && ps.ErrorMessage != "" {
+		errs = multierr.Append(errs, fmt.Errorf("failed to send events: %s", ps.ErrorMessage))
```
irisgve, line 30: Do we also want to record here the # of rejected log records similar to datapoints?

hstan: I removed it.. 'cause unlike data points, each request only send one logRecord. So there can't be more than 1 dropped events.

pkg/backends/otlp/backend.go
```diff
@@ -66,17 +69,45 @@ func NewClientFromViper(v *viper.Viper, logger logrus.FieldLogger, pool *transpo
...
+	if err != nil {
+		atomic.AddUint64(&b.droppedEvents, 1)
+		return err
+	}
+
+	b.requestsBufferSem <- struct{}{}
+	resp, err := b.client.Do(req)
+	<-b.requestsBufferSem
+	if err != nil {
+		atomic.AddUint64(&b.droppedEvents, 1)
```
irisgve, line 100: If this returns here how does statser know to add it in `backend.dropped_events`

hstan: same thing for the metrics

irisgve: :kek:

irisgve (approved):

pkg/backends/otlp/backend.go
irisgve: I'll raise a github issue :hehehe:

PR #699, opened 2024-06-11
Add endpoint configuration for logs in otlp backend

waynewu411:
1. Rename "endpoint" to "metrics_endpoint"
2. Add logs_endpoint

### DIFF
--- a/pkg/backends/otlp/backend.go
+++ b/pkg/backends/otlp/backend.go
@@ -32,7 +32,7 @@ type Backend struct {
 	droppedMetrics uint64
 	droppedEvents  uint64
 
-	endpoint              string
+	metricsEndpoint       string
 	logsEndpoint          string
 	convertTimersToGauges bool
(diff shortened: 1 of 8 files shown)


pkg/backends/otlp/backend.go
```diff
@@ -252,7 +254,7 @@ func (bd *Backend) SendMetricsAsync(ctx context.Context, mm *gostatsd.MetricMap,
 	err := bd.postMetrics(ctx, group.Values())
 	if err != nil {
 		bd.logger.WithError(err).WithFields(logrus.Fields{
-			"endpoint": bd.endpoint,
```
hstan, line 255: we can keep the field as "endpoint" to keep it consistent

waynewu411: Hi @hstan, this has been reverted.

hstan (approved):

PR #713, opened 2024-07-30
Support batch sending metrics for OTLP backend

hstan:
This pull request introduces a batching mechanism for metrics in the OTLP backend's flushing process. Previously, metrics were not batched, causing the size of requests to grow unbounded with increasing flush intervals and the number of metrics handled. This change ensures that metrics are batched according to a configurable batch size, with a default of 1000 metrics per batch.

### DIFF
--- a/pkg/backends/otlp/group.go
+++ b/pkg/backends/otlp/group.go
@@ -15,7 +15,7 @@ func (g *group) values() []data.ResourceMetrics {
 func (g *group) lenMetrics() int {
 	count := 0
 	for _, rm := range *g {
-		count += rm.CouneMetrics()
+		count += rm.CountMetrics()
 	}
 	return count
 }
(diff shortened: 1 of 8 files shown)


pkg/backends/otlp/group.go
```diff
@@ -6,31 +6,54 @@ import (
...
+type group map[uint64]data.ResourceMetrics
+
+func (g *group) values() []data.ResourceMetrics {
+	return maps.Values(*g)
+}
+
+func (g *group) lenMetrics() int {
+	count := 0
+	for _, rm := range *g {
+		count += rm.CouneMetrics()
```
MovieStoreGuy, line 18: Looks like a typo here for the method name?

MovieStoreGuy (approved):

waynewu411 (approved):

MovieStoreGuy (approved):

waynewu411 (approved):

PR #720, opened 2024-08-06
send batches concurrently in each flush and compress payload

hstan:
Send batches concurrently in each flush and compress payload.

This should shorten process time for each flush and also reduce the size of payload.

### DIFF
--- a/pkg/backends/otlp/backend.go
+++ b/pkg/backends/otlp/backend.go
@@ -13,6 +13,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/viper"
 	"go.uber.org/multierr"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/atlassian/gostatsd/pkg/stats"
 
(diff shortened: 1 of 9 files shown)


pkg/backends/otlp/internal/data/metric_request.go
```diff
@@ -1,6 +1,8 @@
 package data
 
 import (
+	"bytes"
+	"compress/gzip"
```
MovieStoreGuy, line 5: I think this will be fine for now, but we will need to look at adopting lz4 compression to help improve performance.

hstan: idk if there's choice for now as the OTLP doc is quite confusing,  [specification here](https://opentelemetry.io/docs/specs/otlp/#protocol-details) says it only need to support `gzip` to meet the minimal requirement for a OTLP server

waynewu411 (approved):

pkg/backends/otlp/backend.go
```diff
@@ -280,27 +283,32 @@ func (bd *Backend) SendMetricsAsync(ctx context.Context, mm *gostatsd.MetricMap,
...
-			bd.logger.WithError(err).WithFields(logrus.Fields{
-				"endpoint": bd.metricsEndpoint,
-			}).Error("Issues trying to submit data")
-			errs = multierr.Append(errs, err)
-		} else {
-			atomic.AddUint64(&bd.batchesSent, 1)
-			atomic.AddUint64(&bd.seriesSent, uint64(b.lenMetrics()))
-		}
+		func(g group) {
+			eg.Go(func() error {
```
irisgve, line 296: Cool this means async sends? Do we have any retries in the otlp backend?

irisgve (approved):

pkg/backends/otlp/backend.go
hstan: yeah retry is in the `postMetrics`

PR #730, opened 2024-11-06
Migrate to aws sdk 2 for cloudwatch/aws cloud provider 

akavatl:


pkg/backends/cloudwatch/cloudwatch.go
```diff
@@ -57,25 +60,23 @@ func NewClient(namespace, transport string, disabled gostatsd.TimerSubtypes, log
 	if err != nil {
 		return nil, err
 	}
-	sess, err := session.NewSession(&aws.Config{
-		HTTPClient: httpClient.Client,
-	})
+	cfg, err := config.LoadDefaultConfig(context.TODO(), config.WithHTTPClient(httpClient.Client))
```
irisgve, line 63: Rather than `context.TODO` which we typically use when we're unclear what context to use and intend to update in the future, we can also do `context.Background` which is typically for empty root contexts with no values or cancellation

pkg/backends/cloudwatch/cloudwatch.go
```diff
@@ -26,11 +25,15 @@ const MAX_DIMENSIONS = 10
 // BackendName is the name of this backend.
 const BackendName = "cloudwatch"
 
+type ApiClient interface {
```
irisgve, line 28: nit: Do we want that to be more specific than ApiClient i.e. CloudwatchClient?

pkg/backends/cloudwatch/cloudwatch_test.go
```diff
@@ -240,15 +240,15 @@ func TestSendHistogram(t *testing.T) {
 	}
 }
 
-func findMetricDatum(input *cloudwatch.PutMetricDataInput, name string, dimenensionName string, dimensionValue string) *cloudwatch.MetricDatum {
+func findMetricDatum(input *cloudwatch.PutMetricDataInput, name string, dimensionName string, dimensionValue string) *types.MetricDatum {
```
irisgve, line 243: Nice catch lol

pkg/cloudproviders/aws/aws.go
```diff
@@ -232,30 +241,35 @@ func NewProviderFromViper(v *viper.Viper, logger logrus.FieldLogger, _ string) (
...
-		}).
-		WithMaxRetries(a.GetInt("max_retries"))
-	metadataSession, err := session.NewSession(sharedConfig)
-	if err != nil {
-		return nil, fmt.Errorf("error creating a new Metadata session: %v", err)
-	}
-	metadata := ec2metadata.New(metadataSession)
-	region, err := metadata.Region()
+		}),
+		config.WithRetryMaxAttempts(5),
```
irisgve, line 250: Should we be setting that from max_retries config?

akavatl: Yes we should, I thought I had updated that, weird

pkg/backends/cloudwatch/cloudwatch.go
akavatl: yes we can

pkg/backends/cloudwatch/cloudwatch.go
akavatl: Thanks, I was going off what the migration guide had but that is better

pkg/backends/cloudwatch/cloudwatch.go
akavatl: My ide does give out with types starting with cloudwatch since it's the package name

pkg/cloudproviders/aws/aws.go
```diff
@@ -74,28 +76,45 @@ func (p *Provider) RunMetrics(ctx context.Context, statser stats.Statser) {
...
 	atomic.AddUint64(&p.describeInstanceInstances, uint64(len(IP)))
 	instancesFound := uint64(0)
 	pages := uint64(0)
+	var err error
+	var nextToken string
 
 	p.logger.WithField("ips", IP).Debug("Looking up instances")
-	err := p.Ec2.DescribeInstancesPagesWithContext(ctx, input, func(page *ec2.DescribeInstancesOutput, lastPage bool) bool {
+	hasPagesRemaining := true
+	for hasPagesRemaining {
```
irisgve, line 101: I wonder can we use `NewDescribeInstancesPaginator` for this?

pkg/cloudproviders/aws/aws.go
```diff
@@ -74,28 +76,45 @@ func (p *Provider) RunMetrics(ctx context.Context, statser stats.Statser) {
...
 		pages++
```
hstan, line 102: I know this is unchanged but just wondering if it should count a page if the describe failed later

pkg/cloudproviders/aws/aws.go
akavatl: We probably can, this is an instance of the migrated functions equivalent not being obvious for the writer

pkg/cloudproviders/aws/aws.go
akavatl: Fair question, don't have a strong opinion tbh

pkg/cloudproviders/aws/aws.go
```diff
@@ -74,28 +76,45 @@ func (p *Provider) RunMetrics(ctx context.Context, statser stats.Statser) {
...
+		nextToken = *page.NextToken
+		if rawErr != nil {
+			atomic.AddUint64(&p.describeInstanceErrors, 1)
+			hasPagesRemaining = false
+
+			if rawErr.Error() != "InvalidInstanceID.NotFound" {
+				err = fmt.Errorf("error listing AWS instances: %v", rawErr)
+			}
+		}
 		for _, reservation := range page.Reservations {
```
hstan, line 118: if `rawErr != nil`,  `page` will be `nil` right? We should probably break the loop above

hstan: https://github.com/aws/aws-sdk-go-v2/blob/main/service/ec2/api_op_DescribeInstances.go#L54
^ what I mean is here, the `page` is nil, and it'll panic when calling `page.Reservations`

hstan: > follow up question there is, should we end the pagination if the expected error from eventual consistency surfaces
I thought it intends to end because `hasPagesRemaining` is set to `false`
But I don't quite get the business logic 🤔

akavatl: Ah then we should definitely break

pkg/cloudproviders/aws/aws.go
irisgve: Nice saw the changes

irisgve:
@akavatl should we move to ready for review now?

akavatl:
@irisgve Yes I will, I want to do some deeper testing  but it shouldn't block this being reviewed(just being merged)

irisgve (approved):

irisgve (dismissed):

Makefile
```diff
@@ -66,14 +66,14 @@ build-gostatsd-fips:
 build-lambda-fips:
 	@$(MAKE) build PKG=$(LAMBDA_PKG) BINARY_NAME="lambda-extension" GOBUILD_OPTIONAL_FLAGS="-tags fips"
 
-build-all: pb/gostatsd.pb.go tools
```
irisgve, line 69: This should be fine to leave here since `build-all` covers building the proto

irisgve (approved):

irisgve (approved):

PR #745, opened 2025-03-18
Fix race condition when we init lambda

tiedotguy:
There's a couple of problems here:

1. We subscribe a telemetry endpoint before the service is ready.  If we receive a flush to that endpoint, then we try to perform a flush and get an NPE.

2. The access to the flusher is unprotected across multiple goroutines.  We just use an RWMutex for this, because it's rarely accessed and doesn't need anything fancy.

3. There's a race condition in the test because the nop write is (was) async.


tiedotguy:
Dangit wrong account

PR #746, opened 2025-03-18
Fix race condition when we init lambda

tiedotguy:
There's a couple of problems here:

1. We subscribe a telemetry endpoint before the service is ready.  If we receive a flush to that endpoint, then we try to perform a flush and get an NPE.

2. The access to the flusher is unprotected across multiple goroutines.  We just use an RWMutex for this, because it's rarely accessed and doesn't need anything fancy.

3. There's a race condition in the test because the nop write is (was) async.

### DIFF
--- a/pkg/statsd/handler_http_forwarder_v2.go
+++ b/pkg/statsd/handler_http_forwarder_v2.go
@@ -308,7 +308,7 @@ func (hfh *HttpForwarderHandlerV2) sendNop(ctx context.Context) {
 
 func (hfh *HttpForwarderHandlerV2) Run(ctx context.Context) {
 	var wg wait.Group
-	wg.StartWithContext(ctx, hfh.sendNop)
+	hfh.sendNop(ctx)
 	wg.Start(func() {
 		for metricMaps := range hfh.consolidatedMetrics {
 			hfh.acquireMergingSem()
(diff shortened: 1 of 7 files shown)


pkg/web/http_receiver_v2_test.go
```diff
@@ -45,7 +45,7 @@ func TestForwardingEndToEndV2(t *testing.T) {
 			ctx = clock.Context(ctx, mockClock)
 
 			ch := &channeledHandler{
-				chMaps: make(chan *gostatsd.MetricMap),
+				chMaps: make(chan *gostatsd.MetricMap, 1),
```
tiedotguy, line 48: We need this because the `sendNop` is now sync.  If that doesn't go in to the channel, `HttpForwarderHandlerV2.Run` never starts using the virtual clock, and `fixtures.NextStep()` waits forever (because it's waiting for something to use the clock)

pkg/statsd/handler_http_forwarder_v2.go
```diff
@@ -308,7 +308,7 @@ func (hfh *HttpForwarderHandlerV2) sendNop(ctx context.Context) {
 
 func (hfh *HttpForwarderHandlerV2) Run(ctx context.Context) {
 	var wg wait.Group
-	wg.StartWithContext(ctx, hfh.sendNop)
+	hfh.sendNop(ctx)
```
irisgve, line 311: Just for my context, why do we need to send a nop before we do actual flushes?

tiedotguy: The deepcheck checks if we've ever sent metrics - if we can't send metrics, we don't want to start.

We don't check how longs it's been though, since this is about misconfiguration, not backend going down.

tiedotguy:
Actually, I guess I should include a changelog.

irisgve (approved):

PR #750, opened 2025-04-04
Emit latency metric for forwarder requests

alyssat2:
Added changes to emit latency metrics for forwarded batches of metrics.

Two new latency metrics are added:
- `http.forwarder.post_latency.sum`: the total of the time taken to forward a batch in the flush interval. This can be used to find the average latency of forwarder requests by dividing by the number of messages sent.
- `http.forwarder.post_latency.max`: the maximum time taken to forward a batch in the flush interval.

### DIFF
--- a/METRICS.md
+++ b/METRICS.md
@@ -60,8 +60,8 @@ Metrics:
 | http.forwarder.sent                         | counter               |                              | The number of batches successfully forwarded
 | http.forwarder.retried                      | counter               |                              | The number of retries sending a batch
 | http.forwarder.dropped                      | counter               |                              | The number of batches dropped due to inability to forward upstream
-| http.forwarder.post_latency.avg             | gauge (time)          |                              | The average time taken to forward a batch in the flush interval
-| http.forwarder.post_latency.max             | gauge (time)          |                              | The maximum time taken to forward a batch in the flush interval
+| http.forwarder.post_latency.sum             | counter               |                              | The total of the time taken to forward a batch in the flush interval
+| http.forwarder.post_latency.max             | gauge (flush)         |                              | The maximum time taken to forward a batch in the flush interval
 | http.incoming                               | counter               | server-name, result, failure | The number of batches forwarded to the server, and the results of processing them
 | http.incoming.metrics                       | counter               | server-name                  | The number of metrics received over http
   


hstan (dismissed):
LGTM

pkg/statsd/handler_http_forwarder_v2.go
```diff
@@ -465,10 +466,14 @@ func (hfh *HttpForwarderHandlerV2) post(ctx context.Context, message proto.Messa
 	b.MaxElapsedTime = hfh.maxRequestElapsedTime
 
 	for {
+		startTime := clock.Now(ctx)
 		if err = post(); err == nil {
 			atomic.AddUint64(&hfh.messagesSent, 1)
 			hfh.lastSuccessfulSend.Store(clock.Now(ctx).UnixNano())
+			statser.TimingDuration("http.forwarder.post_duration", clock.Since(ctx, startTime), gostatsd.Tags{"status:success"})
```
tiedotguy, line 473: These are per-host timers, it will be a huge cardinality explosion.

For things like this we accumulate data in a variable (make sure to be thread safe), and emit a single value for what we actually want when we flush.  See the `emitMetrics` function, and that data it consumes.

hstan: Right, I didn't consider that the internalStatser added a host tag.
But the metrics reported from `emitMetrics` are counters and this is latency, should we aggregate it?

irisgve: For the cardinality explosion, I reckon we can aggregate these on our side and get host dimension nuked out.

tiedotguy: I've got 3 primary concerns here:

1. I have always tried to keep "Atlassianisms" out of gostatsd (there's some in cmd/tester though!).  Saying "Atlassian can use our metrics sink to deal with a non-Atlassian problem" goes against DFTC.  We should avoid making changes to gostatsd which are useful for all, not for us.
2. Sending data through the generic aggregation pipeline burns CPU.  Stripping data out later also burns CPU, all along the pipeline.  I've always tried to make gostatsd efficient, and not do unnecessary work.  An atomic increment vs sending multiple values through a pipeline to eventually be sorted and have numerous calculations performed on it.
3. With **all** my projects, I try to emit telemetry to answer a specific question.  Here, the question seems under-specified: "emit latency".  But what is that?

- "What is the latency" is a question that requires logging every value, because it's a question without aggregation.
- "What is the max latency" can be generated with a timer, or tracking it in a max variable we flush.  The former burns CPU, the latter gives us the ability to decide what we actually want, and only what we want.
- "What is the average latency" can be generated with a timer, or tracking a sum/count which we flush (as two values - we might already have a count though).  Again, the former burns CPU, the latter gives us flexibility.  We can look at long term averages by doing the math on the backend if we want.

pkg/statsd/handler_http_forwarder_v2.go
```diff
@@ -299,6 +302,8 @@ func (hfh *HttpForwarderHandlerV2) emitMetrics(statser stats.Statser) {
 	statser.Report("http.forwarder.sent", &hfh.messagesSent, nil)
 	statser.Report("http.forwarder.retried", &hfh.messagesRetried, nil)
 	statser.Report("http.forwarder.dropped", &hfh.messagesDropped, nil)
+	statser.TimingMS("http.forwarder.post_latency.max", float64(hfh.postLatencyMax.Load()), nil)
+	statser.TimingMS("http.forwarder.post_latency.avg", float64(hfh.postLatencyTotal)/float64(hfh.messagesSent), nil)
```
hstan, line 306: * will `messageSent` ever be 0?
* `hfh.postLatencyTotal` never gets reset to 0, so the result will be incorrect.

pkg/statsd/handler_http_forwarder_v2.go
```diff
@@ -299,6 +302,8 @@ func (hfh *HttpForwarderHandlerV2) emitMetrics(statser stats.Statser) {
 	statser.Report("http.forwarder.sent", &hfh.messagesSent, nil)
 	statser.Report("http.forwarder.retried", &hfh.messagesRetried, nil)
 	statser.Report("http.forwarder.dropped", &hfh.messagesDropped, nil)
+	statser.TimingMS("http.forwarder.post_latency.max", float64(hfh.postLatencyMax.Load()), nil)
```
tiedotguy, line 305: This is a tricky decision on what to emit here.  There's going to be several values sent in the aggregator flush interval (remember the forwarder flushes a lot faster than the aggregator).

Right now, we're getting the lifetime maximum, since we don't reset the max.  That may be acceptable?

If we reset the max:
- If we emit a timer, then we're back in the same boat of "all the cardinality".
- If we emit a gauge, then only the last forwarder flush interval will be recorded.

If we don't reset the max:
- If we emit a timer, then we're back in the same boat of "all the cardinality".
- If we emit a gauge, then we get the process lifetime maximum (which may be fine)

pkg/statsd/handler_http_forwarder_v2.go
tiedotguy, line 306: `hfh.messagesSent` is atomic - it was written before the atomic packages, hence the comment about `atomic` in the struct.  As such, you must use `atomic.LoadInt64()`.

However... `Report` won't just report, if we're in forwarder mode, it will also reset to 0:
```
func (is *InternalStatser) Report(name string, value *uint64, tags gostatsd.Tags) {
	if is.forwarderMode {
		val := float64(atomic.SwapUint64(value, 0))  // <-- swap
```

This is a case of side-effects causing problems.  We can restructure this like:

```
func (hfh *HttpForwarderHandlerV2) emitMetrics(statser stats.Statser) {
	messagesInvalid := float64(atomic.SwapUint64(&hfh.messagesInvalid, 0))
	messagesCreated := float64(atomic.SwapUint64(&hfh.messagesCreated, 0))
	messagesSent := float64(atomic.SwapUint64(&hfh.messagesSent, 0))
	messagesRetried := float64(atomic.SwapUint64(&hfh.messagesRetried, 0))
	messagesDropped := float64(atomic.SwapUint64(&hfh.messagesDropped, 0))

	statser.Count("http.forwarder.invalid", messagesInvalid, nil)
	statser.Count("http.forwarder.created", messagesCreated, nil)
	statser.Count("http.forwarder.sent", messagesSent, nil)
	statser.Count("http.forwarder.retried", messagesRetried, nil)
	statser.Count("http.forwarder.dropped", messagesDropped, nil)
}
```

Note that we can only do this behavior here because we know we're in a forwarder.  If we were somewhere outside of a forwarder, we would have to know if we should use swap or load.

Now that we have our data in a variable, we can then do:
```
	postLatencyTotal := float64(atomic.SwapUint64(&hfh.postLatencyTotal, 0))

	statser.???("http.forwarder.post_latency.avg", hfh.postLatencyTotal/messagesSent, nil)
```
(See previous comment about which metric type to use)

However... averages aren't super helpful, it's more useful to emit a counter:
```
	statser.Count("http.forwarder.post_latency.sum", float64(hfh.postLatencyTotal), nil)
```

What this means is on the backend we can query:
```
Sum = data('http.forwarder.post_latency.sum.count', rollup='sum')
Total = data('http.forwarder.sent.count', rollup='sum')
(Sum/Total).publish('Average latency')
```

This lets us see the average of any time frame, because we can perform a sum across both the latency and the sent, and get meaningful data.

pkg/statsd/handler_http_forwarder_v2.go
```diff
@@ -44,12 +44,15 @@ const (
...
-	messagesSent       uint64 // atomic - messages successfully sent
-	messagesRetried    uint64 // atomic - retries (first send is not a retry, final failure is not a retry)
-	messagesDropped    uint64 // atomic - final failure
+	postId           uint64       // atomic - used for an id in logs
+	messagesInvalid  uint64       // atomic - messages which failed to be created
+	messagesCreated  uint64       // atomic - messages which were created
+	messagesSent     uint64       // atomic - messages successfully sent
+	messagesRetried  uint64       // atomic - retries (first send is not a retry, final failure is not a retry)
+	messagesDropped  uint64       // atomic - final failure
+	postLatencyTotal int64        // atomic - total of the time taken to send messages in a flush interval
```
tiedotguy, line 53: We should make this `atomic.Int64` as well.  This code pre-dates the `atomic` package, hence why we don't have it.

hstan (approved):
@tiedotguy could you take a look as well?

pkg/statsd/handler_http_forwarder_v2.go
```diff
@@ -299,6 +302,11 @@ func (hfh *HttpForwarderHandlerV2) emitMetrics(statser stats.Statser) {
 	statser.Report("http.forwarder.sent", &hfh.messagesSent, nil)
 	statser.Report("http.forwarder.retried", &hfh.messagesRetried, nil)
 	statser.Report("http.forwarder.dropped", &hfh.messagesDropped, nil)
+
+	postLatencyMax := hfh.postLatencyMax.Swap(0)
+	postLatencyTotal := hfh.postLatencyTotal.Swap(0)
+	statser.Gauge("http.forwarder.post_latency.max", float64(postLatencyMax), nil)
+	statser.Count("http.forwarder.post_latency.sum", float64(postLatencyTotal), nil)
```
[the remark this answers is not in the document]
tiedotguy: The `Report` function takes a `*uint64` and uses atomic function, as it was written before we had the `atomic` package.  Since the `atomic` package is better (less error prone due to actual types), it's better to use that.  There's an argument to say we should have `Report` take an `atomic.Uint64`, and refactor everything for it, but...

The `Report` function isn't great - it sounds like it reports a metric, but it doesn't, it also does a reset, and that's not at all clear.  I would advocate to remove it entirely, if we were doing a refactor.

tiedotguy (approved):

pkg/statsd/handler_http_forwarder_v2.go
```diff
@@ -465,9 +473,18 @@ func (hfh *HttpForwarderHandlerV2) post(ctx context.Context, message proto.Messa
...
 
 	for {
+		startTime := clock.Now(ctx)
 		if err = post(); err == nil {
 			atomic.AddUint64(&hfh.messagesSent, 1)
 			hfh.lastSuccessfulSend.Store(clock.Now(ctx).UnixNano())
+
+			postLatency := clock.Since(ctx, startTime).Milliseconds()
+			hfh.postLatencyTotal.Add(postLatency)
+			for old := hfh.postLatencyMax.Load(); old < postLatency; old = hfh.postLatencyMax.Load() {
```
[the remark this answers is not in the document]
alyssat2: Will do

PR #757, opened 2025-05-06
Repo cleanup

tiedotguy:
This performs a number of things around the repo and builds.  First is a bump to 1.24.2, and then a grab-bag of things, including:

- Move from `internal/tools` to `go tool`
- Remove golangci-lint.  It's unused, and needs to be properly integrated again, but until then it's dead code
- Remove coveralls - it's unused
- Move from goimports to gci
- Fix protoc-gen-go being globally installed
- Fix fuzz testing
- Regenerate `pb/gostatsd.pb.go` with actually used version

There is still a few things that are suboptimal:

- Re-add linting
- Fail the build if `pb/gostatsd.pb.go` is out of date
- Move to `go test -fuzz`
- Track coverage


tiedotguy:
This is the start of a huge yak shave.

hstan (approved):

PR #760, opened 2025-05-23
Allow configuring UDP packets buffer in receiver.

Wheezyx:
In systems where sender is not aggregating/consolidating metrics, smaller number of metrics per UDP datagram can be send. During socket.recvmsg golang syscall, even when buffer is not fully filled, entire buffer is being copied between kernel and user spaces. That account into more system CPU time send on this syscall.

In our system, we see ~2-8 metrics per datagram and up to 2-3 datagrams received with each syscall.

Allowing configuring UDP packet buffer size allows to reduce buffer size what will make syscall to copy less data with each call, theoretically reducing CPU overhead. Default is 64kB so theoretical maximal size of UDP datagram.

This PR allows configuring buffer size, if not set, default of 64kB is used to keep backwards compatibility. 

image

### DIFF
--- a/defaults_and_params.go
+++ b/defaults_and_params.go
@@ -79,6 +79,8 @@ const (
 	DefaultHeartbeatEnabled = false
 	// DefaultReceiveBatchSize is the number of datagrams to read in each receive batch
 	DefaultReceiveBatchSize = 50
+	// DefaultReceiveBufferSize is the number of size of a buffer for each datagram during reads
+	DefaultReceiveBufferSize = 0xffff
 	// DefaultEstimatedTags is the estimated number of expected tags on an individual metric submitted externally
 	DefaultEstimatedTags = 4
 	// DefaultConnPerReader is the default for whether to create a connection per reader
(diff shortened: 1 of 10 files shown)


defaults_and_params.go
```diff
@@ -215,7 +219,8 @@ func AddFlags(fs *pflag.FlagSet) {
 	fs.String(ParamStatserType, DefaultStatserType, "Statser type to be used for sending metrics")
 	fs.String(ParamPercentThreshold, strings.Join(toStringSlice(DefaultPercentThreshold), " "), "Space separated list of percentiles")
 	fs.Bool(ParamHeartbeatEnabled, DefaultHeartbeatEnabled, "Enables heartbeat")
-	fs.Int(ParamReceiveBatchSize, DefaultReceiveBatchSize, "The number of datagrams to read in each receive batch")
+	fs.Int(ParamReceiveBatchSize, DefaultReceiveBatchSize, "The number that defines size of buffer of each datagram during reads")
```
tiedotguy, line 222: Looks like the description in each line got mixed up

tiedotguy (changes requested):

tiedotguy (approved):

PR #791, opened 2026-07-19
Reject sample rates that are not positive finite numbers

samarth70:
## Problem

The sample rate parsed from the `@` field is stored without any validation:

```go
case '@':
	input := seekUntil(l, '|')
	v, err := strconv.ParseFloat(string(input), 64)
	if err != nil {
		l.err = err
		return nil
	}
	l.sampling = v
```

That value becomes `Metric.Rate`, which is used as a **divisor** during aggregation:

```go
value := int64(valueSum / m.Rate)   // metric_map.go, receiveCounter
t.SampledCount += 1.0 / m.Rate      // metric_map.go, receiveTimer
```

A rate that is not a positive finite number therefore does not produce a slightly-wrong metric, it produces a corrupt one. Measured on `linux/amd64` semantics for a single `foo:1|c|@<rate>`:

| line | counter value written | timer `SampledCount` |
|---|---|---|
| `@0` | `-9223372036854775808` | `+Inf` |
| `@-1` | `-1` | `-1` |
| `@NaN` | `-9223372036854775808` | `NaN` |
| `@+Inf` | `0` (metric silently disappears) | `0` |
| `@0.1` (valid) | `10` | `10` |

So one malformed packet can write `MinInt64` into a counter, or poison a timer's `SampledCount` with `NaN`, and both then propagate through aggregation to the backends.

The metric *value* is already guarded against this (`errNaN` in `Run`), but the sample rate is not.

## Fix

Reject a sample rate that is NaN, infinite, or not greater than zero, using the same "bad line" path the invalid NaN value already uses.

I deliberately did **not** reject rates greater than 1. Although the statsd convention is `0 < rate <= 1`, a rate above 1 still yields a finite, plausible number rather than a corrupt one, and rejecting it risks breaking senders that currently work. This change is limited to the values that actually corrupt aggregation.

## Testing

Added four cases to the existing `TestInvalidMetricsLexer` table. All four fail on current `master` (the lexer returns no error and the metric is accepted) and pass with the fix:

```
--- FAIL: TestInvalidMetricsLexer/bad.sampling.zero:1|c|@0
--- FAIL: TestInvalidMetricsLexer/bad.sampling.negative:1|c|@-1
--- FAIL: TestInvalidMetricsLexer/bad.sampling.nan:1|c|@NaN
--- FAIL: TestInvalidMetricsLexer/bad.sampling.inf:1|c|@+Inf
```

`go test ./internal/lexer/...` passes, along with 23 of the 25 packages in `go test ./...`. `gofmt` and `go vet` are clean.

The two packages I could not run are `pkg/statsd` and `internal/awslambda/extension/telemetry`; both fail on Windows for environmental reasons (a Unix domain socket path is parsed as a port, and an IPv6 `[::]` dial is refused). I confirmed both fail identically on an unmodified checkout, so they are unrelated to this change, but they are worth a look from CI.


decompil3d:
@atlassian-cla-bot check

samarth70:
Hi, the CLA is signed and green now. Is there anything needed from me to get CI to run on this? Happy to rebase onto current master if that helps.

PR #794, opened 2026-08-28
fix(lexer): prevent panic on overflowing Datadog event title/text lengths

jaideeppyne:
### Problem

`lexEventBody` bounds-checks the declared Datadog event title/text lengths against the remaining datagram bytes using a **uint32** sum:

```go
if l.len-l.pos < l.eventTitleLen+1+l.eventTextLen {
```

Both lengths are attacker-controlled and each may be up to `math.MaxUint32` (`lexUint32` only rejects values strictly greater than MaxUint32). When `eventTitleLen+1+eventTextLen` overflows uint32 it wraps to a small value, bypassing the "not enough data" guard, so the subsequent slice expressions (e.g. `l.input[l.pos : l.pos+l.eventTextLen]`) panic with `slice bounds out of range`.

The datagram parser runs in a goroutine with no `recover()`, so a single crafted UDP packet such as `_e{0,4294967295}:|foo` crashes the whole gostatsd process — a remote denial of service.

### Fix

Perform the length arithmetic and comparison in `uint64` so the guard cannot overflow. Over-long declared lengths now correctly return `errNotEnoughData`, matching the guard's intent. Regression cases added to `TestInvalidEventsLexer`.

### Testing

New cases fail (panic) before the change and pass after; full `internal/lexer` and `pkg/statsd` suites pass under `-race`; `gofmt`/`gci`/`go vet` clean.


jaideeppyne:
Closing because the required CLA remains pending and cannot be completed as part of this contribution.

